PR #46, opened 2014-12-06
debuginfo-install: accept packages names specified as NEVRA (RhBug:1171046)

spacekpe:


rholy:
We agreed that writing tests against plugins might be more work that writing the plugins itself. However, it isn't impossible. So if you don't have time to write them, fix at least the other comments, please.

spacekpe:
Okay, I replaced returnPkgLists with get_best_query instead. I hope this was the last point (except missing unit tests).

IMHO bigger changes like version selection algorithm should be made in separate commits.

mluscon:
No issues spotted during functionality testing.

jsilhan:
Merged, thanks

PR #50, opened 2014-12-07
debuginfo-install: fix handling of subpackages with non-zero epoch

spacekpe:
You can test is e.g. with package bind-utils or java-1.8.0-openjdk-headless (Fedora 21).


rholy:
Merged, thanks.

PR #54, opened 2015-01-22
plugins: add repomanage plugin (RhBug:1048541)

ignatenkobrain:


ignatenkobrain:
@rholy tests are missed, I know. I will write it soon.

jsilhan:
@ignatenkobrain , you're fast. It could be in dnf-plugins-extras.

ignatenkobrain:
> @ignatenkobrain , you're fast. It could be in dnf-plugins-extras.
Probably. We also need to discuss what to do with yum.rpmUtils. Anyway I've
used 3 functions here from it.

rholy:
1. "Avoid ```from x import y``` imports in cases where ```y``` is referenced less than five times in the source file (readability). [https://github.com/rpm-software-management/dnf/wiki/Hacking#coding-style]
2. Never call ```sys.exit```.
3. Could you please use ```os.walk``` in ```_get_file_list``` instead of the recursion?
4. Could you please use ```os.path.splitext[1]``` instead of ```d[-extlen:]```?
5. Could you please use ```set``` instead of ```self.unique```? We wouldn't have to deal with the PSF license and also it simplifies the code.
6. I'd use ```self.base.add_remote_rpm``` instead of the ```hdrFromPackage``` and ```pkgTupleFromHeader``` methods. What do you think, @jsilhan?
7. @jsilhan, can you think of a better way how to avoid ```compareEVR```? I'd store ```Package``` instances in the ```pkgdict``` instead of ```(e,v,r)```s and let the ```sort``` function sort them natively. But we need to document that packages are comparable.
8. Missing documentation.

ignatenkobrain:
@rholy do we need `sack_activation = True` for `self.base.add_remote_rpm`?

rholy:
Yes

ignatenkobrain:
> Yes
I think it will take some time, no?

rholy:
I'd address performance at the time someone complains about it. Simpler code is preferable, I think.

ignatenkobrain:
as discussed in IRC: I will use `add_remote_package()`, move to `dnf-plugins-extras` repo, replace all `sys.exit()` with `raise dnf.exceptions.Error()`

rholy:
One problem is that it's not *officially* allowed to raise exceptions in ```configure()```. So, please either move the command line parsing into ```run()``` (like some other commands do) or file an RFE, please.

jsilhan:
6) of course, it will be much simpler.
7) could be

ignatenkobrain:
Added to `dnf-plugins-extras` because this plugin used by rare people.
https://github.com/rpm-software-management/dnf-plugins-extras/commit/a8d87c2fa322fd00abe8330ad4651a2d490a0370

ignatenkobrain:
@rholy Thanks for review and suggestions. I've simplified plugin multiple.

PR #55, opened 2015-01-23
builddep: Support defining macros for parsing spec files

dm0-:
Some spec files allow enabling features by setting RPM macros, like
with the following snippet for example:

    %if 0%{?enable_png}
    BuildRequires:  libpng-devel
    %endif

This change adds an option consistent with RPM commands to define a
macro to use while parsing spec files.  It allows installing
explicitly requested build dependencies for spec files using macros
like the above.

Argparse was added for this option handling.

### DIFF
--- a/doc/builddep.rst
+++ b/doc/builddep.rst
@@ -17,6 +17,16 @@ Arguments
 ``<file>``
     The path to .src.rpm, .nosrc.rpm or .spec file, to read the needed build requirements from.
 
+-------
+Options
+-------
+
+``--help-cmd``
+    Show this help.
+
+``-D <macro expr>, --define <macro expr>``
+    Define the RPM macro named `macro` to the value `expr` when parsing spec files.
+
 --------
 Examples
 --------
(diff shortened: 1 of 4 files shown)


ignatenkobrain:
LGTM.

How about tests?

rholy:
Please fix these Pylint complaints:
```
C0330: builddep:140,0: Wrong continued indentation.
                            help=_('packages with builddeps to install'))
                            ^    |
C0330: builddep:142,0: Wrong continued indentation.
                            metavar="'MACRO EXPR'", type=macro_def,
                            ^    |
C0330: builddep:143,0: Wrong continued indentation.
                            help=_('define a macro for spec file parsing'))
                            ^    |
W0201: builddep:138,8: Attribute 'parser' defined outside __init__
W0201: builddep:144,8: Attribute 'opts' defined outside __init__
```
At least I don't understand why it's needed to assign the parser results to some instance attributes.

Also update the documentation, please.

dm0-:
I've updated the patch with documentation and tests added.  (The test spec file was modified, but I didn't regenerate the SRPM; the spec change shouldn't affect anything.)

In the previous version, I tried to be consistent with argparse usage in the `download` plugin.  This version tries to be consistent with the `repoquery` plugin, which results in empty output from `scripts/lint plugins/builddep.py`.

rholy:
The (new) code looks good to me. However I don't have time to check the rpm calls and test the functionality now. Maybe someone else has.

MichaelMraka:
Merged.

PR #59, opened 2015-01-29
Pylint ghost

MichaelMraka:


rholy:
That's praiseworthy, thanks.

Just a minor thing, can you please update the copyright notice (I believe in this case there should be an enumeration instead of a range but I'm not a lawyer and I can't find anything similar in our sources, so I'll leave it up to you)? Thanks in advance.

rholy:
LGTM... However I don't have time to test it. Maybe someone else.

jsilhan:
Merged, thanks

PR #60, opened 2015-01-29
pylint fix for Unused variable

MichaelMraka:
W0612: kickstart:126,26: Unused variable '_section'

### DIFF
--- a/plugins/kickstart.py
+++ b/plugins/kickstart.py
@@ -1,6 +1,6 @@
 # kickstart.py, supplies the 'kickstart' command.
 #
-# Copyright (C) 2013  Red Hat, Inc.
+# Copyright (C) 2013-2015  Red Hat, Inc.
 #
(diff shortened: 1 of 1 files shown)


rholy:
That's praiseworthy, thanks.

Just a minor thing, can you please update the copyright notice (I believe in this case there should be an enumeration instead of a range but I'm not a lawyer and I can't find anything similar in our sources, so I'll leave it up to you)? Thanks in advance.

And just a note for the future: There is a script ```scripts/lint``` that incorporate some project-specific false positives. In this case, the variable name was OK but unless it conflicts with our i18n ```_``` function, this is OK too.

rholy:
LGTM... However I don't have time to test it now. Maybe someone else.

MichaelMraka:
Lint scrips has changed so this is no more issue. Dropping.

PR #63, opened 2015-01-29
Buildscripts

MichaelMraka:
script and some minor changes to make building test packages easy one command job


rholy:
LGTM... However I don't have time to test it now. Maybe someone else.

BTW, I'd appreciate if you would have time to write a fix for the [CI repository](https://github.com/rholy/dnf-ci/blob/master/dnf-plugins-git2rpm.sh). But I can do it...

rholy:
Just out of curiosity, is there a use case for the adjustable dnf_version? (We still need to edit it in the spec file if the requirement changes.)

MichaelMraka:
Not very often just in case when you are testing package deps.

PR #64, opened 2015-01-29
let pylint ignore unused variables starting with _

MichaelMraka:


jsilhan:
where are such variables? @rholy knows how to name unused variable and not overwrite transaltion or not?

rholy:
The default is ```_|dummy```. Michael's change is a copy&paste from DNF's script to make it unified across all projects.

jsilhan:
Merged, thanks

PR #65, opened 2015-01-29
dnf config-manager implementation

MichaelMraka:


doc/config_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+  Public License for more details.  You should have received a copy of the
+  GNU General Public License along with this program; if not, write to the
+  Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+  02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+  source code or documentation are not subject to the GNU General Public
+  License and may only be used or replicated with the express permission of
+  Red Hat, Inc.
+
+=====================
```
ignatenkobrain: please add more `=`

MichaelMraka: Is there a different reason than just aesthetic?
(Just curious , I have no problem even with this one.)

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        if (self.opts.save or self.opts.set_enabled or
+                self.opts.set_disabled or self.opts.add_repo):
+            self.cli.demands.root_user = True
+
+
+    def run(self, _):
+        """Execute the util action here."""
+
+        if self.opts.set_enabled and self.opts.set_disabled:
+            logger.error(_("Error: Trying to enable and disable repos at the same time."))
```
ignatenkobrain: you redefined `_` in this function. it will not work.
and I think you want to raise error here instead of logging (`raise dnf.exception.Error`)

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+            self.opts.set_enabled = self.opts.set_disabled = False
+        if self.opts.set_enabled and not self.opts.repo:
+            logger.error("Error: Trying to enable already enabled repos.")
```
ignatenkobrain: Not tanslated

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+
+def sanitize_url_to_fs(url):
+    """Return a filename suitable for the filesystem
+
+    Strips dangerous and common characters to create a filename we
+    can use to store the cache in.
+    """
+
+    # code taken and modified from planet venus code base:
+    # http://intertwingly.net/code/venus/LICENCE
```
ignatenkobrain: I think we need to write our own implementation

plugins/dnfpluginscore/lib.py
```diff
@@ -0,0 +1,118 @@
...
+    else:
+        handle = non_repo_handle()
+    try:
+        librepo.download_url(url, fo.fileno(), handle)
+    except librepo.LibrepoException as e:
+        raise IOError(e.args[1])
+    fo.seek(0)
+    return fo
+
+
```
ignatenkobrain: so many empty strings

plugins/dnfpluginscore/lib.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def write_raw_configfile(filename, section_id, substitutions,
+                         cfgoptions, items, optionobj,
+                         modify=None):
+    """
+    Code adopted from yum-config-manager writeRawRepoFile().
+    """
```
ignatenkobrain: I think we need some more documentations about parameters for this function

plugins/dnfpluginscore/lib.py
```diff
@@ -0,0 +1,118 @@
...
+    except librepo.LibrepoException as e:
+        raise IOError(e.args[1])
+    fo.seek(0)
+    return fo
+
+
+
+
+
+def write_raw_configfile(filename, section_id, substitutions,
```
ignatenkobrain: tests missed for this function

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+
+        self.opts = self.parser.parse_args(args)
+
+        if self.opts.help_cmd:
+            print(self.parser.format_help())
+            return
+
+        if (self.opts.save or self.opts.set_enabled or
+                self.opts.set_disabled or self.opts.add_repo):
+            self.cli.demands.root_user = True
```
ignatenkobrain: use just `demands.root_user = True`. you already got `demands = self.cli.demands`

plugins/config_manager.py
rholy: what does this mean to the license of the DNF package? no change needed (even in the spec file)?

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        self.base = base
+        self.cli = cli
+        if self.cli is not None:
+            self.cli.register_command(ConfigManagerCommand)
+
+
+class ConfigManagerCommand(dnf.cli.Command):
+
+    aliases = ['config-manager']
+    summary = _('manage dnf configuration options and repositories')
```
ignatenkobrain: missed `usage = "something usage"`

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
...
+        demands = self.cli.demands
+        demands.available_repos = True
+
+        self.parser = dnfpluginscore.ArgumentParser(self.aliases[0])
+        self.parser.add_argument('repo', nargs='*',
+                    help=_('repo to modify'))
+        self.parser.add_argument('--save',
+                    default=False, action='store_true',
+                    help=_('save the current options (useful with --setopt)'))
+        self.parser.add_argument('--set-enabled',
```
ignatenkobrain: I'd like to use `--enable` instead. As in yum-config-manager

plugins/config_manager.py
```diff
@@ -0,0 +1,241 @@
+#
```
ignatenkobrain: no module docstring ;)

plugins/config_manager.py
rholy: @ignatenkobrain, why do you think so? Why should we reinvent a wheel?

plugins/config_manager.py
rholy: does it break some written convention? It's just a matter of personal preference. I don't think there is a need to force Michael to change it.

plugins/config_manager.py
rholy: Technically, there is no guideline that forces you to define ```usage```...

plugins/config_manager.py
rholy: Well, this error is explicitly disabled in our PyLint script. That could mean that we don't require them...

plugins/config_manager.py
MichaelMraka: sure, _ is a bug - originally it was _args and I wanted to silent pylint's complains so it ended up as simple _ :-)

MichaelMraka: as for logging vs. raise - yum in this case disables both enable and disable option and continue, so did I in plugin

plugins/config_manager.py
MichaelMraka: Well, it's Python software license and the same code (minus bugs I've fixed in it) is used in yum.
And frankly I've modified it so much that IMHO we can just strip this comment completely.

plugins/config_manager.py
MichaelMraka: nothing broken but I agree if there's already short name let's use it

plugins/config_manager.py
MichaelMraka: what is `usage` used for? I can see repoquery plugin defines `usage = _('[OPTIONS] [KEYWORDS]')` but neither --help or --help-cmd prints it.

plugins/config_manager.py
MichaelMraka: Me too :). Unfortunatelly it's not possible because `--enable` is prefix of `--enablerepo` and as such is eaten by dnf when parsing global options and there's no way to forward it to plugin.

plugins/config_manager.py
rholy: The [documentation](http://dnf.readthedocs.org/en/latest/api_cli.html#dnf.cli.Command.usage) says:

> Usage string for the command used as displayed by the CLI help.

TBH, I don't know where it is used.

plugins/config_manager.py
rholy: OK, IANAL, I trust you...

rholy:
Can we fix/workaround the problem with the ```--enable``` eating?

plugins/config_manager.py
ignatenkobrain: `dnf help cmd`

MichaelMraka:
I'm afraid not at reasonable price. Either parse both global and plugin option at once but that's impossible 'cause different plugins can have the same options... or disable option shortage which will cause regression from current behaviour.

plugins/config_manager.py
MichaelMraka: Yup, that's it. But frankly I'd prefer to see the same verbose help which `dnf cmd --help-cmd` produces.

rholy:
OH, I see what's the problem... (Since there is a DNF's switch ```--enablerepo``` and no other switch that starts with ```--enable``` prefix, the parser accepts ```--enable``` as ```--enablerepo```) Hm...

OK, I'd vote for disabling this feature in the DNF's parser. It's useful but since it limits plugins, I'd disable it until we refactor the parser. Then I'll enable (and document) it again. The behaviour is not documented, so it won't be a regression.

jsilhan:
I would also disable autoguessing of the parser.

MichaelMraka:
I agree disabling abbreviations is way to go but seems that it isn't possible in python 2.7.
http://stackoverflow.com/questions/10750802/disable-abbreviation
And looking into F21 python3.4/argparse.py nothing has changed there.
So... any B plan? :)

rholy:
Leave the plugin switch as ```--enable``` and hope that no one finds this bug? (and/or add an extra undocumented switch as a workaround for those who hit the bug?) :)

jsilhan:
I am against merging patches with bugs. I am fine with `--set-enabled`. We will document the change in https://fedoraproject.org/wiki/Yum_to_DNF_Cheatsheet. If users complain we will give them alternative `enable` as in COPR plugin.

plugins/config_manager.py
```diff
@@ -0,0 +1,242 @@
...
+            myrepodir = self.base.conf.reposdir[0]
+            dnf.util.ensure_dir(myrepodir)
+
+        for url in self.opts.add_repo:
+            if urlparse.urlparse(url).scheme == '':
+                url = 'file://' + os.path.abspath(url)
+            logger.info(_('Adding repo from: %s'), url)
+            if url.endswith('.repo'):
+                # .repo file - download, put into reposdir and enable it
+                destname = os.path.basename(url)
```
rholy: It sounds fearless to call ```basename``` on an URL, but if you are sure what are you doing...

plugins/config_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                destname = myrepodir + '/' + destname
+                try:
+                    with dnfpluginscore.lib.urlopen(self, None, url) as f:
+                        if not save_to_file(destname, f.read()):
```
rholy: Wouldn't it be safer to call ```shutil.copy``` with ```f.name```? I think it may be a security problem to read the whole file. But I'm not an expert.

plugins/config_manager.py
```diff
@@ -0,0 +1,242 @@
...
+                            continue
+                except IOError as e:
+                    logger.error(e)
+                    continue
+            else:
+                # just url to repo, create .repo file on our own
+                repoid = sanitize_url_to_fs(url)
```
rholy: Wouldn't it be nicer to parse the file, read the name of the repository and create the name of the file from the repository name?

xsuchy:
What about introducing ```--enable-reserved``` and document it as "not used, reserved for future use". It will effectively disable ```--enable``` as alias to ```--enablerepo```

rholy:
+1

MichaelMraka:
Guys - please, please - don't. I understand you are trying to find a solution how to allow `--enable` in plugin but all theses variants are more or less ugly workarounds. Let's admit that current code base can't do this in a nice way and use `--set-enabled`. I agree with @jsilhan we may create `--enable` as an alias for it in the future when option parser will be fixed.

rholy:
Just let me note that ```--set-enabled``` is also a workaround and it's the workaround that is the most difficult to revert... And the less systematic (another plugins would have to do the same if they want an ```--enable``` switch) one.
I said my opinion and my preference. That's everything I *can* and *want* to do. If you guys want something else, it's your responsibility...

MichaelMraka:
Tests added. Please merge if possible :).

rholy:
OMG, we should never call ```dnf.Base``` from tests!!! But as I see above, my comments are simply ignored... So, do what you want...

MichaelMraka:
% OMG, we should never call ```dnf.Base``` from tests!!! But as I see above, my comments are simply ignored... So, do what you want...

What other implementation of Base object am I supposed to use in tests?

rholy:
There is no good answer for it. I'd say you should create your own mock providing just the functionality that the plugin needs. We need to refactor ```Base``` before it can be used in tests.
I know, that's a lot of work. But I admit I cannot say *why* ```Base``` cannot be used. That's why I've removed the comment. I just remember there was a problem in the past with tests like this but I don't remember the details...
Just do your best. Sadly, so far I tend to say don't write tests until we find a safe general approach...

MichaelMraka:
Well, if e.g. dnf.Base immediately reads host system configuration and repos it might be a problem.
Therefore I did a test in a way they do not change system configuration, point cache to tmp etc.
And yes, I tried to use mock instead of real dnf.Base first but I ended up with more than a page and half of mock cheats and I still could not see the light in the end of tunnel. So I decided to use real Base before reimplementing half of them just for the tests.

rholy:
OK, it seems that you've made sure that the tests should not damage the host system. That's appreciated.

So, the last comments from me. (1) If I did not miss something, the plugin accesses basically just ```Base.conf```, ```Base.repos``` and ```Base.output```. I cannot think of anything difficult to mock there. Nevermind. (2) With this approach, I'd afraid that this will inspire an author of another plugin that won't make so much effort to make sure that his test won't break the system. But maybe I'm just too suspicious. (3) This approach adds a responsibility to the DNF maintainers to check that every change in ```Base``` does not make the test destructive. (4) The tests are not very readable to me because of the number of setup code. At least creating a new test specific subclass of ```Base``` that incorporates all the setup code would be a step towards better result for me. Or splitting the test case class to multiple classes with appropriate fixture ```setUp``` methods...

I just felt the need to say my opinion. The issues mentioned there are not blockers so they don't need to be addressed and if a DNF developers thinks this can be merged, I won't prevent him doing so.

plugins/config_manager.py
```diff
@@ -0,0 +1,250 @@
...
+                url = 'file://' + os.path.abspath(url)
+            logger.info(_('Adding repo from: %s'), url)
+            if url.endswith('.repo'):
+                # .repo file - download, put into reposdir and enable it
+                destname = os.path.basename(url)
+                destname = myrepodir + '/' + destname
+                try:
+                    f = dnfpluginscore.lib.urlopen(self, None, url, 'w+')
+                    f.close()
+                    shutil.copy2(f.name, destname)
```
xsuchy: Once you call f.close() then the file is deleted so copy2() have nothing to copy. You should pass delete=False to urlopen() to prevent this behaviour.

MichaelMraka: Oops, correct. Or I can call f.close() after it's copied.

MichaelMraka:
Here is fixed test with mocked base.

plugins/config_manager.py
```diff
@@ -0,0 +1,247 @@
...
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+from __future__ import absolute_import
+from __future__ import unicode_literals
+from dnfpluginscore import _, logger
+#from dnf.pycomp import PY3
```
rholy: Please remove this line.

MichaelMraka: Oops, sure. Fixed.

xsuchy:
So can this be merged please?

doc/config_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+Options
+-------
+
+``--help-cmd``
+    Show this help.
+
+``--add-repo=URL``
+    Add (and enable) the repo from the specified file or url.
+
+``--set-disabled``
```
jsilhan: Shouldn't be there the `--set-enabled=<repoid>`?

doc/config_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+    Disable the specified repos (automatically saves).
+
+``--set-enabled``
```
jsilhan: Shouldn't be there the `--set-enabled=<repoid>`?

doc/config_manager.rst
```diff
@@ -0,0 +1,76 @@
...
+``dnf config-manager --add-repo http://example.com/different/repo``
+    Create new repo file with http://example.com/different/repo as baseurl and enable it.
+
+``dnf config-manager``
+    Display main DNF configuration.
+
+``dnf config-manager repo``
+    Display configuration of repo.
+
+``dnf config-manager --set-enabled repo --save``
```
jsilhan: Is the `--save` redundant? There's written that `set-*abled` automatically saves

plugins/config_manager.py
```diff
@@ -0,0 +1,246 @@
...
+            dnf.util.ensure_dir(myrepodir)
+
+        for url in self.opts.add_repo:
+            if dnf.pycomp.urlparse.urlparse(url).scheme == '':
+                url = 'file://' + os.path.abspath(url)
+            logger.info(_('Adding repo from: %s'), url)
+            if url.endswith('.repo'):
+                # .repo file - download, put into reposdir and enable it
+                destname = os.path.basename(url)
+                destname = myrepodir + '/' + destname
```
jsilhan: `destname = os.path.join(myrepodir, destname)`?

plugins/config_manager.py
```diff
@@ -0,0 +1,246 @@
...
+                    shutil.copy2(f.name, destname)
+                    f.close()
+                except IOError as e:
+                    logger.error(e)
+                    continue
+            else:
+                # just url to repo, create .repo file on our own
+                repoid = sanitize_url_to_fs(url)
+                reponame = 'created by dnf config-manager from %s' % url
+                destname = myrepodir + '/' + repoid + '.repo'
```
jsilhan: `destname = os.path.join(myrepodir, "%s.repo" % repoid)`?

doc/config_manager.rst
MichaelMraka: > Is the --save redundant? There's written that set-*abled automatically saves

True. I'll remove it from doc.

plugins/config_manager.py
MichaelMraka: Yep, it's better.

plugins/dnfpluginscore/lib.py
```diff
@@ -0,0 +1,121 @@
...
+                         cfgoptions, items, optionobj,
+                         modify=None):
+    """
+    Code adopted from yum-config-manager writeRawRepoFile().
+    filename   - name of config file (.conf or .repo)
+    section_id - id of modified section (e.g. main, fedora, updates)
+    substitutions - instance of base.conf.substitutions
+    cfgoptions - ptions parsed from conf file (e.g. base.conf.cfg.options)
```
jsilhan: typo ptions -> options

tests/test_config_manager.py
```diff
@@ -0,0 +1,121 @@
...
+      conf = dnf.conf.Conf()
+
+class ConfigManagerCommandTest(unittest.TestCase):
+
+    def setUp(self):
+        cli = dnf.cli.cli.Cli(ConfigManagerBase())
+        self.cmd = config_manager.ConfigManagerCommand(cli)
+        self.cmd.base.conf.reposdir = [tempfile.mkdtemp()]
+
+        self.addCleanup(shutil.rmtree, self.cmd.base.conf.reposdir[0])
```
jsilhan: In unittest there shouldn't be os operations with files but this plugin is all about dealing with files so lets keep it there. Can you at least put it inside tearDown method which is IMO more clear.
```
def tearDown(self):
    shutil.rmtree(self.cmd.base.conf.reposdir[0])
```

rholy: No, ```addCleanup``` is better because it is run even if ```setUp``` fails. In this case, it does not matter. In other cases, ```adCleanup``` is better.

jsilhan:
Merged, thanks

PR #66, opened 2015-02-03
spec: python3 source code by default in f23+

jsilhan:


jsilhan:
@ignatenkobrain , the same change will have to be done in dnf-plugins-extras once it's merged.

package/dnf-plugins-core.spec
```diff
@@ -1,31 +1,36 @@
...
 
-Name:		dnf-plugins-core
+Name:		python-dnf-plugins-core
 Version:	0.1.4
 Release:	1%{?dist}
 Summary:	Core Plugins for DNF
 Group:		System Environment/Base
 License:	GPLv2+
-URL:		https://github.com/rpm-software-management/dnf-plugins-core
+URL:		https://github.com/akozumpl/dnf-plugins-core
```
ignatenkobrain: rebase wasn't good ;)

bkabrda:
This looks good to me, just one comment: the Python 2 subpackage requires python-requests. Shouldn't the Python 3 subpackage require python3-requests?

rholy:
https://bugzilla.redhat.com/show_bug.cgi?id=1194725 :)

rholy:
Another note: This effectively renames the ```dnf-plugins-core``` package to ```python-dnf-plugins-core```. Is that intentional? Doesn't we need to get some approval in advance? And also: one can assume from the name that it's a Python extension (since it starts with ```python-```) while I would argue this is more a DNF extension while the Python is just an inherited property. I think we should stay with ```dnf-plugins-core``` which should be Py3 based for rawhide and Py2 based for the rest.

bkabrda:
Yeah, I think this will need a new review and the old package will need to be retired... In related news, I had to revert Python 3 changes in F22 that were touching atomic host and/or minimal cloud image, which includes DNF too [1]... So please do this only for F23 (where there's enough time for rereview etc)

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1194577#c4

jsilhan:
right, `dnf-plugins-core` package should stay. I did that the same way as in dnf. One main package requiring either python or python3 version and postponed depending on python3 to f23.

rholy:
BTW, the new spec (and probably the current one too) cannot be built successfully on rawhide because we require "pykickstart" which was apparently switched to Python3 too and they did not bother to let us know. That's not an issue that should be fixed here but the PR cannot be merged because it cannot be tested until we fix the requires.

bkabrda:
@rholy it looks like pykickstart is now a "metapackage" that installs python3-pykickstart and there's still python-pykickstart with python 2 support: http://koji.fedoraproject.org/koji/buildinfo?buildID=613480

rholy:
Yes, sure, but "pykickstart" no longer represent the Python 2 version while DNF requires the Python 2 version.

bkabrda:
Don't get me wrong, I think it's a bit weird, I'm #justsaying-in-case-you-havent-noticed :)

rholy:
What is weird?

bkabrda:
It's weird to switch something that provided importables for Python 2 to Python 3. E.g. you were able to use `python2 -c "import kickstart"` having `pykickstart` installed and now you're not. So by `dnf update pykickstart` you break your system. That is the weird thing - although I think that the current packaging actually makes much more sense than the previous one.

rholy:
OH, definitely I agree with this. But isn't the whole Fedora weird then? You suggested us to do the same with DNF, didn't you?

bkabrda:
Heh, that's actually true, I did. But honestly I can't think of any other way to do the switch and not install python 2 along. The thing is that only dnf-* and anaconda-core depend on dnf, so the change is almost self contained and that's why I didn't give it that much thought. With pykickstart it's a different story, since there are > 15 various packages that depend on it, AFAICS.

rholy:
Makes sense. OK, so do you think that we should file a bug against "pykickstart" or should we rather adjust our spec? :)

bkabrda:
Honestly, I really see no way how to properly switch from the old packaging to the new one and at the same time break no stuff... (That's also a weird thing, there seems to be no sane way forward that wouldn't break stuff, at least none that I can think of right now).
So my suggestion is to adjust your spec. If you want to be super nice and prevent other packages from being broken, then maybe contact pykickstart maintainer and advise him to announce this and/or open bug for packages dependent on pykickstart, so that people don't get weird import errors all over the place.

rholy:
Good idea, I'll let the maintainer know.

jsilhan:
I wrote him to IRC if he plan to do the same in f22. If not I will move it into `0%{?fedora} >= 23` condition in new commit.

bkabrda:
FWIW I think this change should only be made in rawhide, since it breaks stuff and F22 is now in alpha (and also it'd bring Python 3 in minimal cloud image/atomic host, which we want to keep on Python 2 only for F22... unfortunately... ;))

jsilhan:
@rholy can you please run your super duper script on it?

rholy:
Sure, give me about 15 hours... :) OK, it probably need some optimizations :)

rholy:
BuildRequires are wrong too.

jsilhan:
fixed

rholy:
Relaunching... give it at least 8 hours...

rholy:
There is the problem with ```requests``` mentioned above.

rholy:
There is the problem with ```requests``` mentioned above. Hopefully, @xsuchy will fix it soon.

rholy:
Why did you move man pages into the subpackages? It would mean that they won't be installed if someone installs just a subpackage.

I did not tested it so maybe I'm wrong but it seems to me that you've removed locales from the Python 3 version, right?

BTW: there are these bugs that prevent me from building the RPM:
https://bugzilla.redhat.com/show_bug.cgi?id=1195715
https://bugzilla.redhat.com/show_bug.cgi?id=1195719

hrw:
It is nice to see you guys discussing stuff but maybe you can get rawhide back into buildable state?

https://bugzilla.redhat.com/show_bug.cgi?id=1198442 has patch for it.

There are people using rawhide daily in outside world...

rholy:
Can we be more constructive, please? Your comment doesn't help with anything. As you can see, we are working on the fix. Your patch was already a part of this PR and BTW doesn't solve the problem. The problem, as described in the bug, is that the package "dnf-plugins-core" currently installs Python 2 extensions, while the package "dnf" currently uses Python 3 extension in rawhide. Your patch fixes an ```ImportError``` while the problem is that there is no "repoquery.py" file in the Python 3 site-packages directory. So, even through your patch fixes the builds for rawhide, a new build won't fix your problem.

jsilhan:
bumped to 0.1.6, man pages and translations in python(3)-dnf-plugins-core packages.

rholy:
FTR, we agreed that you'll add the number of the new bug into the commit message and change the release number to 1.

jsilhan:
Merged

rholy:
Hmmm... None of the current packages is a good place for man pages...
If they were in the metapackage (```dnf-plugins-core```), they won't be available if a user installs just a subpackage (```python-dnf-plugins-core```). If they were in just one of the subpackages (```python-dnf-plugins-core```), they won't be available if a user installs just the other subpackage (```python3-dnf-plugins-core```). If they were in both subpackages (```python(3)-dnf-plugins-core```) and if a user updated just a single subpackage (```python-dnf-plugins-core```), they would conflict with the other old subpackage (```python3-dnf-plugins-core```).... Are you aware of a common practice, @bkabrda ? Moving them in a new subpackage is probably also a bad idea. So, we probably have to have different files for the Py 2 version and different for the Py 3 version....

bkabrda:
I don't think there's a common practice for this, it seems to only be the case of dnf-plugins-core. All the other packages with manpages that were influenced were only shipping executables with manpages in the main package (that would be dnf-plugins-core). There are two solutions I see:
- You could ship manpages in a separate subpackage dnf-plugins-core-man, on which both python{,3}-dnf-plugins-core would depend. The problem with that they should then have strict versioned dependency and updating just one python{,3}-dnf-plugins-core subpackage would not be possible.
- A better solution, IMO, is to ship renamed manfiles. For example:
  - python-dnf-plugins-core ships `python.dnf.plugin.*` manpages
  - python3-dnf-plugins-core ships `python3.dnf.plugin.*` manpages
  - Optionally, dnf-plugins-core ships symlinks `dnf.plugin.*` that point to the Python 3 versions.

I think the second solution is better, because you want people to
a) be able to update the subpackages separately at different times and
b) you may want to document different behaviour that is different (e.g. functionality may be limited on one of the interpreters because of missing dependencies, etc...)

Also, different names actually make even more sense if you take into account that people are going to access the functionality of python-dnf-plugins-core through a different binary - "dnf-2", if I'm not mistaken (in which case it might make more sense to name the manpages `dnf.plugin.<name>-2` and `dnf.plugin.<name>-3` to make it consistent and easy to find through tab completion).

PyDeq:
Yes, that's what I also thought that might be the best solution.

Or one simpler but less user-friendly:
IIRC, we've stated that ```dnf-3``` executable is just experimental. In that case, we can pretend that the only supported executable is ```dnf``` on all Fedoras and thus that the only way how to run DNF plugins is ```dnf <plugin-name>``` and thus we could just ship man pages with ```python-dnf-plugins-core``` on Fedora < 23 and with ```python3-dnf-plugins-core``` on Fedora >= 23. But there might be a catch in cases when upgrading from Fedora < 23 to Fedora >= 23 (some ```Conflicts```/```Obsoletes``` might be needed). Hm, maybe it's not simpler.... :)

jsilhan:
the man pages reflects code executed from `dnf` command only. I've moved that into `dnf-plugins-core` subpackage in 5429b16

PR #67, opened 2015-02-05
repoquery: Always get latest packages in a query output

pnemade:
Here is the use-case compared with yum's repoquery
[parag@f21 dnf]$ sudo repoquery harfbuzz
harfbuzz-0:0.9.36-1.fc21.i686
harfbuzz-0:0.9.36-1.fc21.x86_64

with the current repoquery plugin
[parag@f21 dnf]$ sudo dnf repoquery harfbuzz
harfbuzz-0:0.9.34-1.fc21.i686
harfbuzz-0:0.9.34-1.fc21.x86_64
harfbuzz-0:0.9.36-1.fc21.i686
harfbuzz-0:0.9.36-1.fc21.x86_64

If used this PR fix then the output will be
[parag@f21 dnf]$ sudo dnf repoquery harfbuzz
harfbuzz-0:0.9.36-1.fc21.i686
harfbuzz-0:0.9.36-1.fc21.x86_64

I can't think any usecase where we need to return all the available builds for any package.

Thanks.

### DIFF
--- a/plugins/repoquery.py
+++ b/plugins/repoquery.py
@@ -193,7 +193,7 @@ class RepoQueryCommand(dnf.cli.Command):
             print(QUERY_TAGS)
             return
 
-        q = self.base.sack.query().available()
+        q = self.base.sack.query().available().latest()
         if opts.key:
             if set(opts.key) & set('*[?'):  # is pattern ?
                 fdict = {'name__glob': opts.key}


rholy:
I think that the other use case may be e.g. that the latest version is broken and I need to know to which version should I downgrade. But in fact, I don't need the version, "dnf downgrade pkg" is just enough. Well, I don't know about any use case of repoquery. I understood it just as a tool to query repositories (like SQL) and then it's probably better to return all matching packages. The use case you mentioned does not describe for what will you use the result. BTW, for querying the latest packages, you can use "dnf list available harfbuzz".

pnemade:
For me it looks confusing to list all such available packages. This is also affecting other arguments like provides, requires and all such others. I need to then invest some time to find which lines of output is for which latest package.
See below, I don't think I need all the output lines
$ sudo dnf repoquery --provides libreoffice-langpack-pt-BR
libreoffice-langpack-pt_BR
libreoffice-langpack-pt-BR = 1:4.3.5.2-12.fc21
libreoffice-langpack-pt-BR(x86-64) = 1:4.3.5.2-12.fc21
libreoffice-langpack-pt_BR
libreoffice-langpack-pt-BR = 1:4.3.2.2-5.fc21
libreoffice-langpack-pt-BR(x86-64) = 1:4.3.2.2-5.fc21
libreoffice-langpack-pt_BR
libreoffice-langpack-pt-BR = 1:4.3.5.2-11.fc21
libreoffice-langpack-pt-BR(x86-64) = 1:4.3.5.2-11.fc21

If I have same package with some other version in some other repo say copr repo then I will get many lines of output.

rholy:
OK, this output is confusing, I agree. But it can also be fixed by adding a headline for each package version.

rholy:
> Here is the use-case compared with yum's repoquery
> [parag@f21 dnf]$ sudo repoquery harfbuzz
> harfbuzz-0:0.9.36-1.fc21.i686
> harfbuzz-0:0.9.36-1.fc21.x86_64

For what do you use the output?

pnemade:
I don't use that output in particular just gave that example. But, I do use commands to query what any package provides/requires frequently.  Once while using repoquery I found "repoquery <pkgname>" gives all available package information so I started using it.

rholy:
OK, so I assume the issue is the confusing output of "repoquery --query-option". I think it should be fixed rather by adding the package name there.

pnemade:
okay let me come with that fix

pnemade:
This become invalid now. Filed new PR#68

PR #70, opened 2015-02-10
download: Do not disable user-enabled repos

spacekpe:
It is not possible to disable repos with binary packages because the
same repos can contain source packages. This applies at least to Fedora
COPR and to user-managed repos too.

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -162,10 +162,9 @@ class DownloadCommand(dnf.cli.Command):
                 repo_dict[repo.id] = (repo, self.base.repos[source_repo])
             else:
                 repo_dict[repo.id] = (repo, None)
-        # disable the binary & enable the source ones
+        # enable the source repos
         for id_ in repo_dict:
             repo, src_repo = repo_dict[id_]
-            repo.disable()
             if src_repo:
                 logger.info(_('enabled %s repository') % src_repo.id)
                 src_repo.enable()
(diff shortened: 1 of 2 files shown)


rholy:
(I'd personally remove all the magic that does these repo manipulations.... This is based on some poor undocumented assumptions...)
What about:
```
if src_repo:
    repo.disable()
```
?
@jsilhan, what about removing the magic completely?

spacekpe:
I definitelly agree that hardcoded assumptions in the code are ugly but please do not remove them before we get metadata for repos.

Just removing the code would break current workflow: Users would be forced to either manually enable source repos or have them enabled all the time.

rholy:
...yes, like with every other repo... I mean, if you want to download a package, you need to have it's repo enabled... DNF doesn't enable all repos to find it...

spacekpe:
yum used to enable `-debug` repos in debuginfo-install command and `-source` repos for yum-download `-source`. I would personally improve this by adding metadata to the repo definition ("contains_src=0/1", "contains_debuginfo=0/1") but IMHO it is not desirable have source and debuginfo repos enabled all the time - metadata updates unnecasairly consume bandwith and time. 

To sum up, I find current behavior useful (but hacky, I agree with that).

jsilhan:
I am inclined to enable them when user specify `--source` and it should be documented. The code could be reduced.
https://github.com/rpm-software-management/dnf-plugins-core/pull/73
?

jsilhan:
https://github.com/rpm-software-management/dnf-plugins-core/pull/73 was merged instead.

PR #71, opened 2015-02-10
download: fix package download on Python 3

spacekpe:
Python 3 map() uses lazy evaluation so it packages were never copied to
output directory when the plugin was running under Python 3.


jsilhan:
Merged, thanks

PR #72, opened 2015-02-10
download: accept package names ending with .src too

spacekpe:


mluscon:
If we want to preserve consistency between plugins, we should consider a similar fix for other plugins too(e. g. repoquery). The patch and functionality look ok from my POV.

xsuchy:
Also note that there exist one kind of src.rpm package which have extension .nosrc.rpm
  http://developerblog.redhat.com/2014/12/10/how-to-package-proprietary-software/

jsilhan:
Right, there's another type but this commit does what is says and it's improvement without regression. Merged, thanks.

PR #73, opened 2015-02-11
download: Do not disable user-enabled repos (thanks Spacepe)

jsilhan:


spacekpe:
It looks fine, except the typo in "thanks Spacepe" (it should be "Spacekpe") :-)

PR #74, opened 2015-02-27
builddep: add feature to get builddeps from remote packages

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/plugins/builddep.py
+++ b/plugins/builddep.py
@@ -2,6 +2,7 @@
 # Install all the deps needed to build this package.
 #
 # Copyright (C) 2013-2015  Red Hat, Inc.
+# Copyright (C) 2015 Igor Gnatenko
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions of
@@ -62,9 +63,32 @@ class sink_rpm_logging(object):
 class BuildDepCommand(dnf.cli.Command):
 
     aliases = ('builddep',)
-    msg = "Install build dependencies for .src.rpm, .nosrc.rpm or .spec file"
+    msg = "Install build dependencies for .src.rpm, .nosrc.rpm, .spec file or repo"
     summary = _(msg)
-    usage = _("[PACKAGE.src.rpm|PACKAGE.nosrc.rpm|PACKAGE.spec]")
+    usage = _("[PACKAGE.src.rpm|PACKAGE.nosrc.rpm|PACKAGE.spec|PACKAGE]")
+
+    def __init__(self, args):
+        super(BuildDepCommand, self).__init__(args)
+        self.to_handle = {}
+        self.rpm_ts = rpm.TransactionSet()
+
+    def configure(self, args):
+        demands = self.cli.demands
+        demands.available_repos = True
+        demands.resolving = True
+        demands.root_user = True
+        demands.sack_activation = True
+        self._enable_source_repos()
+
+    @sink_rpm_logging()
+    def run(self, args):
+        for pkgspec in args:
+            if pkgspec.endswith('.src.rpm') or pkgspec.endswith('nosrc.rpm'):
+                self._src_deps(pkgspec)
+            elif pkgspec.endswith('.spec'):
+                self._spec_deps(pkgspec)
+            else:
+                self._remote_deps(pkgspec)
 
     @staticmethod
     def _rpm_dep2reldep_str(rpm_dep):
@@ -79,12 +103,12 @@ class BuildDepCommand(dnf.cli.Command):
             return False
         return True
 
-    def _src_deps(self, rpm_ts, src_fn):
+    def _src_deps(self, src_fn):
         fd = os.open(src_fn, os.O_RDONLY)
         if self.cli.nogpgcheck:
-            rpm_ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES)
+            self.rpm_ts.setVSFlags(rpm._RPMVSF_NOSIGNATURES)
         try:
-            h = rpm_ts.hdrFromFdno(fd)
+            h = self.rpm_ts.hdrFromFdno(fd)
         except rpm.error as e:
             if e[0] == 'public key not available':
                 logger.error("Error: public key not available, add "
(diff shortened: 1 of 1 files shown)


rholy:
See https://bugzilla.redhat.com/show_bug.cgi?id=1074585 for a discussion.

plugins/builddep.py
```diff
@@ -62,9 +63,32 @@ def __exit__(self, exc_type, exc, exc_tb):
 class BuildDepCommand(dnf.cli.Command):
 
     aliases = ('builddep',)
-    msg = "Install build dependencies for .src.rpm, .nosrc.rpm or .spec file"
+    msg = "Install build dependencies for .src.rpm, .nosrc.rpm, .spec file or repo"
     summary = _(msg)
-    usage = _("[PACKAGE.src.rpm|PACKAGE.nosrc.rpm|PACKAGE.spec]")
+    usage = _("[PACKAGE.src.rpm|PACKAGE.nosrc.rpm|PACKAGE.spec|PACKAGE]")
```
jsilhan: we probably want to allow `<package-name-spec>` here [1] and do `get_best_selector`, etc.
[1] http://dnf.readthedocs.org/en/latest/command_ref.html#specifying-packages

plugins/builddep.py
```diff
@@ -119,18 +143,27 @@ def _spec_deps(self, spec_fn):
...
-        demands.root_user = True
-        demands.sack_activation = True
-
-    @sink_rpm_logging()
-    def run(self, args):
-        rpm_ts = rpm.TransactionSet()
-        for fn in args:
-            if fn.endswith('.src.rpm') or fn.endswith('.nosrc.rpm'):
-                self._src_deps(rpm_ts, fn)
+    def _enable_source_repos(self):
```
jsilhan: move the method [1] from download plugin to lib adn import it, please
[1] https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/download.py#L161

jsilhan:
and add `...(RhBug:1074585)` to commit message

plugins/builddep.py
rholy: Yes, see the reference to Panu's comment in https://bugzilla.redhat.com/show_bug.cgi?id=1074585.

rholy:
Because of the Panu's comment, I'm against this "feature".

plugins/builddep.py
rholy: I mean no (I misread your comment).

MichaelMraka:
Merged.

rholy:
OK, so at least the Jan's request wasn't fulfilled and you've merged it, right? Hm...

timlau:
This "feature" is broken by design as discussed in 
 https://bugzilla.redhat.com/show_bug.cgi?id=1074585

It should never have been merged

MichaelMraka:
> OK, so at least the Jan's request wasn't fulfilled and you've merged it, right? Hm...

Which request do you have on mind?

MichaelMraka:
> This "feature" is broken by design as discussed in https://bugzilla.redhat.com/show_bug.cgi?id=905697#c21

It's "broken" in the same way builddep from src.rpm is. There's no difference whether src.rpm is local or comes from repo. So either supporting both makes sense or we should better remove src.rpm option as well.

rholy:
This is even more broken since, as Panu said, the downloaded SRPM is for an arbitrary arch which may have been build with different macros...

MichaelMraka:
The same applies both for srpm downloaded by dnf from repo and srpm  downloaded by user and passed to dnf as local file.

rholy:
> The same applies both for srpm downloaded by dnf from repo and 
This is not an excuse.
> srpm downloaded by user and passed to dnf as local file.
Not necessarily.

MichaelMraka:
>> The same applies both for srpm downloaded by dnf from repo and

> This is not an excuse.

?!? I'm afraid I don't understand this comment.

>>  srpm downloaded by user and passed to dnf as local file.

> Not necessarily.

Could you elaborate more on this?

timlau:
the mayor diffence by using a local .src.rpm and one fetched from the repo, is that the user selected the local one, but the user don't know what src.rpm he wil get in the second case, it depend a lot of the repo configuration.
It is right that 

dnf download foobar --source
yum builddep foobar*.src.rpm

will give the same result as your change, but there is lot of other ways to get .src.rpm files, like download from koji where you can get the specific .src.rpm build for the release you want.

rholy:
>>> The same applies both for srpm downloaded by dnf from repo and

>> This is not an excuse.

>?!? I'm afraid I don't understand this comment.

The fact that our "download" plugin is broken does not excuse the attempt to break another plugin the same way.

>>>  srpm downloaded by user and passed to dnf as local file.

>> Not necessarily.

> Could you elaborate more on this?

I think Tim have described it pretty well.

MichaelMraka:
> but there is lot of other ways to get .src.rpm files, like download from koji where you can get the specific .src.rpm build for the release you want.

Packages in source repos *are* built from koji so getting srpm from koji puts you into the exactly same trouble as a srpm from repo.
E.g. evolution-3.17.1-1.fc23.src.rpm in koji was built on arm - http://koji.fedoraproject.org/koji/taskinfo?taskID=9575140.

So chances srpm you download somwhere from the internet was built on a different arch is roughly equal to the same package from source repo. Therefore `dnf builddep package` makes no difference to `dnf builddep ./package.src.rpm`.

MichaelMraka:
> The fact that our download plugin is broken does not excuse the attempt to break another plugin the same way.

What's broken in download plugin and how it applies to builddep plugin?

rholy:
> So chances srpm you download somwhere from the internet was built on a different arch is roughly equal to the same package from source repo. Therefore dnf builddep package makes no difference to dnf builddep ./package.src.rpm.

Even if that's true, the SRPM still does not need to be download from an untrusted source. You can just build it locally. Or a colleague could build it locally. Or our nightly builds contains SRPMs build for correct architectures.

> What's broken in download plugin and how it applies to builddep plugin?

That it does not download the SRPM from which the given package was built. And that's the same problem that underlays this "discussion".

PR #76, opened 2015-03-02
create --destdir if not exist

MichaelMraka:
fixes issue when use of --destdir without an existing directory
downloads each package on top of each other to --destdir as a file


jsilhan:
Merged, thanks

PR #85, opened 2015-04-16
rename --repoid to --repo

MichaelMraka:


plugins/repoquery.py
```diff
@@ -83,8 +84,11 @@ def parse_arguments(args):
     parser = dnfpluginscore.ArgumentParser(RepoQueryCommand.aliases[0])
     parser.add_argument('key', nargs='?',
                         help=_('the key to search for'))
-    parser.add_argument('--repoid', metavar='REPO',
+    parser.add_argument('--repo', metavar='REPO', action='append',
```
jsilhan: not mention repoid in docs but support it, please.  i.e.: `parser.add_argument('--repo', '--repoid', metavar='REPO', action='append'`. we cannot drop it inmediatelly.

MichaelMraka: Yes, otherwise it would be printed by --help/--help-cmd.

timlau:
As it wrote in 
https://github.com/rpm-software-management/dnf-plugins-extras/pull/23

This would be an API break an inconsistance with all other tools there uses --repoid (both dnf/yum-utils)

timlau:
dnf & plugins takes a lot of beating in the fedora community, and renaming well known option is a bad move at this point of time IMO

MichaelMraka:
> This would be an API break an inconsistance with all other tools there uses --repoid (both dnf/yum-utils)

All dnf tools has been fixed to use consistent naming which is repo. Moreover --repoid does still work (for compatibility reasons) but is hidden from help.

timlau:
as log as --repoid still work every thing shoud be ok

PR #86, opened 2015-04-21
Plugin provides

jsilhan:


dnf-plugins-core.spec
```diff
@@ -1,4 +1,5 @@
-%{?!dnf_version: %global dnf_version 0.6.5}
+%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 0.6.5}
+%{?!dnf_not_compatible: %global dnf_not_compatible 0.7.0}
```
rholy: This is not true. If we are going to implement semantic versioning, the first possible incompatible version is 1.0.
See: http://semver.org/

jsilhan:
changed to 1.0 dnf upper boundary - we should make sure we bump it to 2.0 when we release dnf f22 final release 1.0

dnf-plugins-core.spec
```diff
@@ -26,7 +35,8 @@ protected_packages passive plugins.
 %package -n python-dnf-plugins-core
 Summary:    Core Plugins for DNF
 Group:      System Environment/Base
-BuildRequires:  python-dnf >= %{dnf_version}
+BuildRequires:  python-dnf >= %{dnf_lowest_compatible}
+BuildRequires:  python-dnf < %{dnf_not_compatible}
```
timlau: This feels very strange to me, I what situation is this needed
I understand the need for a minimun dnf version, but why a maximum version

rholy: To prevent breakage in case of an API change.

timlau: It is a build requirement so if you break API in dnf 1.0.0, you would properly change dnf-plugin-core to use the new API and build a new verison with
BuildRequires:  python-dnf >= 1.0 and release it togher with dnf-1.0.0
You can't push a new version of dnf into a repo there breaks requirments of an installed version of dnf-plugins-core, then it would not be able to update.

timlau: Now it get it, It is in the CI your are having issues, when building, every day directly from git

rholy: Sure but imagine this situation (which already happened):
1. a user has both ```dnf-0.6``` and ```dnf-plugins-core-0.1``` installed
2. then we release ```dnf-1.0``` and ```dnf-plugins-core-1.0``` (e.g. together if you want)
3. but the user runs just ```dnf upgrade dnf``` or ```dnf upgrade``` but e.g. ```dnf-plugins-core-1.0``` has some broken dependencies
4. the user now has ```dnf-1.0``` and ```dnf-plugins-core-0.1``` (!) installed
5. because ```dnf-plugins-core-0.1``` uses the old API, it kills DNF now

But with the ```< %{dnf_not_compatible}``` above, the user even would not be able to upgrade only ```dnf```. She would have to upgrade both or none.

MichaelMraka:
Merged.

PR #88, opened 2015-04-30
spec: allow DNF 1.x.x

rholy:
The major version bump did not introduce backwards incompatible changes.

It's needed to fix our CI job!

I did just some sanity tests since I don't expect that something should break.


jsilhan:
Merged, thanks. btw these changes you can push directly without PR ;)

PR #89, opened 2015-05-04
spec: fix an upgrade path from dnf-plugins-core <= 0.1.5

rholy:
Currently, it removes dnf-plugins-core when fresh installations of F22 are upgraded.
```
# dnf upgrade
...
Installing:
 python-dnf-plugins-core  noarch 0.1.7-1.fc22             updates-testing  77 k
     replacing  dnf-plugins-core.noarch 0.1.5-1.fc22
...
 python3-dnf-plugins-core noarch 0.1.7-1.fc22             updates-testing  53 k
     replacing  dnf-plugins-core.noarch 0.1.5-1.fc22
...

```

It would be nice if it could get into F22+ as soon as possible so that as few users as possible end up with no dnf-plugins-core package installed.

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -55,7 +55,7 @@ Requires:   python-kickstart
 Requires:   pykickstart
 %endif
 Requires:   python-requests
-Obsoletes:  dnf-plugins-core <= 0.1.5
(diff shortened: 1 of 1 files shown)


jsilhan:
Merged, thanks

PR #90, opened 2015-05-04
repoquery improvements

MichaelMraka:
- repoquery rewrite to use output.PackageFormatter
- better option separation (search vs. display options)
- `--resolve` fix  to allow `--qf`, `--requires`, `--suggests`, etc. together

### DIFF
--- a/plugins/repoquery.py
+++ b/plugins/repoquery.py
@@ -55,28 +55,42 @@ def parse_arguments(args):
     # Setup ArgumentParser to handle util
     parser = dnfpluginscore.ArgumentParser(RepoQueryCommand.aliases[0])
     parser.add_argument('key', nargs='?',
-                        help=_('the key to search for'))
-    parser.add_argument('--repo', metavar='REPO', action='append',
-                        help=_('show only results from this REPO'))
+        help=_('the key to search for'))
+    insearch = parser.add_argument_group(_('search arguments'))
+    insearch.add_argument('--repo', metavar='REPO', action='append',
+        help=_('show only results from this REPO'))
     # make --repoid hidden compatibility alias for --repo
-    parser.add_argument('--repoid', dest='repo', action='append',
-                        help=argparse.SUPPRESS)
-    parser.add_argument('--arch', metavar='ARCH',
-                        help=_('show only results from this ARCH'))
-    parser.add_argument('-f', '--file', metavar='FILE',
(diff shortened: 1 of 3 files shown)


plugins/repoquery.py
```diff
@@ -18,8 +18,8 @@
 
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from datetime import datetime
 from dnfpluginscore import _
+from dnf.cli.output import PackageFormatter, QUERY_TAGS
```
rholy: >  Avoid from x import y imports in cases where y is referenced less than five times in the source file (readability).

https://github.com/rpm-software-management/dnf/wiki/Hacking#user-content-coding-style

rholy:
Please don't merge this until it becomes clear that we want the new API introduced by https://github.com/rpm-software-management/dnf/pull/268

plugins/repoquery.py
```diff
@@ -132,45 +89,41 @@ def parse_arguments(args):
...
+        'suggests': _('display capabilities that the package suggests'),
+        'supplements': _('display capabilities that the package can supplement')
     }
     for arg in ('conflicts', 'enhances', 'obsoletes', 'provides', 'recommends',
                 'requires', 'suggests', 'supplements'):
         name = '--%s' % arg
-        const = '%%{%s}' % arg
-        outform.add_argument(name, dest='queryformat', action='store_const',
-                             const=const, help=help_msgs[arg])
+        display.add_argument(name, dest='capabilities', action='append_const',
```
rholy: I think it should rather go to `insearch`. It affects (or at least we have agreed that it should) searching while the format is affected only when `--resolve` is used.

plugins/repoquery.py
timlau: is dnf.cli.output public api ?
It is not good pratice to use undocumented, private api in plugins, I will end in every things is API like in yum
It will make it very hard to change things in the future without break stuff.
I know core plugins is maintained by dnf developers, but people making commuty plugins will look at core plugins for how to make plugins.

timlau: OK, I see there is a PR for this in DNF,  you should properly have wait for I was included before starting to use it.

timlau:
PackageFormatter should go into dnfpluginscore lib or just internal in repoquery, if there is no other plugins using it.

MichaelMraka:
Repoquery has been moved to dnf.

PR #92, opened 2015-05-21
add --tree & --alldeps option

max9631:


plugins/repoquery.py
```diff
@@ -222,6 +230,39 @@ def configure(self, args):
...
+
+            for provide in pkg.provides:
+                provideQuery = query.filter(requires=provide)
+                for needsProvidePkg in provideQuery.run():
+                    allpkgs.append(needsProvidePkg)
+        allpkgs = self.remove_pkg_duplicates(allpkgs)
+        alldepsquery = query.filter(pkg=allpkgs)
+        return alldepsquery
+
+    def remove_pkg_duplicates(self, array):
```
rholy: You can use `set(array)` instead of this method. Hopefully it's faster.

rholy: Hm, or even better, replace `allpkgs = []` with `allpkgs = set()` and `allpkgs.append(needsProvidePkg)` with `allpkgs.add(needsProvidePkg)`. You don't need to call `remove_pkg_duplicates` then.

doc/repoquery.rst
```diff
@@ -112,6 +115,9 @@ The following are mutually exclusive, i.e. at most one can be specified. If no q
 ``--resolve``
     resolve capabilities to originating package(s).
 
+``--tree``
+    For the given packages print a tree of the packages that require them.
```
jsilhan: Document with which option it works

plugins/repoquery.py
```diff
@@ -222,6 +230,39 @@ def configure(self, args):
...
 
+    def by_all_deps(self, name, query):
+        defaultquery = query.filter(name=name)
+        requiresquery = self.by_requires(self.base.sack, name, query)
+        allpkgs = []
+        for pkg in defaultquery:
+            for reqpkg in requiresquery.run():
+                allpkgs.append(reqpkg)
+
+            # for filePath in pkg.files:
```
jsilhan: delete commented lines

plugins/repoquery.py
```diff
@@ -260,22 +302,58 @@ def run(self, args):
...
+            q = self.by_all_deps(opts.whatrequires, q)
+        elif opts.whatrequires:
             q = self.by_requires(self.base.sack, opts.whatrequires, q)
+
         if opts.latest_limit:
             latest_pkgs = dnf.query.latest_limit_pkgs(q, opts.latest_limit)
             q = q.filter(pkg=latest_pkgs)
         fmt_fn = build_format_fn(opts)
+        if opts.tree:
+            if not opts.whatrequires:
```
jsilhan: Look if Argparser can check the additional option

plugins/repoquery.py
jsilhan: Update: it is not actually desired to search for all packages requiring files of the package. Delete it, please, it should have the whole `--alldeps` functionality now.

max9631:
Add --tree option to dnf repoquery (RhBug:1128425)

doc/repoquery.rst
```diff
@@ -75,6 +75,9 @@ Together with ``<pkg-spec>``, control what packages are displayed in the output.
 ``--whatrequires <capability>``
     Limit the resulting set only to packages that require ``<capability>``.
 
+``--alldeps``
+    Expand the resulting set of --whatrequires to packages that also require files that ``<capability>`` provides.
```
rholy: Sorry, I just can't understand it. I didn't get what it should do. But maybe it's just me...

rholy: OK, looking at the Bugzilla and at the implementation, it seems I know what's going on now.
So, I think that it would be worth mentioning that it expands the result only if `<capability>` is a package name and thus that it expands it with those "packages that requires any of the capabilities provided by the packages named `<capability>`".

plugins/repoquery.py
```diff
@@ -222,6 +230,20 @@ def configure(self, args):
...
 
+    def by_all_deps(self, name, query):
+        defaultquery = query.filter(name=name)
+        allpkgs = set()
+        requiresquery = self.by_requires(self.base.sack, name, query)
+        for reqpkg in requiresquery.run():
+            allpkgs.add(reqpkg)
+        for pkg in defaultquery.run():
+            for provide in pkg.provides:
+                provideQuery = query.filter(requires=provide)
```
rholy: If I could have a wish, please don't use mixedCase notation for names of variables. We basically follow the (PEP 0008)[https://www.python.org/dev/peps/pep-0008/] which says:

> [...] names should be lowercase, with words separated by underscores as necessary to improve readability.

plugins/repoquery.py
```diff
@@ -97,12 +100,16 @@ def parse_arguments(args):
...
+                        help="shows results that requires package provides and files")
     parser.add_argument('--querytags', action='store_true',
                         help=_('show available tags to use with '
                                '--queryformat'))
     parser.add_argument('--resolve', action='store_true',
                         help=_('resolve capabilities to originating package(s)')
-                       )
+                        )
+    parser.add_argument("--tree", action="store_true", 
+                        help="For the given packages print a tree of the packages.")
```
rholy: What about adding this to the `outform`? I think that this is just another type of output. What do you think?

rholy:
Remove the file `plugins/.repoquery.py.kate-swp` from the PR :)

rholy:
Heh, I like the names `tree_seed` and `grow_tree` :+1:

plugins/repoquery.py
max9631: Well, it is, but it could take argument from the ouform. In that case the outform couldn't be mutually exclusive group. (The argparse will make sure that only one of the arguments in the mutually exclusive group is present on the command line) https://docs.python.org/dev/library/argparse.html

rholy: Oops, I overlooked that it uses queryformat. OK, ignore this then.

rholy: Hm, I think that the arguments should be reimplemented. It should be possible to use `--queryformat` at the same time as e.g. `--requires`. `--requires` should set some other option than `queryformat` (it should set `queryformat` as well but only if `--queryformat` is not used) and thus `--queryformat` should not be in the same mutually exclusive group. Then `--tree` could be in the same mutually exclusive group as e.g. `--info` and `tree_seed` would not have to search for `"%{requires}"` in the `queryformat`. But let's not complicate your task. We can change it later...

plugins/repoquery.py
```diff
@@ -337,7 +413,7 @@ def buildtime(self):
 
     @property
     def conflicts(self):
-        return self._reldep_to_list(self._pkg.conflicts)
+        return self._reldep_to_list(self._pkg.obsoletes)
```
timlau: This dont look right, look like you are undoing another PR, You will proberly need to rebase you changes to the current HEAD

rholy:
By squashing the commits, we've lost the reference to the other bug. Can you please reword the commit message and reference both bugs there? I mean, mention in the message that it adds "--alldeps" and "--tree" and append `(RhBug:1128425) (RhBug:1186689)` to the end, please.

doc/repoquery.rst
```diff
@@ -103,6 +106,9 @@ The following are mutually exclusive, i.e. at most one can be specified. If no q
 ``--provides``
     Display capabilities provided by the package. Same as ``--qf "%{provides}``.
 
+``--tree``
+    For the given packages print a tree of the packages that require them. --tree also requires either of these options to work: --whatrequires --require --conflicts
```
rholy: just a typo: --require**s**

max9631: fixed

max9631:
Oh, didn't notice that. Thanks, i wouldn't find few extra lines that shouldn't be there anymore. It should be ok now

plugins/repoquery.py
```diff
@@ -260,22 +282,72 @@ def run(self, args):
...
         if opts.file:
             q = q.filter(file=opts.file)
         if opts.whatprovides:
             q = self.by_provides(self.base.sack, [opts.whatprovides], q)
-        if opts.whatrequires:
+        if opts.alldeps:
+            if not opts.whatrequires:
+                print("\n--alldeps requires --whatrequires option.")
+                print("Example: dnf repoquery --whatrequires audiofile --alldeps")
+                return
```
rholy: And one more thing. We need to raise an exception here to return a non-zero exit status.
So, can you please replace the `print`s and the `return` with e.g.

```
raise dnf.exceptions.Error(
    _("--alldeps requires --whatrequires option."
      "Example: dnf repoquery --whatrequires audiofile --alldeps"))
```

?

FYI, the ```_()``` function translates the string to the user's preferred language.

plugins/repoquery.py
```diff
@@ -260,22 +282,72 @@ def run(self, args):
...
             q = self.by_requires(self.base.sack, opts.whatrequires, q)
         if opts.latest_limit:
             latest_pkgs = dnf.query.latest_limit_pkgs(q, opts.latest_limit)
             q = q.filter(pkg=latest_pkgs)
         fmt_fn = build_format_fn(opts)
+        if opts.tree:
+            if not opts.whatrequires and "%{requires}" not in opts.queryformat and "%{conflicts}" not in opts.queryformat and "%{obsoletes}" not in opts.queryformat:
+                print("\n--tree requires either of these options: --whatrequires --requires --conflicts --obsoletes")
+                print("Example: dnf repoquery --whatrequires audiofile --tree")
+                return
```
rholy: The same here, please.

rholy:
I hit some problems:

- [ ] I forgot to add `\n` to the message of the exception when I provided the example in the previous comment. Could you please append `\n` at the end of the [first](https://github.com/rpm-software-management/dnf-plugins-core/pull/92/files#diff-fb427d111031f50d712b135c504fa5ddR294) string in [both](https://github.com/rpm-software-management/dnf-plugins-core/pull/92/files#diff-fb427d111031f50d712b135c504fa5ddR306) exceptions? Can you please also remove the `\n` from the beginning of the [first](https://github.com/rpm-software-management/dnf-plugins-core/pull/92/files#diff-fb427d111031f50d712b135c504fa5ddR306) string in the second exceptions?
- [ ] It fails on Python 3 (but I'm not sure whether it is a bug in `hawkey` or in this plugin):

```
# dnf-3 repoquery --conflicts --tree
Last metadata expiration check performed 0:20:34 ago on Fri May 29 17:23:22 2015.
0ad-0.0.18-1.fc22.x86_64
0ad-data-0.0.18-1.fc22.noarch
0install-2.6.1-2.fc21.x86_64
0xFFFF-0.3.9-11.fc22.x86_64
2048-cli-0.9-4.git20141214.723738c.fc22.x86_64
2048-cli-nocurses-0.9-4.git20141214.723738c.fc22.x86_64
2ping-2.0-5.fc22.noarch
389-admin-1.1.38-1.fc22.1.i686
389-admin-1.1.38-1.fc22.1.x86_64
389-admin-console-1.1.10-1.fc22.noarch
389-admin-console-doc-1.1.10-1.fc22.noarch
389-adminutil-1.1.21-2.fc22.i686
389-adminutil-1.1.21-2.fc22.x86_64
389-adminutil-devel-1.1.21-2.fc22.i686
389-adminutil-devel-1.1.21-2.fc22.x86_64
389-console-1.1.9-2.fc22.noarch
389-ds-1.2.2-6.fc21.noarch
389-ds-base-1.3.3.10-1.fc22.x86_64
  File "/usr/bin/dnf-3", line 35, in <module>
  File "/usr/lib/python3.4/site-packages/dnf/cli/main.py", line 185, in user_main
    errcode = main(args)
  File "/usr/lib/python3.4/site-packages/dnf/cli/main.py", line 84, in main
    return _main(base, args)
  File "/usr/lib/python3.4/site-packages/dnf/cli/main.py", line 134, in _main
    cli.run()
  File "/usr/lib/python3.4/site-packages/dnf/cli/cli.py", line 1077, in run
    return self.command.run(self.base.extcmds)
  File "/usr/lib/python3.4/site-packages/dnf-plugins/repoquery.py", line 308, in run
    self.tree_seed(q, orquery, opts)
  File "/usr/lib/python3.4/site-packages/dnf-plugins/repoquery.py", line 342, in tree_seed
    for name in sorted(set(strpkg)):
TypeError: unorderable types: _hawkey.Reldep() < _hawkey.Reldep()
```
- [ ] It fails with obsoletes (I'm not sure whether it's your mistake):
```
dnf-2 repoquery --obsoletes --tree
Last metadata expiration check performed 0:22:59 ago on Fri May 29 17:23:22 2015.
Error: 
--tree requires either of these options: --whatrequires --requires --conflicts --obsoletesExample: dnf repoquery --whatrequires audiofile --tree
```
- [ ] Could you please add a space [between](https://github.com/rpm-software-management/dnf-plugins-core/pull/92/files#diff-fb427d111031f50d712b135c504fa5ddR326) the package name and the list of its requires?

I'll take a look at it on Monday. I'm writing it here just in case someone else will have time to look at it.

rholy:
...that's a good point... What is the purpose of `sorted` here: https://github.com/rpm-software-management/dnf-plugins-core/pull/92/files#diff-fb427d111031f50d712b135c504fa5ddR342 ?

max9631:
- [x] I forgot to add \n to the message of the exception when I provided the example in the previous comment. Could you please append \n at the end of the first string in both exceptions? Can you please also remove the \n from the beginning of the first string in the second exceptions? 
 - Done
- [x] It fails on Python 3 (but I'm not sure whether it is a bug in hawkey or in this plugin)
 - I connot say whether it's bug in the plugin or in the hawkey as I'm not able to run dnf for Python 3. But if it were problem in the `sorted()` function, then it should be ok now. I removed it
- [ ] It fails with obsoletes (I'm not sure whether it's your mistake)
 - Yeah, the problem is in the dnf itself. It swallows the `--obsoletes` parameter so the parameter doesnt even get to the plugin
- [x] Could you please add a space between the package name and the list of its requires?
 - Yep, sorry for that

rholy:
Thanks. I'll wait for https://github.com/rpm-software-management/dnf/pull/284 before testing this.

rholy:
Looks good. Merged, thanks.

PR #93, opened 2015-05-27
repoquery: add srpm option (RhBug:1186382)

kudlav:


plugins/repoquery.py
```diff
@@ -218,35 +222,48 @@ def filter_repo_arch(opts, query):
         return query
 
     def configure(self, args):
+        (self.opts, self.parser) = parse_arguments(args)
+
+        if self.opts.srpm:
```
jsilhan: + `... and not self.opts.help_cmd:`

plugins/repoquery.py
```diff
@@ -218,35 +222,48 @@ def filter_repo_arch(opts, query):
...
-        if opts.key:
-            q = dnf.subject.Subject(opts.key, ignore_case=True).get_best_query(
+        if self.opts.key:
+            q = dnf.subject.Subject(self.opts.key, ignore_case=True).get_best_query(
                 self.base.sack, with_provides=False)
         else:
             q = self.base.sack.query()
 
-        if opts.pkgfilter == "duplicated":
+        if self.opts.srpm:
```
jsilhan: I am not sure if we want to transfer packages to srpm right after packages are found by main provides/pkgname expresion given or after additional filters (`--whatprovides`, `--whatrequires`, `--file`, `unsatisfied`, ..) application before printing package attributes (`fmt_fn = build_format_fn(self.opts)` line). Opinions @rholy, @MichaelMraka, @bkabrda ?

rholy: sooo... I think that
1. it should be applied *after* the filters for `--duplicated`, `--installonly` and `--unsatisfied` since I don't think that the user would ever want to list just *duplicated SRPMs*, *installonly SRPMs* or *unsatisfied SRPMs*...
2. it should be applied *after* the filter for `--arch` since otherwise no SRPM will be found, I guess
3. I'm not sure with the `--repo` filter... the easier variant would be to apply this filter *after* the `--repo` filter but I'm afraid that users probably expect the `--repo` filter to be applied on the binary packages, not the source packages... but if we apply this filter *before* the `--repo` filter it may not find the binary RPM (if e.g. `--repo fedora-sources` is used), or the source RPM (if e.g. `--repo fedora` is used)... (the `--repo` filter probably needs to be a bit smarter)
4. SRPM will never Provide anything, I think. OTOH `--whatrequires` makes sense. My preference is to apply this *before* the `--whatrequires/--whatprovides` filter. Or take a look at YUM.
5. the `--file` filter is questionable. I think that both variants make sense. Maybe take a look at YUM. I'd apply this *after* the `--file` filter.

I'd also reorder the output of `--help` in the order in which the filters are applied.

This plugin just annoys me... SQL rules.

MichaelMraka: IMHO it should go: apply all input filters / restrictions to find out list of packages, then lookup srpms for them and finally display information about these srpms according to display / output options

jsilhan:
Merged, thanks

PR #96, opened 2015-06-04
[copr] disable subcommand set `enabled=0`, not rm .repo file

evilkost:
Subcommand `disable` now only set `enabled=0`, repo file could be deleted by new subcommand `remove`.
This feature will be useful with https://fedoraproject.org/wiki/Changes/DisabledRepoSupport

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -19,6 +19,7 @@
 
 from __future__ import print_function
 from dnf.pycomp import PY3
+from subprocess import Popen, call
 from dnfpluginscore import _, logger
 from dnf.i18n import ucd
 import dnfpluginscore.lib
@@ -96,8 +97,17 @@ class CoprCommand(dnf.cli.Command):
             chroot = extcmds[2]
         except IndexError:
             chroot = self._guess_chroot()
-        repo_filename = "/etc/yum.repos.d/_copr_{}.repo" \
-                        .format(project_name.replace("/", "-"))
+
+        try:
+            copr_username, copr_projectname = project_name.split("/")
+        except ValueError:
+            logger.critical(
+                _('Error: ') +
+                _('use format `copr_username/copr_projectname` '
+                  'to reference copr project'))
+
+        repo_filename = "/etc/yum.repos.d/_copr_{}-{}.repo" \
+                        .format(copr_username, copr_projectname)
         if subcommand == "enable":
             self._need_root()
             self._ask_user("""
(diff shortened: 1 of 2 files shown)


plugins/copr.py
```diff
@@ -254,6 +269,15 @@ def _remove_repo(cls, repo_filename):
             raise dnf.exceptions.Error(str(e))
 
     @classmethod
+    def _disable_repo(cls, copr_username, copr_projectname):
+        exit_code = call(["dnf", "config-manager", "--set-disabled",
```
jsilhan: This is weird to call DNF plugin from DNF plugin externally. We should add this config manager subfunctionality into core lib and call it ftom Python. Opinions @MichaelMraka ?

MichaelMraka: > We should add this config manager subfunctionality into core lib and call it ftom Python. 

Which functionality? Disable, enable, all? IMHO would be better to provide simple way to call other dnf subcommands from plugin. Even now you can call

    repo = "{}-{}".format(copr_username, copr_projectname)
    cm = self.cli.cli_commands['config-manager'](self.cli)
    cm.configure(['--set-disabled', repo])
    cm.run(['--set-disabled', repo])

it's just not part of API.

rholy: > IMHO would be better to provide simple way to call other dnf subcommands from plugin.

I'm against this idea. At least in a form of command lines. I'd be OK if we publish the command classes and change their interface. But using something like ".find_command('foo').run('arg', 'arg')" is not acceptable to me.

jsilhan: We cannot find any better workaround to call plugins from plugins than external `call`. The code in Michaels comment will not work if the called plugin requests different demands. So I am fine with this line until we find better solution. @xsuchy , you can merge it once the issue with search command is fixed.

rholy: What's wrong with moving the functionality into the "lib" or into DNF's API?

rholy: Maybe you'd like to catch `OSError` or add a `Requires: /usr/bin/dnf` or handle the case when `/usr/bin/dnf` is not installed somehow else...

Anyway, what's wrong with moving the functionality into the "lib" or into DNF's API?

jsilhan: > Anyway, what's wrong with moving the functionality into the "lib" or into DNF's API?

time and it doesn't solve the general use case "call any plugin from plugin".

rholy: Well, it does solve the use case of disabling a repository permanently using API. ...unless you want to force every plugin, which wants to disable a repository, to call something like `base.get_command('config-manager').run('--set-disabled', repo)`... which is... hm, do what you want...

mluscon:
Ok to test

mluscon:
Test this please

xsuchy:
Tested, it works fine.
Merged as:
* c66a7bb [copr] fix regression with handling `search` and `list` subcommands
* f07ea54 [copr] terminate execution when failed to parse project name
* 66f0ea9 [copr] unused import
* e2a823b [copr] subcommand `disable` now only set `enabled=0`, repo file could be deleted by new subcommand `remove`

PR #111, opened 2015-09-17
download: clean tmp files after download

mluscon:


plugins/download.py
```diff
@@ -109,7 +124,12 @@ def _download_rpms(self, pkg_specs):
             pkgs = self._get_packages_with_deps(pkg_specs)
         else:
             pkgs = self._get_packages(pkg_specs)
-        self.base.download_packages(pkgs, self.base.output.progress)
+        try:
+            self.base.download_packages(pkgs, self.base.output.progress,
+                                        backup=False)
+        except dnf.exceptions.DownloadError as e:
+            errstring = dnf.i18n.ucd(e)
```
jsilhan: it should be Unicode already or the fix should be in `dnf.exceptions.DownloadError`

PR #113, opened 2015-09-21
Resolve local RPMs when downloading

atodorov:
This makes it easier to specify --source (and a --debuginfo which I'm going to add) and pass a local rpm files as argument. 

My use case is that I'm iterating over the local RPMs from a bash script and want to download their source/debuginfo counterparts w/o necessarily parsing the file name.

### DIFF
--- a/doc/download.rst
+++ b/doc/download.rst
@@ -33,6 +33,8 @@ Arguments
 
 ``<pkg-spec>``
     Package specification for the package to download.
+    Local RPMs can be specified as well. This is useful with the ``--source``
+    option or if you want to download the same RPM again.
 
 -------
 Options
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -171,7 +171,7 @@ class DownloadCommand(dnf.cli.Command):
         """Return a query to match a pkg_spec."""
         if os.path.exists(pkg_spec):
             pkg = self.base.add_remote_rpm(pkg_spec)
-            pkg_spec = pkg.name
+            pkg_spec = "{0.name}-{0.epoch}:{0.version}-{0.release}.{0.arch}".format(pkg)
 
         subj = dnf.subject.Subject(pkg_spec)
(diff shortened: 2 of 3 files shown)


atodorov:
Test implementation depends on https://github.com/rpm-software-management/dnf-plugins-core/pull/118

jsilhan:
how do you expect to behave when neither `--source` nor `--debuginfo` is selected? Download the package of the same nevra again?

atodorov:
Interesting question, but why not ?

For `--source` or `--debuginfo` this is clearly an advantage, give me the corresponding package for ./foo.rpm which is a valid use case.

When neither `--source` or `--debuginfo` is specified we can simply ignore the local package. OTOH you could have a situation where the user really wants to download the package from the remote repo and possibly compare it to the local one.  At this moment I don't have a strong preference b/c this is not my use case. However I'm inclined to leave it as-is (maybe add a few lines of help though). 

What are your preferences ?

atodorov:
Summary of changes:
1) I've fixed your comment and now using nevra for pkg_spec when trying to search for a local package. Tested it on a live system and appears to work

2) I've updated the tests:
  - there is one new commit which fixes the tests b/c fe1306 introduced changes which were not detected by the existing test suite (the one with the bad stubs/mocks). I will make a blog post explaining why this happened, ping me for more info.
  - previously the entire base object was a Mock but now this breaks with the latest hawkey version because it expects a string or a Package object in one of the constructors down the line. With this change this is resolved.
  - now the base.add_remote_rpm and base.download_packages methods are mocks so that I don't have to change the existing tests in any way. 

If you are happy with this PR please merge it b/c there start to be quite a few changes to the test suite which makes it harder for me to update the code on my other PRs.

jsilhan:
Can you document this feature in `download.rst`, please

mluscon:
Merged, thank you!

PR #114, opened 2015-09-21
Add --debuginfo to download

atodorov:
I don't have any tests though. Will look into that tomorrow, filing here so it doesn't get lost.

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -67,8 +67,11 @@ class DownloadCommand(dnf.cli.Command):
         self.parser = dnfpluginscore.ArgumentParser(self.aliases[0])
         self.parser.add_argument('packages', nargs='+',
                                  help=_('packages to download'))
-        self.parser.add_argument("--source", action='store_true',
-                                 help=_('download the src.rpm instead'))
+        target = self.parser.add_mutually_exclusive_group()
+        target.add_argument("--source", action='store_true',
+                             help=_('download the src.rpm instead'))
+        target.add_argument("--debuginfo", action='store_true',
+                             help=_('download the -debuginfo package instead'))
         self.parser.add_argument(
             '--destdir',
             help=_('download path, default is current dir'))
@@ -88,11 +91,16 @@ class DownloadCommand(dnf.cli.Command):
         if self.opts.source:
             dnfpluginscore.lib.enable_source_repos(self.base.repos)
 
+        if self.opts.debuginfo:
+            dnfpluginscore.lib.enable_debug_repos(self.base.repos)
+
     def run(self, args):
         """Execute the util action here."""
 
         if self.opts.source:
             locations = self._download_source(self.opts.packages)
+        elif self.opts.debuginfo:
+            locations = self._download_debuginfo(self.opts.packages)
         else:
             locations = self._download_rpms(self.opts.packages)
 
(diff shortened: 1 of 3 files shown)


atodorov:
I will look into it once I update the stubs in test_download.py to allow for proper testing.

atodorov:
Test implementation depends on https://github.com/rpm-software-management/dnf-plugins-core/pull/118

atodorov:
Correct installation of debuginfo depends on #117. Needs to be merged first.

plugins/download.py
```diff
@@ -122,6 +129,31 @@ def _download_source(self, pkg_specs):
...
+                src_spec = src_spec.replace(".rpm", "")
+                nevra = hawkey.split_nevra(src_spec)
+# TODO: specifying anything else than name doesn't work ATM
+# or at least it can't be tested. Need to look into it.
+#                dbg_name = "{}-debuginfo-{}:{}-{}.{}".format(
+#                                    nevra.name, nevra.epoch,
+#                                    nevra.version, nevra.release,
+#                                    pkg.arch
+#                            )
+                dbg_name = "{}-debuginfo".format(nevra.name)
```
jsilhan: #117 merged, reuse the logic from `lib.py`, please

atodorov: Will do, thanks!

atodorov:
I think this is as ready as it can be. Please review.

plugins/download.py
```diff
@@ -122,6 +129,34 @@ def _download_source(self, pkg_specs):
...
+        q = self.base.sack.query()
+        q = q.available()
+
+        for pkg in self._get_packages(pkg_specs):
+            for dbg_name in [dnfpluginscore.lib.package_debug_name(pkg),
+                            dnfpluginscore.lib.package_source_debug_name(pkg)]:
+                dbg_available = q.filter(
+                                        name=dbg_name,
+                                        epoch=int(pkg.epoch),
+                                        version=str(pkg.version),
```
jsilhan: any reason why you convert it to str? they are already unicoded and it could IMO cause more harm than good.

atodorov: > any reason why you convert it to str? 

Yeah, copy & paste :). I've taken this directly from `debuginfo-install.py` at the time, see _dbg_available. I will submit an updated PR in a second, but have in mind it isn't tested b/c of the test suite changes in PR #113. Please merge that one first.

plugins/download.py
```diff
@@ -69,6 +69,8 @@ def configure(self, args):
                                  help=_('packages to download'))
         self.parser.add_argument("--source", action='store_true',
                                  help=_('download the src.rpm instead'))
+        self.parser.add_argument("--debuginfo", action='store_true',
+                                 help=_('download the -debuginfo package instead'))
```
jsilhan: can add `--source` and `--debuginfo` options to the same group to be mutually exclusive. please? see repoquery.py how it's done

plugins/download.py
```diff
@@ -122,6 +129,34 @@ def _download_source(self, pkg_specs):
...
+                                        version=pkg.version,
+                                        release=pkg.release,
+                                        arch=pkg.arch
+                                    )
+                dbg_found = False
+                for p in dbg_available:
+                    dbg_pkgs.add(p)
+                    dbg_found = True
+
+                if dbg_found:
```
jsilhan: Do we need `dbg_found` variable? Isn't `dbg_pkgs` check sufficient?

tests/test_download.py
```diff
@@ -47,10 +49,18 @@ def evr(self):
 
     @property
     def sourcerpm(self):
+        name = self.name
+
+        # special cases for debuginfo tests
```
jsilhan: could source pkg name be defined by optional param in constructor, please? (If not defined than it would be same as pkg name)

jsilhan:
Moreover, document `--debuginfo` switch in download.rst, please

plugins/download.py
atodorov: It is used to break out of the for dbg_name in loop. I can give it more testing once #113 and its corresponding test changes are merged.

atodorov:
@jsilhan - I'm not sure about the dbg_found variable, but I need up-to-date tests before removing it. 
See the last commit, do you intended this to remove the if/then logic from the source property and simplify the tests ? If so, I can update the tests and squash the last few commits into one.

plugins/download.py
jsilhan: oh, I haven't noticed the nested loop. Never mind then.

mluscon:
Merged, thank you!

PR #115, opened 2015-09-22
Remove tests/mock.py

atodorov:
I've removed this file and made use of the system wide module. On Python3 we're not using it anyway and for Python2 we just need to update the BuildRequires.

On my system both Python2 and 3 tests pass.


rholy:
We just need to make sure that the package is also in EPEL...
BTW, it would be nice to do the same in DNF as well... I wonder why we bundled it at all...

atodorov:
It is. I've also requested its inclusion in RHEL 7, see
https://bugzilla.redhat.com/show_bug.cgi?id=1078300

Do you want me to look into dnf as well ?

dnf-plugins-core.spec
```diff
@@ -69,6 +70,8 @@ Group:      System Environment/Base
 BuildRequires:  python3-devel
 BuildRequires:  python3-dnf >= %{dnf_lowest_compatible}
 BuildRequires:  python3-dnf < %{dnf_not_compatible}
+# not needed b/c we're using unittest.mock
```
jsilhan: delete these comments, please (and add them optionally into commit description)

jsilhan:
Good patch.

> Do you want me to look into dnf as well ?

we would like to but we would like to have `python-mock` in RHEL. Have you submitted this package?

dnf-plugins-core.spec
atodorov: Hmm, I've added these b/c nobody will see the commit message after a while but they will see the comments in the spec file. The idea here is to not add python3-mock as unnecessary BuildRequires. 

I can remove them if you still feel that way.

atodorov:
Jan,
let's bump the bug in Bugzilla. Other components need mock too.

jsilhan:
Since the referred BZ is closed as WONTFIX then removing mock from the project does not make sense. Anyway thanks for the effort. We would like to get rid of bundling too but  we cannot get through bureaucracy issues...

PR #117, opened 2015-09-24
Fix for Bug 1265622 - 'dnf builddep' is not always as 'clever' as yum-builddep or dnf download --source

MichaelMraka:
This PR depends on #116 and should be merged after it.


jsilhan:
Merged, thanks

PR #118, opened 2015-09-25
Update stubs so we can properly test the _get_query methods

atodorov:
Previously the _get_query() and _get_query_source() methods were overridden with a stub_fn() method. IMO this is a poor test design because it doesn't allow these methods to be tested properly. 

In this PR I've updated the stub classes which now override functionality outside the DownloadCommand class. This allows new tests to assert directly against the _get_query methods. This is needed so I can write tests for PR#113 (directly affects _get_query) and PR#114 (may introduce a variation of _get_query).

### DIFF
--- a/tests/test_download.py
+++ b/tests/test_download.py
@@ -20,6 +20,8 @@ from __future__ import unicode_literals
 from tests.support import mock, RepoStub
 
 import dnf.repodict
+import dnf.sack
+import dnf.subject
 import dnfpluginscore
 import download
 import unittest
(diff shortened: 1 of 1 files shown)


atodorov:
ping, can we get any feedback on this one? it's blocking 2 other PRs

jsilhan:
I would like to see how looks it's usage (the improvement). Can you merge it into #114 where you and and tests for #114 so we can see it in action, please? 
(I am sorry but we are busy and the code review will not be done within next few days)

atodorov:
OK, I'll give it a go at #114 and if it's acceptable we can close this one.

atodorov:
Note: the benefit of this patch are quite subtle. I've played around with creating a few more tests and the benefit I see affect only a few lines of code. 

For #114 there doesn't seem to be any need to test _get_query directly, although we call
```
       q = self.base.sack.query()
       q = q.available()
```

which will benefit from this PR b/c we're stubbing out the entire Sack object. I will work on a test later today/tomorrow to see how it looks.

OTOH for #113 where we modify _get_query the test can look something like this:

```
+    def test_get_query_with_local_rpm(self):
+        try:
+            (fs, rpm_path) = tempfile.mkstemp('foobar-99.99-1.x86_64.rpm')
+            # b/c self.cmd.cli.base is a mock object add_remote_rpm
+            # will not update the available packages while testing.
+            # it is expected to hit an exception
+            with self.assertRaises(dnf.exceptions.PackageNotFoundError):
+                self.cmd._get_query(rpm_path)
+            self.cmd.cli.base.add_remote_rpm.assert_called_with(rpm_path)
+        finally:
+            os.remove(rpm_path)
```

Note the comment above the with block. If we leave out `_get_query` as before (a simple stub function) we're not going to be able to use `assert_called_with` later.

jsilhan:
@atodorov, this could be closed, right? The changes are AFAIK contained in https://github.com/atodorov/dnf-plugins-core/commit/6aab09093efa9544aa7354cec0f981f63e608e1f from #113

atodorov:
Right, closing

PR #122, opened 2015-10-06
Add flag to exit with non-zero status code

alde:
Add --fail flag to exit with a non-zero status code if a package is not found.

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -31,6 +31,7 @@ import hawkey
 import itertools
 import os
 import shutil
+import sys
 
 
 class Download(dnf.Plugin):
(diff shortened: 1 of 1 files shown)


jsilhan:
It should rather reuse `self.base.conf.strict` [option from DNF](http://dnf.readthedocs.org/en/latest/conf_ref.html?highlight=strict#repo-options).
* If all packages are found -> download them and exit 0
* If any package does not exist and strict=0 -> will not download any package at all and exit 1
* If any package does not exist and strict=0 -> will not download any package at all and exit 1

jsilhan:
Can you implement it this way ^, please?

alde:
I can, but since strict is True by default, this will change the default behaviour

PR #123, opened 2015-10-06
Exit with non-zero status if strict and package not found

alde:
If all packages are found, download packages and exit with status code
0.

If the `strict` base conf is True, and one or more packages are not
found, don't download any package and exit with status code 1.

If `strict` is False and one or more packages are not found, download
the found packages and exit with status code 0.

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -31,6 +31,7 @@ import hawkey
 import itertools
 import os
 import shutil
+import sys
 
 
 class Download(dnf.Plugin):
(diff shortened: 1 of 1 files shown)


plugins/download.py
```diff
@@ -131,6 +132,10 @@ def _get_packages(self, pkg_specs, source=False):
                 queries.append(func(pkg_spec))
             except dnf.exceptions.PackageNotFoundError as e:
                 logger.error(dnf.i18n.ucd(e))
+                if self.base.conf.strict:
+                    logger.error(_("Exiting due to strict setting.")
```
jsilhan: better. It should `raise dnf.exceptions.Error(e)`. (Plugins should not end with `sys.exit` but with raising the error)

alde: Good to know, have fixed

mluscon:
heads-up: This will influence the dnf download cli api, since the strict option is set to True by default.

mluscon:
retest this please

alde:
Yes, I mentioned this in #122 when I was asked to do it this way instead.

jsilhan:
squashed and merged, thanks

PR #129, opened 2015-11-06
Fix cmdline conversion to unicode (RhBug:1265210)

dmnks:
Previously, in Python 2, an implicit conversion to unicode was attempted
when the cmdline string was passed to str.split().  However, that failed
if some non-ascii bytes were present because the default ascii codec was
used.

This commit makes sure the string is converted properly.


jsilhan:
looks good. Jenkins, test this please.

jsilhan:
Jenkins, test this, please

mluscon:
Merged, thank you!

PR #131, opened 2015-11-13
repoquery: do not require loading metadata when we want to query syst…

jsilhan:
…em only

@j-mracek , this is your desired feature, test it, please.


jsilhan:
merged

PR #132, opened 2015-11-16
Translation

jsilhan:


piotrdrag:
dnf-plugins-core has many more translations in Zanata. This only pulls updated Catalan. :(

jsilhan:
@piotrdrag, can you be more specific, please. I have checked the **pl** translation and it seems to be complete in .po file. We use the same [zanata setup](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/po/zanata.xml) as in DNF.

jsilhan:
btw this was merged,

piotrdrag:
I don't know the reason for this, but https://github.com/rpm-software-management/dnf-plugins-core/tree/master/po has only ca and (not up to date) da translations, while https://fedora.zanata.org/iteration/view/dnf-plugins-core/master/languages has 10 translations at various stages of completeness.

jsilhan:
Thanks, I was blind. I had these files locally but wasn't staged in git. Fixed in https://github.com/rpm-software-management/dnf-plugins-core/commit/e83feb867825788844ffbd14017131f1a864e56f

piotrdrag:
Thanks a lot!

PR #138, opened 2015-12-02
lib: fix extraction of name from srpm string (RhBug:1286362)

mluscon:


jsilhan:
works as expected, merged, thanks

PR #139, opened 2015-12-31
builddep: do not download source package

whydoubt:
The build-dependency information for a source package is contained in
the repodata.  Because of this, it is not necessary to download the 
source package to evaluate the build dependencies.

### DIFF
--- a/plugins/builddep.py
+++ b/plugins/builddep.py
@@ -216,6 +216,11 @@ class BuildDepCommand(dnf.cli.Command):
                 name=(sourcenames + [package]), arch="src").latest().run()
         if not pkgs:
             raise dnf.exceptions.Error(_('no package matched: %s') % package)
-        self.base.download_packages(pkgs)
+        done = True
         for pkg in pkgs:
-            self._src_deps(pkg.localPkg())
+            for req in pkg.requires:
+                done &= self._install(str(req))
+
+        if not done:
+            err = _("Not all dependencies satisfied")
+            raise dnf.exceptions.Error(err)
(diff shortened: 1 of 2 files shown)


jsilhan:
Is this linked to any bugzilla report?

plugins/builddep.py
```diff
@@ -216,6 +216,11 @@ def _remote_deps(self, package):
                 name=(sourcenames + [package]), arch="src").latest().run()
         if not pkgs:
             raise dnf.exceptions.Error(_('no package matched: %s') % package)
-        self.base.download_packages(pkgs)
+        done = True
         for pkg in pkgs:
-            self._src_deps(pkg.localPkg())
+            for req in getattr(pkg, 'requires'):
```
jsilhan: why not just `pkg.requires`?

jsilhan:
Jenkins, test this, please

plugins/builddep.py
whydoubt: I just missed that when I switched from a variable to the constant 'requires'.

whydoubt:
1. No, there is no associated bugzilla report.  Is that preferred?
2. I cannot see the results from the failed check, so I do not know what may have happened there.

j-mracek:
Jenkins, test this, please

plugins/builddep.py
jsilhan: Can you change it to `pkg.requires` then, please. I find it more readable and so better for maintenance.

jsilhan:
> 1. No, there is no associated bugzilla report.  Is that preferred?

NO, just asking so we can close it once merged.

> 2. I cannot see the results from the failed check, so I do not know what may have happened there.

our CI needs improvement, nobody from outside is able to see logs. This was error during COPR build though.

jsilhan:
You can also place your name into [AUTHORS file](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/AUTHORS), see https://github.com/rpm-software-management/dnf#contribution

whydoubt:
I have made the requested updates.  Could you recheck it?

jsilhan:
Merged, thanks

PR #141, opened 2016-01-14
latest-limit option moved to base set of options making it compatible with --queryformat and other output formatters (RhBug: 1292475)

clime:
latest-limit option moved to base set of options making it compatible with --queryformat and other output formatters (see https://bugzilla.redhat.com/show_bug.cgi?id=1292475)


jsilhan:
Perfect. Can you add into commit message reference to bug in [described format](https://github.com/rpm-software-management/dnf-plugins-core#contribution), please? + you can [add your name to the AUTHORS file](https://github.com/rpm-software-management/dnf-plugins-core#contribution) ;).

clime:
Done!

jsilhan:
Merged, thanks

PR #144, opened 2016-01-27
Enable strings for translation (RhBug:1302214)

pnemade:
This will enable "available." string for translation.


ignatenkobrain:
+1 from me.

mluscon:
Merged, thank you.

PR #140, opened 2016-01-05
update debuginfo packages (RhBug: 1024701)

MichaelMraka:
enable debuginfo repositories if there are debuginfo packages installed and autoupdate=on

PR #145, opened 2016-02-13
Quit using rstrip for package_source_name

whydoubt:
In package_source_name, rstrip is being used to remove one string from
the end of another string.  However, that is not how rstrip is meant to
be used, and sometimes too much is stripped.

To fix this, a new _rchop method is introduced and used instead of
rstrip.

### DIFF
--- a/plugins/dnfpluginscore/lib.py
+++ b/plugins/dnfpluginscore/lib.py
@@ -137,6 +137,11 @@ def package_debug_name(package):
     """
     return "{}-debuginfo".format(package.name)
 
+def _rchop(origstr, chopstr):
+    if chopstr and origstr.endswith(chopstr):
+        return origstr[:-len(chopstr)]
+    return origstr
+
 def package_source_name(package):
     """"
     # :api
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
patch looks ok, BUT I have never seen that rstrip doesn't work correctly?

Is there some examples?

whydoubt:
In python documentation for rstrip: "The _chars_ argument is not a suffix; rather, all combinations of its values are stripped".  Conceptually, it is stripping characters from the end, one at a time, until a character NOT found in the _chars_ argument is encountered.

`'libtiff-4.0.4-1.fc23'.rstrip('-4.0.4-1.fc23')` → `'libti'`

ignatenkobrain:
Wow, now I found something new in python :D Thanks!

@jsilhan @mluscon merge then!

MichaelMraka:
The fix for this issue is already part of #140.

jsilhan:
@whydoubt thanks for the fix although the pull request for that was already there - we were just slow at reviewing (sorry). The original PR was merged and this is no longer needed. For that reason I am closing this.

PR #148, opened 2016-03-20
copr: Add logic to detect correct reposdir location and override detected copr chroot support

Conan-Kudo:
The `copr` plugin currently assumes that repos are installed to `/etc/yum.repos.d`. This is a problem as DNF itself can have repositories in `/etc/yum.repos.d`, `/etc/yum/repos.d`, or `/etc/distro.repos.d` by default.

We also expect that Fedora 25 will switch to using `/etc/distro.repos.d`, and Mageia will be using this path in Mageia 6. Even if this wasn't the case, it's a good idea to detect where the reposdir actually is and use that location.

Additionally, derivative distributions (such as Korora) need a way to override the detected chroots, so that they can use the parent distribution's chroots properly.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -47,29 +47,6 @@ if PY3:
 else:
     from ConfigParser import ConfigParser
 
-def get_reposdir():
-    myrepodir = None
-    # put repo file into first reposdir which exists or create it
-    for rdir in self.base.conf.reposdir:
-        if os.path.exists(rdir):
-            myrepodir = rdir
-
-        if not myrepodir:
-            myrepodir = self.base.conf.reposdir[0]
-            dnf.util.ensure_dir(myrepodir)
-    return myrepodir
-
-# Useful for forcing a distribution
-def chroot_config(self):
-    cp = self.read_config(self.base.conf, PLUGIN_CONF)
-    distribution = (cp.has_section('main')
-                  and cp.has_option('main', 'distribution')
-                  and cp.get('main', 'distribution'))
-    releasever = (cp.has_section('main')
-                  and cp.has_option('main', 'releasever')
-                  and cp.get('main', 'releasever'))
-    return [distribution, releasever]
-
 class Copr(dnf.Plugin):
     """DNF plugin supplying the 'copr' command."""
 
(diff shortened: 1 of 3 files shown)


jsilhan:
LGTM. We probably need to do something similar in `config-manager` plugin.

Conan-Kudo:
@jsilhan I actually stole the function for detecting reposdir from config-manager. ;)

plugins/copr.py
```diff
@@ -45,6 +47,29 @@
 else:
     from ConfigParser import ConfigParser
 
+def get_reposdir():
+    myrepodir = None
+    # put repo file into first reposdir which exists or create it
+    for rdir in self.base.conf.reposdir:
```
xsuchy: self is not defined here.

plugins/copr.py
```diff
@@ -45,6 +47,29 @@
...
+        if os.path.exists(rdir):
+            myrepodir = rdir
+
+        if not myrepodir:
+            myrepodir = self.base.conf.reposdir[0]
```
xsuchy: again. self is not defined

Conan-Kudo:
@xsuchy I think I fixed that now.

xsuchy:
And there is some PEP8 formating warnings (try to run python3-pylint on this module). But this is is minor and I can actually fix it myself later. But please fix that issue with self variable.

Conan-Kudo:
@xsuchy Fixed in latest amended commit.

plugins/copr.py
```diff
@@ -79,6 +81,29 @@ class CoprCommand(dnf.cli.Command):
   copr search tests
     """)
 
+# Useful for forcing a distribution
+    def _chroot_config(self):
+        cp = self.read_config(self.base.conf, PLUGIN_CONF)
```
xsuchy: File "/usr/lib/python3.4/site-packages/dnf-plugins/copr.py", line 86, in _chroot_config
    cp = self.read_config(self.base.conf, PLUGIN_CONF)
AttributeError: 'CoprCommand' object has no attribute 'read_config'

plugins/copr.py
```diff
@@ -240,9 +265,11 @@ def _need_root(cls):
 
     @classmethod
     def _guess_chroot(cls):
-        """ Guess which choot is equivalent to this machine """
+        """ Guess which chroot is equivalent to this machine """
         # FIXME Copr should generate non-specific arch repo
-        dist = platform.linux_distribution()
+        dist = self._chroot_config()
```
xsuchy: here is reference to self, while this method has cls.

Conan-Kudo:
@xsuchy I moved the method into the config() hook and adjusted accordingly. hopefully it should work...

jsilhan:
> @jsilhan I actually stole the function for detecting reposdir from config-manager. ;)

Then it would be a good idea to move it into `dnfcorelib` to not duplicate the code.

plugins/copr.py
```diff
@@ -240,9 +267,11 @@ def _need_root(cls):
 
     @classmethod
     def _guess_chroot(cls):
-        """ Guess which choot is equivalent to this machine """
+        """ Guess which chroot is equivalent to this machine """
         # FIXME Copr should generate non-specific arch repo
-        dist = platform.linux_distribution()
+        dist = CHROOT_CONFIG
```
xsuchy: I do not like the usage of global variable. Can you make CHROOT_CONFIG class attribute and refer to it as self.cli.CHROOT_CONFIG (and rename cls to self)?

plugins/copr.py
```diff
@@ -56,6 +60,17 @@ def __init__(self, base, cli):
         if cli is not None:
             cli.register_command(CoprCommand)
 
+    # Useful for forcing a distribution
+    def config(self):
+        cp = self.read_config(self.base.conf, PLUGIN_CONF)
```
xsuchy: File "/usr/lib/python3.4/site-packages/dnf-plugins/copr.py", line 65, in config
    cp = self.read_config(self.base.conf, PLUGIN_CONF)
AttributeError: 'Copr' object has no attribute 'base'

Please before you push another commit, do "tito --build --rpm -i; dnf copr enable fooo/bar" and test if it really works.

PR #150, opened 2016-03-22
Support reposdir detection and chroot override in copr plugin

Conan-Kudo:
The `copr` plugin currently assumes that repos are installed to `/etc/yum.repos.d`. This is a problem as DNF itself can have repositories in `/etc/yum.repos.d`, `/etc/yum/repos.d`, or `/etc/distro.repos.d` by default.

We also expect that Fedora 25 will switch to using `/etc/distro.repos.d`, and Mageia will be using this path in Mageia 6. Even if this wasn't the case, it's a good idea to detect where the reposdir actually is and use that location.

Additionally, derivative distributions (such as Korora) need a way to override the detected chroots, so that they can use the parent distribution's chroots properly.

This PR also adds a new API for getting `reposdir` in the `dnfpluginscore.lib`, and changes config_manager to use it.

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -161,16 +161,8 @@ class ConfigManagerCommand(dnf.cli.Command):
     def add_repo(self):
         """ process --add-repo option """
 
-        # put repo file into first reposdir which exists or create it
-        myrepodir = None
-        for rdir in self.base.conf.reposdir:
-            if os.path.exists(rdir):
-                myrepodir = rdir
-                break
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
This PR obsoletes #148.

Conan-Kudo:
@jsilhan @xsuchy Please review this PR. From what I've tested on my F23 system, it should work.

xsuchy:
Merged. Thank you.

PR #152, opened 2016-04-04
spec: improve python packaging according to new guidelines

ignatenkobrain:
* Subpackage with python2 files now prefixed with python2- instead of python-
* Added %python_provide macro for each subpackage to correctly provide python/python2
  versions
* Fix BuildRequires/Requires to require python2 packages in f23+ instead of just python
* Removed Group tags because it's not used anywhere
* Remove hacks around copying sources for building python3 version:
  There is no point to do this, because CMake provides ability to easy do builds
  out-of-tree, so let's just have 2 build directories for py2/py3.
* Use %license macro for license
* Use %make_build and %make_install macros
* Use %python2_sitelib for python2 package instead of %python_sitelib

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -2,130 +2,139 @@
 %{?!dnf_not_compatible: %global dnf_not_compatible 2.0}
 %global hawkey_version 0.6.1
 
-Name:       dnf-plugins-core
-Version:    0.1.20
-Release:    1%{?snapshot}%{?dist}
-Summary:    Core Plugins for DNF
-Group:      System Environment/Base
-License:    GPLv2+
-URL:        https://github.com/rpm-software-management/dnf-plugins-core
-# source archive is created by running package/archive from a git checkout
-Source0:    dnf-plugins-core-%{version}.tar.gz
-BuildArch:  noarch
+%if 0%{?rhel} && 0%{?rhel} <= 7
+%bcond_with python3
+%else
+%bcond_without python3
+%endif
+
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Looks good to me. It's closer in-line to my [dnf-plugins-core spec for Mageia](http://svnweb.mageia.org/packages/cauldron/dnf-plugins-core/current/SPECS/dnf-plugins-core.spec?view=markup). There might be some improvements there that you might want to "steal" for Fedora. 😀

Overall, :+1:

jsilhan:
This should go to rawhide only to avoid unintentional breaks in updates of plugins.

mluscon:
retest please

ignatenkobrain:
retest please

ignatenkobrain:
retest this please

ignatenkobrain:
@mluscon now it works with fixed tito ;)

dnf-plugins-core.spec
```diff
@@ -2,130 +2,139 @@
...
+Version:        0.1.20
+Release:        1%{?dist}
+Summary:        Core Plugins for DNF
+License:        GPLv2+
+URL:            https://github.com/rpm-software-management/dnf-plugins-core
+Source0:        %{url}/archive/%{name}-%{version}/%{name}-%{version}.tar.gz
+BuildArch:      noarch
+BuildRequires:  cmake
+BuildRequires:  gettext
+Requires:       python-%{name} = %{version}-%{release}
```
mluscon: Shouldn't this be python3-%{name}?

PR #157, opened 2016-04-12
support globs in --what<weak_dep> (RhBug:1303311)

MichaelMraka:


plugins/repoquery.py
```diff
@@ -226,11 +226,9 @@ class RepoQueryCommand(dnf.cli.Command):
...
     def by_dep(sack, pattern, query, dep):
-        try:
-            reldep = hawkey.Reldep(sack, pattern)
-        except hawkey.ValueException:
-            return query.filter(empty=True)
-        kwarg = {dep: reldep}
+        if dnf.util.is_glob_pattern(pattern):
+            dep += '__glob'
+        kwarg = {dep: pattern}
         return query.filter(**kwarg)
```
mluscon: `- import hawkey` then. BTW, isn't hawkey.Reldep faster than filter?

MichaelMraka: Reldep doesn't work with globs.

mluscon: Sure, but you can use it if the pattern is not glob.

MichaelMraka: > BTW, isn't hawkey.Reldep faster than filter?

It's the same. (Exactly same string to reldep conversion is the first thing which happen in the filter.)

mluscon:
Ok, LGTM now.

mluscon:
Merged

PR #158, opened 2016-04-18
Incorrect option --repo instead of --repoid in manual

xenithorb:
Also added a line break to the synopsis to fix ambiguity

### DIFF
--- a/doc/repoquery.rst
+++ b/doc/repoquery.rst
@@ -26,6 +26,7 @@ Synopsis
 --------
 
 ``dnf repoquery [<select-options>] [<query-options>] [<pkg-spec>]``
+
 ``dnf repoquery --querytags``
 
 -----------
@@ -76,7 +77,7 @@ Together with ``<pkg-spec>``, control what packages are displayed in the output.
 ``--recent``
     Limit the resulting set to packages that were recently edited.
 
-``--repo <repoid>``
+``--repoid <repoid>``
     Limit the resulting set only to packages from repo identified by ``<repoid>``.
     Can be used multiple times with accumulative effect.
 


ignatenkobrain:
LGTM!

xenithorb:
As I investigate this further it's not simply a mis-documented switch, something else is awry:

https://github.com/xenithorb/dnf-plugins-core/blob/master/plugins/repoquery.py#L88-92

Here we see:

```
    parser.add_argument('--repo', metavar='REPO', action='append',
                        help=_('show only results from this REPO'))
    # make --repoid hidden compatibility alias for --repo
    parser.add_argument('--repoid', dest='repo', action='append',
                        help=argparse.SUPPRESS)
```

Of course the reason I'm here at all is simply because `--repo` does not work, the parent DNF command complains about it being bad format for `--repofrompath`, which is when `--repoid` works as intended.

Example: 

```
 #> dnf repoquery --repo fedora-nvidia 
usage: dnf [--allowerasing] [-b] [-C] [-c [config file]] [-d [debug level]]
           [--debugsolver] [--showduplicates] [-e ERRORLEVEL]
           [--rpmverbosity [debug level name]] [-q] [-v] [-y] [--assumeno]
           [--version] [--installroot [path]] [--enablerepo [repo]]
           [--disablerepo [repo]] [-x [package]] [--disableexcludes [repo]]
           [--repofrompath [repo,path]] [--noplugins] [--nogpgcheck]
           [--disableplugin [plugin]] [--color COLOR]
           [--releasever RELEASEVER] [--setopt SETOPTS] [--refresh] [-4] [-6]
           [--downloadonly] [-h]
Command line error: argument --repofrompath: bad format: fedora-nvidia
```

Further sleuthing shows this commit: https://github.com/xenithorb/dnf-plugins-core/commit/d1089a7ad58cab5daf9004ca41f9e58d9003806b

by @MichaelMraka which is basically the reverse of what I'm trying to accomplish. Perhaps "--repo" is no longer warranted due to the incompatibility with dnf proper, and we can just revert Michael's commit? 

I would not merge my PR at this point

xenithorb:
I tried reverting the commit and pushing to this PR but it resulted in a few conflicts that I'm afraid I am not willing to attempt. This appears to be out of my capabilities at this point.

mluscon:
Hi, we are currently working on a better argument parser which should also cover this case. Thank you for your attempt but I suppose it would be better to wait for a new parser than fix this bit separately. If you want to coordinate your work with us, you can find us on yum/freenode channel.

PR #163, opened 2016-05-09
Set --available option as a default for dnf repoquery

j-mracek:


doc/repoquery.rst
```diff
@@ -55,7 +55,7 @@ Together with ``<pkg-spec>``, control what packages are displayed in the output.
     Limit the resulting set to leaves packages that were installed as dependencies so they are no longer needed. This switch lists packages that are going to be removed after executing ``dnf autoremove`` command.
 
 ``--available``
-    Limit the resulting set to available packages only.
+    Limit the resulting set to available packages only (set by default).
```
ignatenkobrain: shouldn't we say that it's deprecated?

ignatenkobrain:
what about `--installed` option?

j-mracek:
--installed works like expected, same as --available. But you know, there can be only one default value :-)  .

j-mracek:
I am going to allow combination of --installed and --available options

doc/repoquery.rst
j-mracek: and what about only (default) ? I am enabling of --installed with --available, therefore it is still handy.

ignatenkobrain:
I don't think that it's good idea

j-mracek:
@ignatenkobrain Why? - It was default behavior of repoquery (installed + available) and may be someone like it. Now we change default to available, but it will be nice to provide an option combination that can mimic formal behavior.

j-mracek:
Jenkins, test this please

mluscon:
again, incompatible change -> dnf-2.0

PR #164, opened 2016-05-09
Set --alldeps as defaults in repoquery and add --nodeps option

j-mracek:
It sets --alldeps as default according to yum-dnf compatibility.

It also introduce --nodeps option that represent formal default status of
"dnf repoquery --whatrequires"

### DIFF
--- a/doc/release_notes.rst
+++ b/doc/release_notes.rst
@@ -23,6 +23,12 @@ Core DNF Plugins Release Notes
 
 
 ====================
+0.1.21 Release Notes
+====================
+
+Setted --alldeps as defaults in Repoquery plugin and added --exactdeps option.
+
+====================
 0.1.20 Release Notes
 ====================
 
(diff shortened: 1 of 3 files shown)


j-mracek:
Jenkins, test this please

jsilhan:
I've told you wrong option name. Please change `--nodeps` to `--exactdeps`. You can also reuse the man description form original repoquery ;)

jsilhan:
Also with these bigger changes, can you document it in advance in release notes so we don;t miss them?

j-mracek:
All changes applied and rebased

doc/release_notes.rst
```diff
@@ -23,6 +23,12 @@ Core DNF Plugins Release Notes
 
 
 ====================
+0.1.21 Release Notes
+====================
+
+Setted --alldeps as defaults in Repoquery plugin and added --exactdeps option.
```
ignatenkobrain: AFAIK in English there is no such word - `setted`, there is just `set`.

mluscon:
Labeling by dnf-2.0 as this is an incompatible change.

PR #170, opened 2016-05-13
Fix reading the copr.conf file and add Mageia chroot selection support

Conan-Kudo:
This pull request fixes how `copr.conf` was read so that it doesn't trigger a traceback anymore (reported by Korora's @csmart).

In addition, I've added support for selecting Mageia chroots automatically.

(Obsoletes #169)

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -96,7 +96,7 @@ class CoprCommand(dnf.cli.Command):
 
 
         # Useful for forcing a distribution
-        raw_copr_plugin_config = ConfigParser()
+        copr_plugin_config = ConfigParser()
         config_file = None
         for path in self.base.conf.pluginconfpath:
             test_config_file = '{}/{}.conf'.format(path, PLUGIN_CONF)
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Please also push these to the `dnf-1.1` branch.

plugins/copr.py
```diff
@@ -96,23 +97,21 @@ def configure(self, args):
...
-            releasever = (cp.has_section('main')
-                          and cp.has_option('main', 'releasever')
-                          and cp.get('main', 'releasever'))
-            self.chroot_config = [distribution, releasever]
-
+            copr_plugin_config.read(config_file)
+            if copr_plugin_config.has_option('main', 'distribution') and copr_plugin_config.has_option('main', 'releasever'):
+                distribution = copr_plugin_config.get('main', 'distribution')
+                releasever = copr_plugin_config.get('main', 'releasever')
+                self.chroot_config = ["{}".format(distribution), "{}".format(releasever)]
```
ignatenkobrain: no point to do here `format()`

plugins/copr.py
```diff
@@ -276,6 +275,18 @@ def _guess_chroot(cls):
                 chroot = ("fedora-rawhide-x86_64")
             else:
                 chroot = ("fedora-{}-x86_64".format(dist[1]))
+        elif "Mageia" in dist:
+            # Detect architecture (Mageia does not use $basearch)
+            distarch = platform.machine()
+            if re.match("^i[3-6]86$", distarch, flags=0) is not None:
```
ignatenkobrain: 1. No need to use `flags=0`
2. No need to use `is not None`

Conan-Kudo:
@xsuchy Could you please review this?

plugins/copr.py
```diff
@@ -96,23 +97,21 @@ def configure(self, args):
...
-                            and cp.get('main', 'distribution'))
-            releasever = (cp.has_section('main')
-                          and cp.has_option('main', 'releasever')
-                          and cp.get('main', 'releasever'))
-            self.chroot_config = [distribution, releasever]
-
+            copr_plugin_config.read(config_file)
+            if copr_plugin_config.has_option('main', 'distribution') and copr_plugin_config.has_option('main', 'releasever'):
+                distribution = copr_plugin_config.get('main', 'distribution')
+                releasever = copr_plugin_config.get('main', 'releasever')
```
xsuchy: This read distribution and releasever only when both are set. While previously it was:
```
  if copr_plugin_config.has_option('main', 'distribution')
      distribution = ...
  if copr_plugin_config.has_option('main', 'releasever')
      releasever = ...
```
Hmm but it does not matter, because i _guess_chroot we set it from platform.linux_distribution()
if one of them is not set, overriding the other. So I'm waiving out this difference.

xsuchy:
Merged to master. Thank you.
I will leave backport to 1.1 branch on @ignatenkobrain or @jsilhan

ignatenkobrain:
Merged into 1.1 branch. @xsuchy thanks for review.

PR #172, opened 2016-05-26
spec: explicitly conflict with python-%{name} with different version

ignatenkobrain:
When we have installed python-dnf-plugins-core (before %python_provide) it is
possible to install python3-dnf-plugins-core with new version because it conflicts
with python2-dnf-plugins-core which is not installed. Let's fix it.


jsilhan:
If it works I am fine with this and could be merged directly.

ignatenkobrain:
@j-mracek could you test this PR if it fixes problem as you was able to reproduce it?

j-mracek:
@ignatenkobrain Sorry I had PTO

PR #177, opened 2016-06-22
Adjust to privatization of function from dnf.sack.rpmdb_sack()

j-mracek:
To make more distinguishable API and private functions the def rpmdb_sack was
renamed to _rpmdb_sack, therefore it has to be reflected in DNF-core.


j-mracek:
Requires rpm-software-management/dnf#519

j-mracek:
Requires rpm-software-management/dnf#520

MichaelMraka:
rebased and merged
76848ea3ad6ea57f8258e86337ad83e9cb63b47b

PR #181, opened 2016-07-12
Fix string puzzle in translatable message

tosky:
Use only one string for each message, more clear for translators (and
properly translatable for RTL languages).


mluscon:
@dnf-bot r+

ignatenkobrain:
@dnf-bot retry

ignatenkobrain:
@dnf-bot retry

PR #185, opened 2016-07-27
introduced config-manager --dump-variables (RhBug:1360752)

MichaelMraka:


plugins/config_manager.py
```diff
@@ -88,6 +91,9 @@ def modify_repo(self):
         modify = {}
         if hasattr(self.opts, 'main_setopts') and self.opts.main_setopts:
             modify = dict(self.opts.main_setopts._get_kwargs())
+        if self.opts.dump_variables:
+            for name, val in self.base.conf.substitutions.items():
+                print("%s = %s" % (name, val))
```
ignatenkobrain: * I think there should be no spaces (I think should be quite useful to do `| grep "^key=" | eval`
* value must be quoted (see `shlex.quote()`)

MichaelMraka: But there are already spaces in config dump so let's make it consistent.

ignatenkobrain: then I think quoting should be added in config dump as well.... but it's for another PR.

ignatenkobrain:
@dnf-bot r+

plugins/config_manager.py
MichaelMraka: Could be... but then it'll break  the damn yum compatibility ;).

ignatenkobrain:
@dnf-bot retry rollup

ignatenkobrain:
@dnf-bot retry

PR #187, opened 2016-07-27
spec: bump lowest dnf compatible version to 2.0.0

mluscon:
We forgot on this...


mluscon:
lol we hit the same issue with @j-mracek :-). I am closing this one.

PR #188, opened 2016-07-28
cls.chroot_config inside _guess_chroot returns None (RhBug: 1361003)

xenithorb:
1. L#259: cls.chroot_config returns None
2. L#107: self.chroot_config is defined properly but is not passed to
_guess_chroot()
   - self changed to class identifier CoprCommand - works

Disclaimer: I am not a Python developer and I merely stumbled on a solution that worked for me. I verified with `pdb` that the necessary values were being passed correctly before submitting the patch. I have no idea why the original code doesn't work versus specifying the class name. 

My personal use case is Fedy on Korora. Fedy relies on some COPR repositories, and it is completely broken right now (that is, all COPR repos on any system that needs to use `/etc/dnf/plugins/copr.conf` don't currently work because `dnf copr` returns that it can not find any builds

```
Error: This repository does not have any builds yet so you cannot enable it now. 
```

https://bugzilla.redhat.com/show_bug.cgi?id=1361003


Conan-Kudo:
@xenithorb Can you please test this patch against the `dnf-1.1` branch of this repository? Fedora/Korora does not yet ship DNF 2.0 (which is `master`), so you'd need to make a fix for the `dnf-1.1` branch. Please make a separate pull request for that, as this fix may be valid for 2.0 as well.

plugins/copr.py
```diff
@@ -104,9 +104,9 @@ def configure(self):
             if copr_plugin_config.has_option('main', 'distribution') and copr_plugin_config.has_option('main', 'releasever'):
                 distribution = copr_plugin_config.get('main', 'distribution')
                 releasever = copr_plugin_config.get('main', 'releasever')
-                self.chroot_config = [distribution, releasever]
+                CoprCommand.chroot_config = [distribution, releasever]
```
ignatenkobrain: it should stay as it is.

ignatenkobrain:
actually you should change `cls.chroot_config` to `self.chroot_config` in `_guess_chroot()` ir pass that parameter.

xenithorb:
> actually you should change cls.chroot_config to self.chroot_config in _guess_chroot() ir pass that parameter.

That doesn't work

ignatenkobrain:
@xenithorb it does, but you should remove `@classmethod` and change `cls` to `self`. Though @FrostyX or @xsuchy can argue.

Just checked code. Honestly I prefer if you will change `@classmethod` to `@staticmethod` and instead of passing `cls` you will pass `chroot_config`

ignatenkobrain:
@rpm-software-management/dnf-team thoughts?

xenithorb:
Both suggestions work once I understood the context. 

I have pushed again, is this what you were talking about?

plugins/copr.py
```diff
@@ -252,11 +252,11 @@ def _need_root(cls):
             raise dnf.exceptions.Error(
                 _('This command has to be run under the root user.'))
 
-    @classmethod
-    def _guess_chroot(cls):
+    @staticmethod
+    def _guess_chroot(self):
```
ignatenkobrain: just replace `self` with `chroot_config` here and below and looks good to me

PR #189, opened 2016-07-28
cls.chroot_config inside _guess_chroot returns None (RhBug: 1361003)

xenithorb:
Redo of failed PR rpm-software-management/dnf-plugins-core#188 which was against master instead of dnf-1.1 branch. 

@Conan-Kudo 
@ignatenkobrain 

-----

##### Original Post:

1. L#259: cls.chroot_config returns None
2. L#107: self.chroot_config is defined properly but is not passed to
_guess_chroot()
   - self changed to class identifier CoprCommand - works

Disclaimer: I am not a Python developer and I merely stumbled on a solution that worked for me. I verified with `pdb` that the necessary values were being passed correctly before submitting the patch. I have no idea why the original code doesn't work versus specifying the class name. 

My personal use case is Fedy on Korora. Fedy relies on some COPR repositories, and it is completely broken right now (that is, all COPR repos on any system that needs to use `/etc/dnf/plugins/copr.conf` don't currently work because `dnf copr` returns that it can not find any builds

```
Error: This repository does not have any builds yet so you cannot enable it now. 
```

https://bugzilla.redhat.com/show_bug.cgi?id=1361003

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -136,7 +136,7 @@ class CoprCommand(dnf.cli.Command):
         try:
             chroot = extcmds[2]
         except IndexError:
-            chroot = self._guess_chroot()
+            chroot = self._guess_chroot(self.chroot_config)
 
         # commands without defined copr_username/copr_projectname
         if subcommand == "list":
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
merged as #193

Conan-Kudo:
@ignatenkobrain This patch still needs to go into dnf-plugins-core for `dnf-1.1`, since that is what's shipping in Fedora and Mageia today.

ignatenkobrain:
@Conan-Kudo we are hardly working to get 2.0 done in general and in Fedora. Unfortunately I don't have time to invest for backporting this patch into 1.1 (I mean in Fedora).

xenithorb:
I'm a bit confused. My original PR was against `master` and then I rebased it to be against `dnf-1.1` as instructed?

Conan-Kudo:
@xenithorb You needed two versions of the PR: one for `dnf-1.1` and one for master. They are two different codebases.

Conan-Kudo:
@ignatenkobrain Well, @xenithorb already did the necessary backporting work for you, so it's just a patch apply and a cherry pick for the dnf-plugins-core package in Fedora. Not much work at all.

PR #190, opened 2016-08-05
DNF-completion plugin print some info (RhBug:1322562)

stepasm:
Dnf-completion plugin will print info also to stdout when generating cache.

References:https://bugzilla.redhat.com/show_bug.cgi?id=1322562

### DIFF
--- a/plugins/generate_completion_cache.py
+++ b/plugins/generate_completion_cache.py
@@ -53,6 +53,7 @@ class BashCompletionCache(dnf.Plugin):
             try:
                 with sqlite3.connect(self.cache_file) as conn:
                     self._out('Generating completion cache...')
+                    logger.info('Generating completion cache for available packages.')
                     cur = conn.cursor()
                     cur.execute(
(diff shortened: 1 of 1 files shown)


stepasm:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;esYsPOXW&#x2F;480-stepan-bug-list"> Stepan bug list</a>

ignatenkobrain:
I definitely dont agree for info loglevel as this is debug.

stepasm:
I think that there should be given some feedback to user.

ignatenkobrain:
Not really. It's useless information for user.

stepasm:
If the user needs to wait he should know that something is happening.

ignatenkobrain:
user does not care about such things. For him program is still not finished.

ignatenkobrain:
This should be done in DNF itself.

Before each hook it should print to stdout list of plugins which will be executed and into log when and which plugin was executed.

PR #191, opened 2016-08-12
builddep: adjust to new config (dnf-2.0)

mluscon:
The builddep plugin stopped working after 1679f9f6.


ignatenkobrain:
dunno why we still add unittests, but looks nice!

@dnf-bot r+

PR #192, opened 2016-08-15
dnf dowload --resolve should download everytime requested packages (RhBug:1276611)

stepasm:
dnf download --resolve should download packages that are already installed with lower version and are marked as upgrade

References: https://bugzilla.redhat.com/show_bug.cgi?id=1276611

### DIFF
--- a/doc/download.rst
+++ b/doc/download.rst
@@ -58,6 +58,9 @@ Options
 ``--urlprotocol``
     Limit the protocol of the urls output by the --url option. Options are http, https, rsync, ftp.
 
+``--resolve``
+    Resolves dependencies of specified packages and downloads missing dependencies in the system.
+
 --------
 Examples
 --------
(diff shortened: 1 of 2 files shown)


stepasm:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;esYsPOXW&#x2F;480-stepan-bug-list"> Stepan bug list</a>

ignatenkobrain:
`--resolve` is broken by design. It should download all packages except minimal environment.

j-mracek:
@stepasm If I good understand the bug report and documentation, the patch do not solve the problem. The problem is that I have on system acpi package installed. If I use command ``dnf download acpi`` it downloads package acpi.rpm. But with same command + ``--resolve`` it doesn't download anything and reporter want to download acpi.rpm + unsatisfied dependency on your system (in this case it mean nothing because all packages required by ``acpi``` have been installed).

stepasm:
@j-mracek Now will be downloaded specified packages every time and dependencies from list of  installs and upgrades - unsatisfied dependencies.

j-mracek:
Thanks a lot, code looks fine

j-mracek:
@dnf-bot r+

j-mracek:
@dnf-bot retry

PR #197, opened 2017-01-05
Add new sub-package dnf-utils

j-mracek:
As a Yum-utils CLI compatibility layer, supplies in /usr/bin/ builddep,
repoquery, repoclosure, download, debuginfo-install, and config-manager and
redirecting to DNF.

Requires: https://github.com/rpm-software-management/dnf-plugins-core/pull/199
Requires fedora release of yum-utils-1.1.32 with deprecated binaries - PR https://github.com/rpm-software-management/yum-utils/pull/16
After merge of the required PR the presented patch can be fully tested.


MichaelMraka:
retest

dnf-plugins-core.spec
```diff
@@ -86,6 +86,52 @@ reposync commands. Additionally provides generate_completion_cache and noroot
 passive plugins.
 %endif
 
+%package -n dnf-utils
+Conflicts:      yum-utils < 1.1.32
+Requires:       dnf >= %{dnf_lowest_compatible}
+Requires:       %{name} = %{version}-%{release}
+%if %{with python3}
+Requires:       python3-dnf-utils = %{version}-%{release}
```
ignatenkobrain: it doesn't really require it...

dnf-plugins-core.spec
```diff
@@ -86,6 +86,52 @@ reposync commands. Additionally provides generate_completion_cache and noroot
...
+Requires:       python2-dnf-utils = %{version}-%{release}
+Requires:       python2-dnf-plugins-extras-repoclosure
+%endif
+Summary:        Yum-utils CLI compatibility layer
+
+%description -n dnf-utils
+As a Yum-utils CLI compatibility layer, supplies in /usr/bin/ builddep, repoquery, repoclosure,
+download, debuginfo-install, and config-manager and redirecting to DNF.
+
+%package -n python2-dnf-utils
```
ignatenkobrain: I would prefer to not pollute repomd with such useless packages as they don't provide any python modules.

j-mracek: python-2 python-3 packages solve a problem if user has python-2 or python-3 version on system.

ignatenkobrain:
and actually I'm wondering why we should introduce subpackages, why don't we put this into main package...

dnf-plugins-core.spec
j-mracek: This is work-eround how to allow:
(python3-dnf, python3-dnf-utils) -> repoquery - runs with python3
(python2-dnf, python2-dnf-utils) -> repoquery - runs with python2
(python2-dnf, python2-dnf-utils, python3-dnf, python3-dnf-utils)) -> repoquery - runs with python3
The PR shadows the solution of dnf how to redirect dnf to proper python version (dnf-2, dnf-3).

j-mracek:
@ignatenkobrain The changes introduce conflict with yum-utils, and people probably use that. Still you can install yum package, but you call yum by yum-deprecated.

j-mracek:
test this please

Conan-Kudo:
@j-mracek It sounds like yum-utils should be altered to have their utils suffixed with deprecated, like yum was.

j-mracek:
@Conan-Kudo Good point, thanks.

MichaelMraka:
Please reference this bug in commit message https://bugzilla.redhat.com/show_bug.cgi?id=1381917.

j-mracek:
We should support all of these:
/usr/bin/debuginfo-install
/usr/bin/find-repos-of-install
/usr/bin/needs-restarting
/usr/bin/package-cleanup
/usr/bin/repo-graph
/usr/bin/repo-rss
/usr/bin/repoclosure
/usr/bin/repodiff
/usr/bin/repomanage
/usr/bin/repoquery
/usr/bin/reposync
/usr/bin/repotrack
/usr/bin/show-changed-rco
/usr/bin/show-installed
/usr/bin/verifytree
/usr/bin/yum-builddep
/usr/bin/yum-config-manager
/usr/bin/yum-debug-dump
/usr/bin/yum-debug-restore
/usr/bin/yum-groups-manager
/usr/bin/yumdownloader

j-mracek:
The plan: 
Movement of some of extras to core.
Rename yum-utils binary as deprecated.
Keep conflict with dnf-utils < yum-utils depricated

j-mracek:
I added blocked, because after last changes in upstream, I have to retest the PR again.

j-mracek:
@Conan-Kudo What you request I already tried, but I don't like that on system I will have an executable that will be impossible to use. Even showing help of dnf in case of using call snd-utils could be confusing.

jsilhan:
I understand that having `repoquery` containing the core functionality would be minimalist approach however "easier to see what's going on" approach would be having `dnf-util` containing the script. What about installing `dnf-utils` outside of PATH?

Conan-Kudo:
@jsilhan That actually seems like the reasonable approach. `/usr/libexec/dnf-utils/dnf-util` could be the actual script while all the other things are symlinks in `/usr/bin`.

ignatenkobrain:
okay, there is revised PR for this, #218.

PR #198, opened 2017-01-08
Add --url option back

dustymabe:
Resolves: rhbz#1250115


plugins/download.py
```diff
@@ -56,6 +57,9 @@ def set_argparser(parser):
                             help=_('download path, default is current dir'))
         parser.add_argument('--resolve', action='store_true',
                             help=_('resolve and download needed dependencies'))
+        parser.add_argument('--url', action='store_true',
+                            help=_('just list http(s) urls where the rpms can '
```
ignatenkobrain: I think it's better to `s/just/print/`

plugins/download.py
```diff
@@ -128,9 +146,34 @@ def _download_debuginfo(self, pkg_specs):
...
+            # have a mirrorlist we can then just choose one of the mirrors
+            # and merge the mirror url with the relative location of the
+            # package on the mirror.
+            for pkg in pkgs:
+                h = pkg.repo._get_handle()
+                h.fetchmirrors = True  # sets LRO_FETCHMIRRORS
+                h.destdir = tmpdir     # store tmp files in tmpdir
+                h.perform()
+                for mirror in h.mirrors:
+                    if mirror.startswith('https'):
```
ignatenkobrain: hm, what about ftp or http?

dustymabe: oops, I originally had this as http, not https. but I decided to try out https and forgot to change it back. Will fix it to http.

My thoughts about ftp and rsync is that they are generally less useful than the http(s) ones. almost everything can use http(s). The alternative to this is that we give some option for the user to specify what protocol they want. I think that is overcomplicating it though. What do you think?

dustymabe:
@ignatenkobrain - tests should pass now

ignatenkobrain:
Looks good to me now, can you add it into documentation and squash commits?

dustymabe:
@ignatenkobrain - added docs and got it down to the smallest number of commits that I'd like to keep.

dustymabe:
@ignatenkobrain anything left to do on this?

ignatenkobrain:
@dnf-bot r+

j-mracek:
@dnf-bot retry

j-mracek:
@dnf-bot retry

PR #199, opened 2017-01-25
Reworked content of plugins in core and extras

j-mracek:
Requires: https://github.com/rpm-software-management/dnf-plugins-extras/pull/81
Requires: https://github.com/rpm-software-management/ci-dnf-stack/pull/241


dnf-plugins-core.spec
```diff
@@ -9,7 +10,8 @@
 %endif
 
 Name:           dnf-plugins-core
-Version:        1.0.0
+Epoch:          1
```
ignatenkobrain: please, no epochs...

dnf-plugins-core.spec
```diff
@@ -51,16 +59,25 @@ BuildRequires:  python2-devel
 Requires:       python2-dnf >= %{dnf_lowest_compatible}
 Requires:       python2-dnf < %{dnf_not_compatible}
 Requires:       python-hawkey >= %{hawkey_version}
+%if ! 0%{?fedora}
```
ignatenkobrain: any reasons for this?

j-mracek: To provide fedora users copr wit core

dnf-plugins-core.spec
j-mracek: Unfortunatelu some extras has provide with epoch 1 therefore I have to overwrite it with same epoch and higher version.

dnf-plugins-core.spec
ignatenkobrain: then you should remove `!`. but rather I would go with `%if %{undefined rhel}`

dnf-plugins-core.spec
ignatenkobrain: but this is needed only for Obsoletes/Provides, not for dnf-plugins-core themselves

dnf-plugins-core.spec
j-mracek: @ignatenkobrain Thanks, but what exactly ```%if %{undefined rhel}``` does?

MichaelMraka:
This is a big change in packaging so I'd like everybody to take a look.

dnf-plugins-core.spec
```diff
@@ -74,18 +102,168 @@ BuildRequires:  python3-sphinx
...
-Core Plugins for DNF. This package enhances DNF with builddep, config-manager,
-copr, debuginfo-install, download, needs-restarting, repoquery and
-reposync commands. Additionally provides generate_completion_cache and noroot
-passive plugins.
+Core Plugins for DNF. This package enhances DNF with builddep, config-manager, debuginfo-install,
+download, needs-restarting, repoclosure, repograph, repomanage and reposync commands. Additionally
+provides generate_completion_cache and noroot passive plugins.
 %endif
 
+%package -n python2-dnf-plugin-leaves
```
jsilhan: is dnf-plugin-leaves alternative to yum-plugin-show-leaves (1:1)?

jsilhan: oh I see, show-leaves is alternative to yum-plugin-show-leaves. So is dnf-plugin-leaves alternative to yum-plugin-remove-with-leaves then?

ignatenkobrain: the plugin(s?) (in extras) implement both, remove-with-leaves and show-leaves.

Although none of them work correctly.

dnf-plugins-core.spec
```diff
@@ -74,18 +102,168 @@ BuildRequires:  python3-sphinx
...
+Provides:       python2-dnf-plugins-extras-migrate = %{epoch}:%{version}-%{release}
+Provides:       dnf-command(migrate)
+Provides:       dnf-plugins-extras-migrate = %{epoch}:%{version}-%{release}
+Conflicts:      dnf-plugins-extras-common-data < %{dnf_plugins_extra}
+Obsoletes:      python2-dnf-plugins-extras-migrate < %{dnf_plugins_extra}
+
+%description -n python2-dnf-plugin-migrate
+Migrate Plugin for DNF, Python 2 version. Migrates history, group and yumdb data from yum to dnf.
+
+%package -n python2-dnf-plugin-show-leaves
```
ignatenkobrain: Probably we want to name it dnf-plugin-leaves and add provides for both `remove-with-leaves` and `show-leaves`

doc/index.rst
```diff
@@ -27,13 +27,20 @@ This documents core plugins of DNF:
    release_notes
    builddep
    config_manager
-   copr
```
ignatenkobrain: yum-utils have this plugin in `yum-plugin-copr` package

dnf-plugins-core.spec
jsilhan: I also would like to get rid of epoch bump and go with igor's approach.

jsilhan:
I am fine with the concept and reordering of plugins. Just the epoch bump concerns me.

dnf-plugins-core.spec
```diff
@@ -74,18 +102,168 @@ BuildRequires:  python3-sphinx
...
+Local Plugin for DNF, Python 2 version. Automatically copy all downloaded packages to a
+repository on the local filesystem and generating repo metadata.
+
+%package -n python3-dnf-plugin-local
+Summary:        Local Plugin for DNF
+Requires:       %{_bindir}/createrepo_c
+Requires:       python3-%{name} = %{epoch}:%{version}-%{release}
+Provides:       dnf-plugin-local =  %{epoch}:%{version}-%{release}
+Provides:       python3-dnf-plugins-extras-local = %{epoch}:%{version}-%{release}
+Provides:       dnf-command(local)
```
ignatenkobrain: I don't think there's command `local`

dnf-plugins-core.spec
esmil: I don't care what the current leaves plugin ends up being called, but I would like to be able to install only the current leaves plugin and not have to run it through the show-leaves plugin.

dnf-plugins-core.spec
MichaelMraka: Epoch is not nice but it's much easier to maintain upgrades if all packages have the same evr rather than some having provides with different epoch. Bumping the epoch in the past was a mistake but now it's too late. Let's not complicate things again.

MichaelMraka: Just to make my note clear:  I support Jarda's decision to bump epoch globally.

ignatenkobrain: but we need epoch only for extras (the obsoletes), let's not pollute core with epoch.

ignatenkobrain: moreover! we don't have epoch in dnf-plugins-extras, so thi is totally useless and jsut creating problems

MichaelMraka: We do:
```
$ grep 1: dnf-plugins-extras.spec
Provides:       dnf-plugin-tracer = 1:%{version}-%{release}
Provides:       dnf-plugin-versionlock = 1:%{version}-%{release}
```

ignatenkobrain: because it is part of merging some other plugins or such (at least with tracer, not sure about versionlock). Anyway, we can drop those epochs from such provides

j-mracek:
@jsilhan @ignatenkobrain @mhatina @jrohel Rebased, refactored, please take a second look.

jsilhan:
Guys, is this ready for merge? IIRC we already eliminated all concerns. There's no epoch bump and COPR will stay in plugins core. Are there any other issues or can we finally merge these three PRs https://github.com/rpm-software-management/ci-dnf-stack/pull/241 , https://github.com/rpm-software-management/dnf-plugins-extras/pull/81 ?

Conan-Kudo:
@jsilhan If those issues are resolved, then I'm fine with this.

j-mracek:
@ignatenkobrain Please what do you mean separate? Did you mean separate subpackage from core? I try to use same schematic as yum used.

PR #200, opened 2017-01-31
setup SideCI to ignore some PEP8 violations

j-mracek:
unfortunately tox.ini and sideci.yml has to be in the root directory.
SideCI doesn't currently support custom path to flake8 configuration file.


MichaelMraka:
@dnf-bot r+

PR #204, opened 2017-02-23
builddep: install requirements by provides (RhBug:1332830)

ignatenkobrain:
This way we will also get support for rich dependencies for free.
    
Reported-by: Neal Gompa <ngompa13@gmail.com>
References: https://bugzilla.redhat.com/show_bug.cgi?id=1332830


plugins/builddep.py
```diff
@@ -135,13 +135,13 @@ def _rpm_dep2reldep_str(rpm_dep):
...
     def _install(self, reldep_str):
-        try:
-            self.base.install(reldep_str)
-        except dnf.exceptions.MarkingError:
-            msg = _("No matching package to install: '%s'")
-            logger.warning(msg, reldep_str)
+        sltr = dnf.selector.Selector(self.base.sack)
+        sltr = sltr.set(provides=reldep_str)
+        if sltr.matches():
+            self.base._goal.install(select=sltr)
```
j-mracek: what about to add ```optional=(not self.base.conf.strict)```

Conan-Kudo: I'm not sure `strict` would necessarily be defined to act in the manner you suggest...

ignatenkobrain: Builddep is always strict

j-mracek: Then ```optional=False``` but I think it is default behavior of libdnf

jsilhan:
LGTM with libdnf change. I would expect some tests though.

MichaelMraka:
retest this please

ignatenkobrain:
yes, tests are currently broken (because now we use selector which is not mocked). I will fix it at some point soon.

MichaelMraka:
retest this please

plugins/builddep.py
```diff
@@ -135,13 +135,15 @@ def _rpm_dep2reldep_str(rpm_dep):
         return rpm_dep.DNEVR()[2:]
 
     def _install(self, reldep_str):
-        try:
-            self.base.install(reldep_str)
-        except dnf.exceptions.MarkingError:
+        sltr = dnf.selector.Selector(self.base.sack)
+        sltr = sltr.set(provides=reldep_str)
```
jsilhan: this seems clean but shouldn't it consider `sltr.set(file=reldep_str)` if no such provide exists?

j-mracek: Good point, then it fails when file provide will be required but it will be not listet in provides.

ignatenkobrain: @jsilhan aren't files are also in provides?

At least it works here when I have spec with `BuildRequires: %{_bindir}/cmake`

Conan-Kudo: I think not all of them are, just the ones in `primary.xml.gz` (the synthesis).

ignatenkobrain: hmm, looks like it didn't work now.. will check tomorrow

ignatenkobrain: @jsilhan @j-mracek wdyt?

```diff
diff --git a/plugins/builddep.py b/plugins/builddep.py
index f615103..fb99f26 100644
--- a/plugins/builddep.py
+++ b/plugins/builddep.py
@@ -135,10 +135,19 @@ class BuildDepCommand(dnf.cli.Command):
         return rpm_dep.DNEVR()[2:]
 
     def _install(self, reldep_str):
-        sltr = dnf.selector.Selector(self.base.sack)
-        sltr = sltr.set(provides=reldep_str)
-        if sltr.matches():
-            self.base._goal.install(select=sltr)
+        sltr_prov = dnf.selector.Selector(self.base.sack)
+        sltr_prov.set(provides=reldep_str)
+        sltr_file = dnf.selector.Selector(self.base.sack)
+        sltr_file.set(file=reldep_str)
+        sltrs = [sltr_prov, sltr_file]
+        for sltr in sltrs:
+            if not sltr.matches():
+                continue
+            already_inst = self.base._sltr_matches_installed(sltr)
+            if already_inst:
+                for package in already_inst:
+                    dnf.base._msg_installed(package)
+            self.base._goal.install(select=sltr, optional=False)
             return True
         else:
             msg = _("No matching package to install: '%s'")
```

ignatenkobrain:
@j-mracek tests are still broken, but feel free to review code itself in meanwhile

ignatenkobrain:
since test would need a lot of changes including implementing SackStub, PackageStub and probably SelectorStub it should be better to just replace it with functional test: https://github.com/rpm-software-management/ci-dnf-stack/pull/271

j-mracek:
Code looks fine, but replacement of unitest by ci-tests is inadequate. I think that test cases has to be enlarged. We should test all options like requires file provide, normal dep and now only implemented richdep. There should be also packages available that should be not touched during the test (negative control).

ignatenkobrain:
@j-mracek are you volunteering to fix test?

j-mracek:
@ignatenkobrain Actually I am in the middle of refactoring of patch for lead cached RPM.

plugins/builddep.py
```diff
@@ -135,21 +135,29 @@ def _rpm_dep2reldep_str(rpm_dep):
...
+            sltrs.append(sltr_files)
+
+        for sltr in sltrs:
+            if not sltr.matches():
+                continue
+            already_inst = self.base._sltr_matches_installed(sltr)
+            if already_inst:
+                for package in already_inst:
+                    dnf.base._msg_installed(package)
+            self.base._goal.install(select=sltr, optional=False)
```
j-mracek: Sorry Igor, but last changes I really don't like. The new construct could result in install of multiple packages that will have a file provide. By that way you can create two goals for single requires, that mean that it could result in conflict in solver.

ignatenkobrain: That's how it is done in DNF itself...

plugins/builddep.py
```diff
@@ -135,13 +135,29 @@ def _rpm_dep2reldep_str(rpm_dep):
         return rpm_dep.DNEVR()[2:]
 
     def _install(self, reldep_str):
-        try:
-            self.base.install(reldep_str)
-        except dnf.exceptions.MarkingError:
+        sltr_provides = dnf.selector.Selector(self.base.sack)
+        sltr_provides.set(provides=reldep_str)
+        sltrs = [sltr_provides]
```
j-mracek: I don't like the construct with list of selectors. This is not good idea when you can end up with more than one goal per provide.

plugins/builddep.py
j-mracek: Ok, it will work, but the approach is quit difficult to understand. I think you have to change work-flow, where you will test if selector is valid at the time of creation, or add proper description. The fact that search for validity of selector is done here ('if not sltr.matches():') and this is place where cpu is used is hard to understand. 
Also I have doubt about else: on line 157

ignatenkobrain:
@j-mracek fixed statements flow.

ignatenkobrain:
retest this please

ignatenkobrain:
retest this please

ignatenkobrain:
retest this please

plugins/builddep.py
```diff
@@ -135,14 +135,30 @@ def _rpm_dep2reldep_str(rpm_dep):
...
             return False
+
+        already_inst = self.base._sltr_matches_installed(sltr)
+        if already_inst:
+            for package in already_inst:
+                dnf.base._msg_installed(package)
+        self.base._goal.install(select=sltr, optional=False)
         return True
 
+
```
j-mracek: I think that this empty line is not required

j-mracek:
It looks nice, thanks

j-mracek:
@dnf-bot r+

ignatenkobrain:
@dnf-bot clean

ignatenkobrain:
@dnf-bot retry

ignatenkobrain:
@dnf-bot clean

ignatenkobrain:
looks like jenkins doesn't process that commit anymore, so rebased @dnf-bot r=jmracek

PR #205, opened 2017-03-10
Po update

ignatenkobrain:


ignatenkobrain:
@j-mracek since @jrohel is on vaction and we need this ASAP, assigning to you :P

po/zanata.xml
```diff
@@ -4,101 +4,4 @@
   <project>dnf-plugins-core</project>
   <project-version>master</project-version>
   <project-type>gettext</project-type>
-  <locales>
```
j-mracek: @ignatenkobrain By that way do you get translation for all available languages?

j-mracek:
Because I cannot test the functionality please if it works merge it.

po/zanata.xml
ignatenkobrain: yes, now I get translations for **all** languages instead of specified ones.

PR #208, opened 2017-03-15
copr: Use python-distro for distro identification

Conan-Kudo:
`platform.linux_distribution()` is slated for removal in Python 3.7.

The "distro" Python module provides a suitable replacement.

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -76,6 +76,9 @@ BuildRequires:  python-nose
 BuildRequires:  python2-nose
 %endif
 BuildRequires:  python2-devel
+%if 0%{?fedora}
+Requires:       python2-distro
+%endif
 Requires:       python2-dnf >= %{dnf_lowest_compatible}
 Requires:       python2-dnf < %{dnf_not_compatible}
 Requires:       python2-hawkey >= %{hawkey_version}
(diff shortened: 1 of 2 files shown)


xsuchy:
👍

Conan-Kudo:
@ignatenkobrain Why is this blocked?

xsuchy:
@Conan-Kudo Because this package is part of RHEL and of course will be part of next RHEL. So merge of this PR will mean that we will have to add python-distro to RHEL.
We may postpone merge of this PR for some time.

dmach:
@Conan-Kudo the change is straightforward and very simple.
Could you resolve the conflicts?
When you're done with it, we can merge this PR.

dmach:
@Conan-Kudo actually @xsuchy was right about the availability of the python modules in RHEL.
It's better to avoid adding new packages to the distro.

I suppose we need something like this:
```
try:
    import platform
except ImportError:
    import distro as platform
```
(or vice-versa)

Conan-Kudo:
@dmach Rebasing now. One moment. :)

Conan-Kudo:
@dmach @xsuchy @ignatenkobrain I've rebased it and set it up to fall back to the deprecated `platform` module when `distro` is not available.

Conan-Kudo:
@dmach ping?

clime:
It would be good if dnf itself provided information whether we are on rawhide or not so we could read it out from dnf core in the plugin.

Related to this is issue https://pagure.io/copr/copr/issue/267 - it would be cool if there was a variable (e.g. `$rawver`) available in .repo file that would return 'rawhide' on Rawhide and $releasever otherwise.

Would this be possible, please? If not, then I think this can be merged (seems it is not blocked by anything for quite some time) but having that variable would solve more problems than just this one.

Conan-Kudo:
@clime This would be hard, and not necessarily predictable across distro families.

Conan-Kudo:
@xsuchy @dmach I don't believe there's a reason for this to be blocked from being merged any longer. I fall back gracefully to `platform` when `distro` isn't available for import, so it shouldn't mess up RHEL/CentOS 7.

xsuchy:
@Conan-Kudo  True. Can you please update the PR so that requires of python3-distro is inside of `%if 0%{?fedora}`? Or specify the condition that way that this requires is not used on RHEL (any version).

Conan-Kudo:
@xsuchy You don't expect `distro` to be in EL8? I'd be surprised if it isn't, since `linux_distribution()` is gone in Python 3.7.

Conan-Kudo:
@xsuchy Well, then the `copr` plugin will fail even on EL8 if it has Python 3.7, as we need _something_ for distribution identification...

xsuchy:
What about:
```
except ImportError:
    try:
       from platform import linux_distribution as __linux_distro
    except Import Error:
         def __linux_distro():
             return ("RedHat Enterprise Linux", None, None)
```
I.e. when the module platform and module distro are not available, it will always fake to RHEL.

Conan-Kudo:
@xsuchy Sounds like a plan.

Conan-Kudo:
@xsuchy So how does this look, instead?

PR #210, opened 2017-04-04
dont run versionlock on non-transactional operations

jsilhan:
request from user on fedora devel list

PR https://github.com/rpm-software-management/dnf-plugins-extras/pull/90 submitted for dnf-plugins-core

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -59,7 +59,8 @@ class VersionLock(dnf.Plugin):
 
     def sack(self):
         if not self.cli.demands.resolving:
-            raise dnf.exceptions.Error(NO_VERSIONLOCK)
+            logger.info(NO_VERSIONLOCK)
+            return
 
         if not locklist_fn:
             raise dnf.exceptions.Error(NO_LOCKLIST)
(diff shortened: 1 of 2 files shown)


plugins/versionlock.py
```diff
@@ -57,6 +58,9 @@ def config(self):
                        and cp.get('main', 'locklist'))
 
     def sack(self):
+        if not self.cli.demands.resolving:
```
j-mracek: This will terminate any non-transactional operation for dnf. That means, that dnf repoquery, repolist, info ... will be terminated.

jsilhan: changed to `return`

j-mracek: @jsilhan Thanks it works, but I am not sure if message "Ignoring versionlock for non-transactional operation" cannot be improved. What about something like: "Excludes from Versionlock plugin were not applied". I am also not sure if the message should be only for debug, but in some cases it is better to see it every time.

j-mracek: Also the new behavior should be mentioned in documentation.

jsilhan: @j-mracek , right, the hackish implementation of versionlock by applying excludes should be hidden from user because by common sense he don't expect packages to be excluded at all, just have them locked in concrete version -> changed to debug, modified log message and documented.

j-mracek:
Thanks for your contribution

j-mracek:
@dnf-bot r+

PR #211, opened 2017-04-11
Check argument if it is a file ending with .rpm (RhBug:1436570)

j-mracek:
https://bugzilla.redhat.com/show_bug.cgi?id=1436570

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -219,7 +219,7 @@ class DownloadCommand(dnf.cli.Command):
 
     def _get_query(self, pkg_spec):
         """Return a query to match a pkg_spec."""
-        if os.path.exists(pkg_spec):
+        if pkg_spec.endswith('.rpm') and os.path.isfile(pkg_spec):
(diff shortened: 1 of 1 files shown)


plugins/download.py
```diff
@@ -219,7 +221,7 @@ def _get_source_packages(pkgs):
 
     def _get_query(self, pkg_spec):
         """Return a query to match a pkg_spec."""
-        if os.path.exists(pkg_spec):
+        if pkg_spec.endswith('.rpm') and os.path.exists(pkg_spec) and os.path.isfile(pkg_spec):
```
ignatenkobrain: why not `os.path.exists(pkg_spec) and os.path.isfile(pkg_spec)` -> `os.path.isfile(pkg_spec)`?

j-mracek: @ignatenkobrain Thanks, good point

ignatenkobrain:
@dnf-bot r+

PR #214, opened 2017-04-22
modularity support for `dnf copr`

clime:
Hello, this PR extends `dnf copr enable` and `dnf copr disable ` subcommands to also work with module repositories of a COPR project.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -164,6 +164,8 @@ class CoprCommand(dnf.cli.Command):
 
         repo_filename = "{}/_copr_{}-{}.repo" \
                         .format(self.base.conf.get_reposdir, copr_username, copr_projectname)
+        modules_repo_filename = "{}/_copr_modules_{}-{}.repo"\
+                                .format(self.base.conf.get_reposdir, copr_username, copr_projectname)
         if subcommand == "enable":
             self._need_root()
             msg = _("""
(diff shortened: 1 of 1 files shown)


plugins/copr.py
```diff
@@ -148,6 +148,8 @@ def run(self):
 
         repo_filename = "{}/_copr_{}-{}.repo" \
                         .format(self.base.conf.get_reposdir, copr_username, copr_projectname)
+        modules_repo_filename = "/etc/fm.modules.d/_copr_{}-{}.cfg" \
```
ignatenkobrain: the main problem here is that this directory doesn't exist on most of machines... either you should add some Requires or own directory or ...

clime: Would something like this be preferable over the current solution where /etc/fm.modules.d is dynamically created if it does not exist yet?

[0001-added-etc-fm.modules.d.patch.txt](https://github.com/rpm-software-management/dnf-plugins-core/files/951629/0001-added-etc-fm.modules.d.patch.txt)

(Please, consider this to be just a proof of concept)

ignatenkobrain: you could just own that directory, but question is still same... who should really own that directory and why **this** name?

clime: Perhaps this could apply? https://fedoraproject.org/wiki/Packaging:Guidelines#The_directory_is_owned_by_a_package_which_is_not_required_for_your_package_to_function. It would be multiple ownership then.

clime: Can you, please, review again?

clime:
Any progress here?

xsuchy:
@ignatenkobrain can you please review again?

mhatina:
@clime As this is POC, do you think it is worth rebasing and updating paths ignatenkobrain mentioned in review? For example `modules_repo_filename` to `/etc/yum.repos.d/`, since there is repo file at `$copr/coprs/$project_name/repo/modules/`.

clime:
> @clime As this is POC, do you think it is worth rebasing and updating paths ignatenkobrain mentioned in review? For example modules_repo_filename to /etc/yum.repos.d/, since there is repo file at $copr/coprs/$project_name/repo/modules/.

Ye, sure, can do that.

clime:
I updated the path of a stored repofile to be at /etc/yum.repos.d/. But the /repo/modules entry point (the collection of all present modules in a project) have disappeared from copr code, which means that this PR can probably be closed because we are not able to enable all the built modules per a given copr project (which is what `dnf copr enable` should do). @FrostyX. Any other idea here?

mhatina:
Closing. Explanation in comment above.

PR #216, opened 2017-04-25
Solve a problem in performance of versionlock (RhBug:1431493)

j-mracek:
https://bugzilla.redhat.com/show_bug.cgi?id=1431493

Requires: https://github.com/rpm-software-management/dnf/pull/796

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -60,7 +60,9 @@ class VersionLock(dnf.Plugin):
         if not locklist_fn:
             raise dnf.exceptions.Error(NO_LOCKLIST)
 
-        locked = set()
+        excludes_query = self.base.sack.query().filter(empty=True)
+        include_query = self.base.sack.query().filter(empty=True)
+
         for pat in _read_locklist():
             excl = False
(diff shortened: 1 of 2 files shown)


plugins/versionlock.py
```diff
@@ -60,26 +60,28 @@ def sack(self):
...
+
         for pat in _read_locklist():
             excl = False
             if pat and pat[0] == '!':
                 pat = pat[1:]
                 excl = True
 
             subj = dnf.subject.Subject(pat)
-            pkgs = subj.get_best_query(self.base.sack)
+            pkgs = subj.get_best_query(self.base.sack, with_nevra=True, with_provides=False,
```
ignatenkobrain: I'm not sure about disabling by provides here...

j-mracek: According to documentation, versionlock handles only packages by nevra. It excludes packages and not packages based by provides. I disable it because I think it can provide inconsistency in result. What do you think?

ignatenkobrain:
@dnf-bot r+

PR #217, opened 2017-04-28
Fix incorrect exclude of locked version in versionlock

j-mracek:
Now it will exclude only packages with same nevra.name but different evra.


ignatenkobrain:
@dnf-bot r+

ignatenkobrain:
@dnf-bot retry

PR #220, opened 2017-05-15
Download plugin: Move --destdir to dnf (RhBug:1279001)

hoschiCZ:
Because we can't define CLI option --destdir twice (once in download plugin, once in dnf), I moved the copying functionality to dnf and made download plugin use it.

This pull request and the [pull request for dnf](https://github.com/rpm-software-management/dnf/pull/810) have to be accepted at the same time, because one doesn't work without the other (breaks download plugin).

Because --destdir will be run under root, all downloaded (to another location) packages now have permissions set to 777 since most expected usecases involve a non-privileged user doing something with the files (probably moving).

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -57,8 +57,6 @@ class DownloadCommand(dnf.cli.Command):
         parser.add_argument("--arch", '--archlist', dest='arch', default=[],
                             action=OptionParser._SplitCallback, metavar='[arch]',
                             help=_("limit  the  query to packages of given architectures."))
-        parser.add_argument('--destdir',
-                            help=_('download path, default is current dir'))
         parser.add_argument('--resolve', action='store_true',
                             help=_('resolve and download needed dependencies'))
         parser.add_argument('--url', '--urls', action='store_true', dest='url',
@@ -92,6 +90,11 @@ class DownloadCommand(dnf.cli.Command):
         else:
             pkgs = self._get_pkg_objs_rpms(self.opts.packages)
 
+        if self.opts.destdir:
+            self.base.conf.destdir = self.opts.destdir
+        else:
+            self.base.conf.destdir = dnf.i18n.ucd(os.getcwd())
+
         # If user asked for just urls then print them and we're done
         if self.opts.url:
             for pkg in pkgs:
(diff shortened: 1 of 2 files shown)


plugins/download.py
```diff
@@ -248,16 +244,4 @@ def _get_query_source(self, pkg_spec):
...
-        """Copy the downloaded package to target, not move.
-           If package is from local repo it must not be deleted there
-           and download routines will remove them from cache automatically.
-        """
-        if not os.path.exists(target):
-            os.makedirs(target)
-        for pkg in locations:
-            shutil.copy(pkg, target)
-        return target
+        return q
```
j-mracek: @hoschiCZ Please solve the new line at the end of file :-)

hoschiCZ: @j-mracek Consider it done.

plugins/download.py
```diff
@@ -92,6 +90,11 @@ def run(self):
         else:
             pkgs = self._get_pkg_objs_rpms(self.opts.packages)
 
+        if self.opts.destdir:
+            self.conf.destdir = self.opts.destdir
```
j-mracek: AttributeError: 'DownloadCommand' object has no attribute 'conf'
Change line to: self.base.conf.destdir = self.opts.destdir

plugins/download.py
```diff
@@ -92,6 +90,11 @@ def run(self):
...
+        else:
+            self.conf.destdir = dnf.i18n.ucd(os.getcwd())
```
j-mracek: Same here: self.base.conf.destdir = dnf.i18n.ucd(os.getcwd())

mhatina:
@hoschiCZ Use `git commit --fixup` on original commit, please. Use `git rebase --autosquash --interactive HEAD~3` to squash commits and then `git commit --amend` to change commit message of original commit (bug # should be 1279001)

j-mracek:
@hoschiCZ Please can you rebase it. The upstream version was changed and please also require hibher version of dnf (2.5.2 or 2.6.0, just according your other PR)

j-mracek:
@hoschiCZ Please change lines in dnf-plugins-core.spec:
``%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 2.6.0}`` to ``%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 2.6.1}``

``Version: 2.1.2`` to ``Version: 2.1.3``

j-mracek:
Thanks, only changes in version blocks merge.

hoschiCZ:
Versions and version required bumped in both dnf-plugins-core.spec and dnf.spec.

j-mracek:
@hoschiCZ Thanks a lot.

PR #221, opened 2017-06-02
debuginfo-install: install only requested packages (RhBug:1322599)

jrohel:
Change of behavior of command 'debuginfo-install'. The dependencies based on the non-debug package have been added before. This code was removed because old solution installs a lot of useless packages. Gdb/whatsoever will tell you what to install.

### DIFF
--- a/plugins/debuginfo-install.py
+++ b/plugins/debuginfo-install.py
@@ -58,10 +58,8 @@ class DebuginfoInstallCommand(dnf.cli.Command):
     summary = _('install debuginfo packages')
 
     dbgdone = []
-    reqdone = []
     packages = None
     packages_available = None
-    packages_installed = None
 
     @staticmethod
     def set_argparser(parser):
@@ -78,7 +76,6 @@ class DebuginfoInstallCommand(dnf.cli.Command):
     def run(self):
         self.packages = self.base.sack.query()
         self.packages_available = self.packages.available()
-        self.packages_installed = self.packages.installed()
 
         for pkgspec in self.opts.package:
             for pkg in sorted(dnf.subject.Subject(pkgspec).get_best_query(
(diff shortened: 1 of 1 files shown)


jrohel:
I added option "--noadddeps". But i don't like this name. Some better idea?

ignatenkobrain:
please, make it default.. all tools must produce all debuginfo package which people need. This is how it used to work in yum btw.

Conan-Kudo:
Wait, we're not going to auto-install deps needed for debuginfo stuff? That seems.. broken.

jrohel:
## Explanation
Debug packages usually do not have dependencies. So 'debuginfo-install' adds dependencies based on the non-debug package with the same name. The new option disables this algorithm. But does not disable installing of dependencies of debug packages.

I think about default behavior.
1. Will install dependencies based on the non-debug package with the same name (current state, as yum) and add option '--noadddeps' (or --exactdeps) for disabling this.
2. Or will install only dependencies required by debug package and add '--alldeps' option for old behavior (for adding dependencies based on the non-debug package with the same name)

## Examples
**dnf debuginfo-install iftop**
_Installing:_
 glibc-debuginfo           x86_64    2.24-4.fc25               updates-debuginfo     11 M
 iftop-debuginfo           x86_64    1.0-0.14.pre4.fc25        updates-debuginfo    108 k
 libpcap-debuginfo         x86_64    14:1.7.4-2.fc24           fedora-debuginfo     441 k
 ncurses-debuginfo         x86_64    6.0-6.20160709.fc25       fedora-debuginfo     1.5 M
_Installing dependencies:_
 glibc-debuginfo-common    x86_64    2.24-4.fc25               updates-debuginfo     13 M
**Install  5 Packages**
Total download size: 25 M
Installed size: 156 M

If we add '--noadddeps' option, only 'iftop-debuginfo' will be installed
**dnf debuginfo-install --noadddeps iftop**
_Installing:_
 iftop-debuginfo      x86_64      1.0-0.14.pre4.fc25         updates-debuginfo      108 k
**Install  1 Package**
Total download size: 108 k
Installed size: 330 k

But if we install 'glibc-debuginfo' then 'glibc-debuginfo-common' will be installed too. Because this is dependency of 'glibc-debuginfo' 
**dnf debuginfo-install --noadddeps glibc**
_Installing:_
 glibc-debuginfo             x86_64      2.24-4.fc25         updates-debuginfo       11 M
_Installing dependencies:_
 glibc-debuginfo-common      x86_64      2.24-4.fc25         updates-debuginfo       13 M

Install  2 Packages
Total download size: 23 M
Installed size: 146 M

ignatenkobrain:
@Conan-Kudo this is just broken by design and we should remove that. gdb/whatsoever will tell you what to install.. Now it is slow and doesn't really give you anything useful.

ignatenkobrain:
I have to admit that few years ago when I was writing this plugin I made this error.

jrohel:
@ignatenkobrain I think that you are right. So I removed old code as you suggested.

ignatenkobrain:
@dnf-bot r+

PR #224, opened 2017-06-16
Unify user confirmation in copr with dnf itself (RhBug:1436321)

j-mracek:
It use same methods like dnf itself, therefore no additional problem could be
there.

https://bugzilla.redhat.com/show_bug.cgi?id=1436321

Requires: https://github.com/rpm-software-management/dnf/pull/841


ignatenkobrain:
@dnf-bot r+

j-mracek:
@dnf-bot abbort

PR #226, opened 2017-07-13
copr: added list command to show enabled Copr repositories

dturecek:
Adds ``dnf copr list`` subcommand to show enabled Copr repositories. Changes the original ``dnf copr list name`` subcommand (which lists available Copr repositories for a given ``name``) to ``dnf copr list-user name``.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -204,20 +204,15 @@ Do you want to continue?""")
                     parser.read(directory + '/' + file_name)
 
         for copr in parser.sections():
-            if parser.getboolean(copr, "enabled"):
-                if disabled_only:
-                    pass
-                else:
-                    copr_name = copr.split('-', 1)
-                    msg = copr_name[0] + '/' + copr_name[1]
-                    print(msg)
-            elif not parser.getboolean(copr, "enabled"):
-                if enabled_only:
-                    pass
-                else:
-                    copr_name = copr.split('-', 1)
-                    msg = copr_name[0] + '/' + copr_name[1] + " (disabled)"
-                    print(msg)
+            enabled = parser.getboolean(copr, "enabled")
+            if (enabled and disabled_only) or (not enabled and enabled_only):
+               continue
+
+            copr_name = copr.split('-', 1)
+            msg = copr_name[0] + '/' + copr_name[1]
+            if not enabled:
+                msg += " (disabled)"
+            print(msg)
 
     def _list_user_projects(self, user_name):
         # http://copr.fedorainfracloud.org/api/coprs/ignatenkobrain/
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -31,6 +31,7 @@ import platform
 import shutil
 import stat
 import rpm
+import re
 
 PLUGIN_CONF = 'copr'
 
@@ -55,7 +56,8 @@ class CoprCommand(dnf.cli.Command):
   enable name/project [chroot]
   disable name/project
   remove name/project
-  list name
+  list
+  list-user name
   search project
 
   Examples:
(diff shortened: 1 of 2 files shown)


xsuchy:
@Conan-Kudo This PR is trying to address https://bugzilla.redhat.com/show_bug.cgi?id=1222562

> then make it match how config-manager works

Can you elaborate how you can list enabled repos with `config-manger`?

plugins/copr.py
```diff
@@ -174,6 +180,21 @@ def run(self):
             raise dnf.exceptions.Error(
                 _('Unknown subcommand {}.').format(subcommand))
 
+    def _list_installed_repositories(self):
+        self._list_repos_from_directory("/etc/yum.repos.d")
+        self._list_repos_from_directory("/etc/dnf.repos.d")
```
xsuchy: You should not list from both directories. Only from that one, which is being used. The path should be stored in `self.base.conf.reposdir`.

plugins/copr.py
```diff
@@ -174,6 +180,21 @@ def run(self):
...
+    def _list_repos_from_directory(self, directory):
+        for root, dir, files in os.walk(directory):
+            for file_name in files:
+                if re.match("^_copr_", file_name):
+                    repo_file = open(directory + "/" + file_name, "r").read()
+                    copr_repo = re.search("^\[(.*?)-(.*)\]", repo_file)
+                    msg = copr_repo.group(1) + "/" + copr_repo.group(2)
+                    if re.search("enabled=0", repo_file):
+                        msg = msg + " (disabled)"
+                    print(msg)
```
xsuchy: It will be much better to read the content of the file using ConfigParser

Conan-Kudo:
@xsuchy Hmm, I guess you can't. But `dnf repolist` shows you what repos you have enabled already. I don't like the idea of changing the name of subcommands unless there's a good reason for it.

Assuming this is a good reason, the `list-user` subcommand name is stupid and non-obvious. Maybe `list-available`?

Or perhaps the `list` subcommand can be reworked to support all this functionality using options.

xsuchy:
But `dnf repolist` is in main dnf. I do not think it is good idea to spread copr functionality over two distinct parts.

Well we can keep semantics of `list` and name the new command `list-installed` or `list-enabled`. But that will be IMHO awkward. I think it is better to change the semantics now rather that insist on previous bad naming.

Just think about it. What would you think 'dnf copr list` will do if you never ever used dnf copr plugin before?

Conan-Kudo:
I would think of the COPR repos configured on the system (either enabled or disabled). It's a fair argument. But What I'm saying is that all of the variations of listing things could simply be options.

Like `list --installed`, `list --enabled`, `list --disabled`, and `list --available-by-user=<foo>`, etc.

xsuchy:
> Like list --installed, list --enabled, list --disabled, and list --available-by-user=<foo>, etc.

That sounds reasonable.

dturecek:
I've changed it so it works like this

    dnf copr list --installed
    dnf copr list --enabled
    dnf copr list --disabled
    dnf copr list --available-by-user=NAME

with `--installed` option being default.

plugins/copr.py
```diff
@@ -174,6 +195,30 @@ def run(self):
...
+    def _list_installed_repositories(self, directory, enabled_only, disabled_only):
+        parser = ConfigParser()
+
+        for root, dir, files in os.walk(directory):
+            for file_name in files:
+                if re.match("^_copr_", file_name):
+                    parser.read(directory + '/' + file_name)
+
+        for copr in parser.sections():
+            if parser.getboolean(copr, "enabled"):
```
mhatina:
```
enabled = parser.getboolean(copr, "enabled")
if (enabled and disabled_only) or
        (not enabled and enabled_only):
   continue

copr_name = copr.split('-', 1)
msg = copr_name[0] + '/' + copr_name[1]
if not enabled:
    msg += " (disabled)"
print(msg)
```

What do you think about this solution?

dturecek: I agree this is a better way to do it. Thank you.

PR #227, opened 2017-07-21
Solve conflict with migrate plugin (RhBug:1470843)

j-mracek:
It produce a problem with upgrade from fc25 to fc26.

https://bugzilla.redhat.com/show_bug.cgi?id=1470843


ignatenkobrain:
@dnf-bot r+

PR #228, opened 2017-07-27
Rename --destdir to --downloaddir in doc

j-mracek:
This will unify documentation of options with dnf core where downloaddir option
is described. Both options remain as aliases, but it should provide better user
experiences.


j-mracek:
It is related to https://github.com/rpm-software-management/dnf/pull/889

PR #229, opened 2017-08-01
Fix repoclosure --check interaction with --best

pghmcfc:
This is a follow-up of rpm-software-management/dnf-plugins-extras/pull/50

With `--check`, we want to check repo closure for all packages in the specified repositories.

With `--best`, we want to only use the latest versions of each package across all repos.

When using both options, we need to enumerate the packages to be checked from the `--check` repos before reducing the package set to the latest versions for `--best`, as that might eliminate some of the packages from the `--check` repos. The existing code already inhibits using `available.latest()` early on for `--best` when using `--check`, and this commit does the deferred application of `available.latest()` for the case when `--check` and `--best` are used together.

I would like to write a test case for this but can't get my head around how to do it. It really needs two repos, and the existing tests only seem to be using `@commandline`. The main `dnf` test code appears to have infrastructure for mock repos with dummy packages and dependencies but I'm afraid I can't figure out how it works or how much code would need to be copied from there to do what I want. Maybe at some point in the future there could be some common test infrastructure packaged with `dnf` that plugins could use?

### DIFF
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -94,6 +94,9 @@ class RepoClosureCommand(dnf.cli.Command):
                 for pkgs_filtered in available.filter(reponame=repo):
                     checkpkgs.add(pkgs_filtered)
             pkgs.intersection_update(checkpkgs)
+            # --best not applied earlier due to --check, so do it now
+            if self.base.conf.best:
+                available = available.latest()
 
         for pkg in pkgs:
             unresolved[pkg] = set()


plugins/repoclosure.py
```diff
@@ -92,6 +92,10 @@ def _get_unresolved(self, arch=None):
                 for pkgs_filtered in available.filter(reponame=repo):
                     checkpkgs.add(pkgs_filtered)
             pkgs.intersection_update(checkpkgs)
+            # --best not applied earlier due to --check, so do it now
+            if self.base.conf.best:
+                available = available.latest()
+                available.run()
```
j-mracek: the line here ``available.run()`` is completely  not necessary. it returns list of packages from query. Why did you used it here?

pghmcfc: I wanted to enumerate the package list, which doesn't happen until either you iterate over the query or use run(). The package list needs to be enumerated by the time unresolved_deps is calculated, and it wouldn't happen without the run().

j-mracek: I guess that ``available = available.latest().apply()`` would be better.

j-mracek: I see here several problem from my PR:
line 75:
``available.latest()`` has to be replaced with ``available = available.latest()``
line 77:
``available.filter(latest=True)`` has to be replaced with ``available = available.filter(latest=True)``

pghmcfc:
Where is `apply()` documented?

plugins/repoclosure.py
pghmcfc: Where is `apply()` documented?

pghmcfc: Regarding the problems with your PR, this is what I was getting at in the last part of rpm-software-management/dnf-plugins-extras/pull/50#issuecomment-264415495.

j-mracek: Unfortunately query apply() is not documented anywhere. Query.run() is mentioned in hawkey documentation, but hawkey is not alive anymore. Libdnf that replaces hawkey doesn't have documentation yet.

pghmcfc: Using `apply()` does seem to work for me, though I don't have a "failing" condition yet so I can't say for certain.

j-mracek:
I fixed problems mentioned above with previous implementation. Please can you rebase the pull request to upstream and then test it? Thanks a lot

pghmcfc:
I got rid of the `available.run()` as it's no longer necessary, due to your addition of `available.apply()` a little further on. I trust that `apply` operates on the query itself and doesn't return a new one like `filter`?

j-mracek:
Thanks a lot for your contribution

PR #233, opened 2017-09-30
AUTHORS: add leaves author

esmil:
I think you forgot me when copying the leaves plugin from dnf-plugin-extras


ignatenkobrain:
Time to kill AUTHORS file...

PR #237, opened 2017-10-20
Enable JSON output option in repoclosure plugin

kellinm:
- add --json-output option which outputs JSON that is
  readable by those writing scripts against the output
  from the repoclosure plugin
- update the documentation to reflect the new feature

Signed-off-by: Robert Marshall <rmarshall@redhat.com>

### DIFF
--- a/doc/repoclosure.rst
+++ b/doc/repoclosure.rst
@@ -57,6 +57,9 @@ Options
 ``--repo <repoid>``
     Specify repo ids to query, can be specified multiple times (default is all enabled).
 
+``--json-output``
+    Outputs the list of unresolved package dependencies as JSON for use in scripts
+
 
 --------
 Examples
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -24,6 +24,7 @@ from __future__ import unicode_literals
 from dnfpluginscore import _
 
 import dnf.cli
+import json
 
 
 class RepoClosure(dnf.Plugin):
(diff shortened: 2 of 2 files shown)


ignatenkobrain:
I'm kinda reluctant to this. There is already output which is yum compatible and I don't think that json makes it anyhow better..

Also there is concern that you should not use repoclosure at all if you want to parse output. You should just do whatever repoclosure does, but yourself through dnf API.

j-mracek:
The pull request is in conflict with https://github.com/rpm-software-management/dnf-plugins-core/pull/237

j-mracek:
I think that json output is a good idea, but not into output but directly into file. Stdout is often modified by other plugins that could damage an json output. I am just adding support of additional analyses for modular content and structured output is more than welcome. I also have a plan to json output but using a different way and different structure. @kellinm How do you think we should handle this PR?

kellinm:
> I think that json output is a good idea, but not into output but directly into file. Stdout is often modified by other plugins that could damage an json output. I am just adding support of additional analyses for modular content and structured output is more than welcome. I also have a plan to json output but using a different way and different structure. @kellinm How do you think we should handle this PR?

If I recall correctly, the original purpose here was for output that could be piped into `shell` scripts that were making use of `jq` to parse JSON and get fields. I'm not sure that writing output to a file is sensible for that kind of use case; it adds a step and if someone just wanted a file they could use the option and redirect the output.

I am curious though - could you provide an example use case for when another plugin's interaction with this flag would break things?

m-blaha:
We would like to resolve this in dnf5.

PR #238, opened 2017-10-23
[migrate]: use swdb_transformer for database migration

edynox:
Migrate data from yum directly into unified software database using
swdb_transformer from dnf.db

Refs: https://github.com/rpm-software-management/dnf/pull/966


dnf-plugins-core.spec
```diff
@@ -1,4 +1,4 @@
-%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 2.7.0}
+%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 2.8.0}
```
j-mracek: @edynox Please could you change to 2.8.1

edynox: @j-mracek done

j-mracek:
Tested and it solves the issue with migrate plugin.

PR #241, opened 2017-11-10
version lock works on unavailable packages (RhBug:1431491)

m-blaha:
lock/exclude of packages not available in repositories any more was
not applying.

https://bugzilla.redhat.com/show_bug.cgi?id=1431491

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -36,6 +36,9 @@ EXCLUDING_SPEC = _('Adding exclude on:')
 DELETING_SPEC = _('Deleting versionlock for:')
 NOTFOUND_SPEC = _('No package found for:')
 NO_VERSIONLOCK = _('Excludes from versionlock plugin were not applied')
+APPLY_LOCK = _('Versionlock plugin: number of lock rules from file "{}" applied: {}')
+APPLY_EXCLUDE = _('Versionlock plugin: number of exclude rules from file "{}" applied: {}')
+NEVRA_ERROR = _('Versionlock plugin: could not parse pattern:')
 
 locklist_fn = None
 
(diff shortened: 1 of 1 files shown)


m-blaha:
@rh-atomic-bot try

plugins/versionlock.py
```diff
@@ -75,15 +78,21 @@ def sack(self):
...
+            nevra = possible_nevras[0]
+            pat_query = nevra.to_query(self.base.sack)
 
             if excl:
-                excludes_query = excludes_query.union(solution['query'])
-            elif solution['nevra'] is not None and solution['nevra'].name is not None:
-                locked_names.add(solution['nevra'].name)
-                locked_query = locked_query.union(solution['query'])
-        if locked_query:
+                logger.info("%s %s", APPLY_EXCLUDE, pat)
```
j-mracek: It cannot be info, but debug. 
I also would prefer if it will log all excludes or locks at once. See "sudo dnf install -v acpi"

plugins/versionlock.py
```diff
@@ -75,15 +78,21 @@ def sack(self):
                 excl = True
 
             subj = dnf.subject.Subject(pat)
-            solution = subj._get_nevra_solution(self.base.sack, with_nevra=True,
-                                                with_provides=False, with_filenames=False)
+            possible_nevras = list(subj.get_nevra_possibilities(forms=[hawkey.FORM_NEVRA]))
+            if not possible_nevras:
+                logger.info("%s %s", NEVRA_ERROR, pat)
```
j-mracek: This is an error, not an info. It only happens if user was too creative.

m-blaha:
@j-mracek required changes completed

plugins/versionlock.py
```diff
@@ -66,24 +69,38 @@ def sack(self):
...
+                excludes_query = excludes_query.union(pat_query)
+            else:
+                locked_count += 1
+                locked_names.add(nevra.name)
+                locked_query = locked_query.union(pat_query)
+
+        if excluded_count:
+            logger.debug(APPLY_EXCLUDE, locklist_fn, excluded_count)
+        if locked_count:
+            logger.debug(APPLY_LOCK, locklist_fn, locked_count)
```
j-mracek: What about to replace "APPLY_LOCK, locklist_fn, locked_count" by APPLY_LOCK.format(locklist_fn, locked_count) if APPLY_LOCK = _('Versionlock plugin: number of lock rules from file "{}" applied: {}'). It allows for translators to exchange order of arguments.

PR #242, opened 2017-12-05
Add sentence about security updates

dhgutteridge:
Users often need to determine what processes require restarting because of security updates to shared libraries and the like. This really merits mentioning to reinforce this practice.


m-blaha:
@rh-atomic-bot r+

PR #245, opened 2018-01-01
 Fix repoclosure --best on multilib arches (RhBug:1530081)

pghmcfc:
See https://bugzilla.redhat.com/show_bug.cgi?id=1530081

query.filter(latest=True) now only picks the one latest version of a package across all repos and arches (arbitrarily picking one arch when there are multiple with the same NEVR). The previous behavior can be obtained by using query.filter(latest_per_arch=True), which fixes this problem.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,6 +27,7 @@ DNF-PLUGINS-CORE CONTRIBUTORS
     Michal Domonkos <mdomonko@redhat.com>
     Michal Novotny <clime@redhat.com>
     Neal Gompa <ngompa13@gmail.com>
+    Paul Howarth <paul@city-fan.org>
     Rickard Dybeck <r.dybeck@gmail.com>
     Tomas Babej <tomasbabej@gmail.com>
     Vladan Kudlac <vladankudlac@gmail.com>
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -74,7 +74,7 @@ class RepoClosureCommand(dnf.cli.Command):
         if self.base.conf.best and not self.opts.check:
             available = available.latest()
         elif self.opts.newest or self.base.conf.best:
-            available = available.filter(latest=True)
+            available = available.filter(latest_per_arch=True)
         if arch is not None:
             available = available.filter(arch=arch)
         pkgs = set()
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -68,37 +68,55 @@ class RepoClosureCommand(dnf.cli.Command):
 
     def _get_unresolved(self, arch=None):
         unresolved = {}
-
         deps = set()
-        available = self.base.sack.query().available()
-        if self.base.conf.best and not self.opts.check:
-            available = available.latest()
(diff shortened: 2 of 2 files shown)


m-blaha:
@rh-atomic-bot try

m-blaha:
I don't think this is OK. Basically this patch means that --newest and --best will do the same thing - because available.filter(latest_per_arch=True) and available.latest() are equivalent. Don't we need something like this?

``available = self.base.sack.query().available()
        if self.base.conf.best and not self.opts.check:
            available = available.latest()
        elif self.base.conf.newest:
            available = available.filter(latest=1)
``

pghmcfc:
As it stands I don't know how to make --newest work like it used to do in Fedora 25, as I don't know how to do a filter that finds the latest versions of each package in each repo. I agree that the options processing could do with some work and I have a thought on how to do that, but without a latest-per-arch-per-repo filter I don't know how to implement --newest that will be any different from --best.

As it stands though, --best just doesn't work for multilib repos, at least unless you do separate repoclosures for each arch.

j-mracek:
To ensure that latest packages per repo in set I would propose something like:
```
query_latest_per_repo = self.base.sack.query().filter(empty=True)
for repo in self.base.repos.iter_enabled():
    query_latest_per_repo = query_latest_per_repo.union(self.base.sack.query().filter(reponame=repo.id).latest())
```
Additionally what does each latest filter:
The behavior of query().filter(latest=1) was chenge because we discovered that that filter does not sort packages therefore result was random but in 95-99% it returns something like 
latest packages per repo. After commit (https://github.com/rpm-software-management/libdnf/commit/06c5040ab1fd75bbe44cb0cf1f2bda3a6d3535d5 ). The new behavior was not released yet, it is only part of copr dnf-nightly repo (```sudo dnf copr enable rpmsoftwaremanagement/dnf-nightly```).

New behavior:
For query().filter(latest_per_arch=1) and query().latest()
 - returns all latest packages per arch
 - if packageA-1-1.x86_64 and packageA-1-2.i686 it returns both

For query().filter(latest=1)
 - returns all latest packages
 - if packageA-1-1.x86_64 and packageA-1-2.i686 it returns only packageA-1-2.i686

pghmcfc:
@j-mracek, thanks for that suggestion. I've now updated the pull request to make --best and --newest do the different things they were intended to do, and untangled the options processing to make it more clear how it works.

pghmcfc:
@rh-atomic-bot retry

j-mracek:
@rh-atomic-bot try

pghmcfc:
Can't see what failed in that rh-atomic-bot test as the server is unresolvable outside redhat. Hence I can't fix it.

j-mracek:
@rh-atomic-bot retry

j-mracek:
@pghmcfc Some problem on infrastructure

j-mracek:
Thank you very much for contribution

j-mracek:
@rh-atomic-bot r+

j-mracek:
@rh-atomic-bot retry

PR #246, opened 2018-01-11
Note versionlock works on packages absent from repos (RhBug:1431491)

dhgutteridge:
Adding a note about the resolution of https://bugzilla.redhat.com/show_bug.cgi?id=1431491. (I'd tripped over this myself before.)


dhgutteridge:
I realize my proposed addition is perhaps incomplete in its explanation. For instance, a package would have to exist at the time the lock or exclude is added, e.g. to exclude a package presently in Fedora's updates-testing, the user would have to enable that repo in order to supply the package NEVRA.

dhgutteridge:
I've revised the original patch to reflect versionlock's workings more carefully.

PR #247, opened 2018-01-18
Reduce query creation for leave plugin about 18x

j-mracek:


m-blaha:
@rh-atomic-bot try

m-blaha:
@rh-atomic-bot r+

PR #248, opened 2018-03-19
Enhance error msg for download --resolve (RhBug:1550006)

j-mracek:
It didn't provide any information why there is any download.

https://bugzilla.redhat.com/show_bug.cgi?id=1550006


m-blaha:
@rh-atomic-bot r+

PR #249, opened 2018-03-21
Try

PotatoSan:


j-mracek:
I guess that alternative patch was merged that is nearly same like proposal here. Please if any issue in present code will be discovered, don't hesitate to provide additional patch.

PR #250, opened 2018-03-26
Bump dnf max compatible version

edynox:
Required for: https://github.com/rpm-software-management/libdnf/pull/433


j-mracek:
What about to remove l dnf_not_compatible requires and buildrequires completely. I don't see any reason why to keep it.

edynox:
@j-mracek I removed it.

PR #251, opened 2018-03-27
Conditionalize building of python2 subpackages

m-blaha:


torsava:
The conditionalization looks good to me!
One thing though, you're dropping the python2 subpackages for Fedora > 28, but one of the subpackages (python2-dnf-plugins-core) is not a leaf and stuff depend on it. Therefore I don't think it can be dropped from Fedora just yet.

stratakis:
A nice graphical [representation](http://fedora.portingdb.xyz/pkg/dnf-plugins-core/) of which packages are leaves in Fedora.

stratakis:
Is it possible to move this forward?

j-mracek:
@ignatenkobrain How do you think it will be possible to check users of python2 plugins?

j-mracek:
@rh-atomic-bot try

j-mracek:
The code looks fine.

j-mracek:
@rh-atomic-bot r+

PR #252, opened 2018-03-28
Not build local plugin on rhel

j-mracek:
The plugin is not required by rhel user due to availability of all packages.


m-blaha:
This will disable the local plugin also on CentOS. Just making sure (because you mentioned only rhel in description) - is that desired behavior?

j-mracek:
I also would prefer to exclude local plugin for centos

m-blaha:
rpmbuild on CentOS is failing with error:
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/mblaha/rpmbuild/BUILDROOT/dnf-plugins-core-2.1.5-1.el7.centos.x86_64
error: Installed (but unpackaged) file(s) found:
   /etc/dnf/plugins/local.conf
   /usr/lib/python2.7/site-packages/dnf-plugins/local.py
   /usr/lib/python2.7/site-packages/dnf-plugins/local.pyc
   /usr/lib/python2.7/site-packages/dnf-plugins/local.pyo
   /usr/share/man/man8/dnf.plugin.local.8.gz

j-mracek:
@m-blaha Hope that the new version will work

Conan-Kudo:
I find it very strange that `createrepo_c` isn't going to be available in RHEL 8. It seems like it'd be a good idea for such a basic tool to be part of RHEL, as it has in the past (i.e. `createrepo` in RHEL <= 7)...

PR #253, opened 2018-04-06
copr: add support for multiple copr instances (RhBug:1478208)

dturecek:
This addresses https://bugzilla.redhat.com/show_bug.cgi?id=1478208.


praiskup:
Could the name of instance (the id probably) be part of the filename which is created when the repository is added?  Consider that the same project, say `jdoe/foo` exists in multiple copr instances.

etc/dnf/plugins/copr.conf
```diff
@@ -0,0 +1,6 @@
+[fedora-copr]
+url = https://copr.fedorainfracloud.org
+
+# To add another Copr instance, make a new section with a URL to it.
```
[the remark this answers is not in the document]
xsuchy: No, we should not post hostnames of our internal hosts. You should propably even delete this comment.

clime:
Note how git solve the problem: of having multiple related config sections. E.g.:

    [branch "master"]
        remote = origin
        merge = refs/heads/master
    [branch "detached_builds"]
        remote = origin
        merge = refs/heads/detached_builds

So maybe we should follow that pattern and have:

    [instance "fedora"]
    ...
    [instance ...]

the problem with the current solution is what happens when you will want to add a generic plugin setting applicable across all instances. You will need to add section, let's say `[plugin]` that shouldn't be selectable with ``--instance``.

xsuchy:
+1 to s/--instance/--copr/

plugins/copr.py
```diff
@@ -110,6 +102,12 @@ def configure(self):
 
         if config_file is not None:
             copr_plugin_config.read(config_file)
+            if copr_plugin_config.has_option(copr_instance, 'url'):
+                self.copr_url = copr_plugin_config.get(copr_instance, 'url')
+            else:
+                print(_("Warning: No such instance '{}' in configuration file. "
+                        "Using the default one instead '{}'.").format(copr_instance, self.copr_url))
```
xsuchy: Use `{0}` and `{1}`. The order of the placeholders does not need to be the same in all translations.

etc/dnf/plugins/copr.conf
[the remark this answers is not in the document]
praiskup: I'm curious where's such rule, and who we help with this rule.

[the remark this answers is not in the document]
praiskup: ... and if we really have to have such artificial obstacles, I'd rather propose to have layout like:
```
/etc/dnf/plugins/copr/
/etc/dnf/plugins/copr/fedora.conf
```
And I could provide internally something like:
```
/etc/dnf/plugins/copr/redhat-internal.conf
```

praiskup:
+1 for the `--copr` option; -1 for the `[instance "sth"]` format.  `[sth]` seems to be enough.

clime:
> +1 for the --copr option; -1 for the [instance "sth"] format. [sth] seems to be enough.

[sth] is not enough. Please, read my previous comment.

praiskup:
You can use `[DEFAULT]` section, IMO, if you really need that.  But do we need that actually?

clime:
I see this as an alternative to git remote:

    [remote "origin"]
        url = ssh://git@pagure.io/copr/copr.git
        fetch = +refs/heads/*:refs/remotes/origin/*
    [remote "pr194"]
        url = https://pagure.io/forks/praiskup/copr/copr.git
        fetch = +refs/heads/*:refs/remotes/origin/*

So maybe we could even use "remote" as the keyword. I leave the final decision to dturecek here.

My main points here are:
   - explicit is better than implicit (see `python -m this`)
   - given that COPR will become a system closely integrated with git, it would be good to follow patterns that git already uses

Btw. copr might start meaning "module" in some time so maybe that's not the best keyword in config there. I vote for `--instance` or `--remote` there. I am scratching my previous suggestion.

praiskup:
> Btw. copr might start meaning "module" in some time ...

I don't think so, but yeah -- the "copr" term is ambiguous even now (is that project in "copr" or "copr" as a service?).

> I see this as an alternative to git remote:

No big problem with me, I'm just giving it -1 ;) since I don't believe we really need that (cons: it is syntax which configparser doesn't know, so it means some additional custom handling of string).

xsuchy:
> the "copr" term is ambiguous

For some time I  am pushing for term "project" for individual project and term "copr" for the server itself.

clime:
In the end, I am suggesting to use word "hub" and `--hub, -h` switch. I think that's the most precise term because a COPR server provides the same kind of service as Github. In our case we are providing yum/dnf repos instead of Git repos but otherwise it's the same thing. I am also suggesting to name config sections explicitly with e.g. `[hub "fedora"]` because it should be clear what config sections are being indexed with `--hub` switch.

For easy-adding hubs we can implement `dnf copr add-hub` command but that would be in some later PR.

clime:
Sorry for all the previous suggestions. Sometimes, it takes time to arrive to a correct solution.

dturecek:
Thanks for all the suggestions. In the end, I went with the `--hub` switch and `[hub "fedora"]` config section format.

praiskup:
We really need config directory, not just one config file (as long as we refuse to add other copr instances).

clime:
> We really need config directory, not just one config file (as long as we refuse to add other copr instances).

Can we do it in a next PR? Also it would deserve some explanation why it is needed.

praiskup:
> Can we do it in a next PR?

I would prefer not to, to not change the config API in the future PR.

> Also it would deserve some explanation why it is needed.

It's above ... I'd like to wrap internal package (say called `rhpkg-cli`) which will provide
the configuration for internal copr users.  It would be clumsy to require everybody to
configure this manually.  As I mentioned before, I would really prefer to have the internal
copr configured by default so nobody has to do anything ... ;-) but if that's not kosher..

clime:
> It's above ... I'd like to wrap internal package (say called rhpkg-cli) which will provide
> the configuration for internal copr users.

Well, then from the spec file you can do:

    cat >> /etc/dnf/plugins/copr.conf <<EOF                                                                                                            
    [hub "internal"]
    url=...
    EOF

Is this a sufficient solution for you? I don't think we need to change the current solution when this is available.

praiskup:
> Is this a sufficient solution for you? 

This is just hack... (I can not touch others' packages configs from my package)  :/ What's the problem with the proposed config-directory approach?

clime:
> Is this a sufficient solution for you? I don't think we need to change the current solution when this is available.

Do you agree?

clime:
> This is just hack... (I can not touch others' packages configs from my package) :/ 

Well, we can add `dnf copr enable add-hub <>`? I assume your package will install `dnf-plugins-core` as a requirement? You can then put the command into a `%post` scriplet.

> What's the problem with the proposed config-directory approach?

We already have a good short implementation that works with just with one file that dnf provides to us. What you suggest is a custom copr plugin behaviour probably completely different from what other plugins do.

praiskup:
On Tuesday, April 17, 2018 11:13:44 AM CEST Michal Novotný wrote:
> > This is just hack... (I can not touch others' packages configs from my package) :/ 
> 
> Well, we can add `dnf copr enable add-hub <>`? I assume your package
> will install `dnf-plugins-core` as a requirement? You can then put the
> command into a `%post` scriplet.

Touching files which I do not own is hack.  That's not correct packager's
thinking...

> > What's the problem with the proposed config-directory approach?
> 
> We already have a good 

not good enough

> short implementation that works with just with one file that dnf
> provides to us.

Alright, I checked the code now and I see what you mean here
(`self.base.conf.pluginconfpath`).  Everything else is provided by
dnf-plugins-core... so it is trivial to add support for e.g.
`"{}/{}.d".format(path, PLUGIN_CONF)` same as we do with
`"{}/{}.conf".format(path, PLUGIN_CONF)` files.

> What you suggest is a custom copr plugin behaviour probably completely
> different from what other plugins do.

Completely different is the ini format change you proposed :-( What I
propose is consistent with the `/etc/yum.repos.d` semantics.

praiskup:
On Tuesday, April 17, 2018 3:12:09 PM CEST Michal Novotný wrote:
> ... then later extend it when needed.

It _is_ needed (as per f2f mtg).  Your suggestion feels like attempt to
delay something for no reason.

etc/dnf/plugins/copr.conf
```diff
@@ -0,0 +1,6 @@
+[hub "fedora"]
```
praiskup: Let's dump it here in the diff .. I would simply prefer `[fedora]` ini header here:
- it is standard and native for configparser
- every other configuration in dnf/yum was fine with that so far, so I doubt we need to invent something new here (what `hub "fed" "ora"` means?, etc.)

etc/dnf/plugins/copr.conf
```diff
@@ -0,0 +1,6 @@
...
+url = https://copr.fedorainfracloud.org
+
+# To add another Copr hub, make a new section with a URL to it.
+# [hub "name-of-hub"]
```
j-mracek: What do you think to use directly hub name as section name like [fedora]? It will make code much easier. Additional what about to add section [main], where you can define default_hub=fedora? Or something like that.

clime:
Well, let's really take an inspiration from Git, which is fully distributed and I would really just follow it because Copr does the same (similar) thing except sources are already built and ready to be installed. 

We would be writing something like:

    dnf copr enable copr.fedorainfracloud.org/owner/project
    dnf copr list copr.fedorainfracloud.org/owner

protocol could be assumed to be https:// or we could make it explicit as well.

If I e.g. look at:

    $ dnf copr list
    clime/rpkg-client 
    pjp/init-vars
    group_copr/copr-dev

which currently lists enabled coprs, the problem is that you could also have e.g ` pjp/init-vars` in internal COPR and then if you are working with both COPR instances in one system, you don't know whether the `pjp/init-vars` copr comes from the internal instance or Fedora instance and that might be a potentially important information.

We should really be adding new functionality after we are all aware of what we are doing. That will require some communication and planning though.

So let's really discuss this one and then get back to implementation.

xsuchy:
> dnf copr enable copr.fedorainfracloud.org/owner/project
> dnf copr list copr.fedorainfracloud.org/owner

This is IMHO overengineering. While there are a lot of different gits, most people will use only one copr. Some of them two. We (developers) maybe 3 (staging). And that's all. Writing name of the instance to every enable/list is overkill.

clime:
> This is IMHO overengineering. While there are a lot of different gits, most people will use only one copr. Some of them two. We (developers) maybe 3 (staging). And that's all. Writing name of the instance to every enable/list is overkill.

I don't think it's overengineering. The `dnf copr list` problem that I mentioned is a real problem. There obviously are people who work with multiple Copr instances and they should now from which given instance the copr listed with `dnf copr list` comes.

> Writing name of the instance to every enable/list is overkill.

Not really if one can just copy-paste it from a text-input field the same way we are doing for every Git repo on each clone. Btw. we can have a shortcuts like `--hub` or default URL specification in case `<owner>/<project>` is used and not `<baseurl>/<owner>/<project>`.

I would like to talk first about how implement the multi-site support so that it really work for each `dnf copr` command and for any number of possible Copr instances. Implementing the multi-site support with assumption that there are only two or three COPR instances (which is true currently) will simply just lead to an implementation that doesn't scale.

praiskup:
> Alternatively we could put the hubname to the repo filename, e.g. _copr_internal_phracek-PyCharm.repo

Ah, I forgot about this, yes -- this needs to be done anyways (requested above already).

praiskup:
On Wednesday, April 18, 2018 9:18:50 PM CEST Miroslav Suchý wrote:
> > dnf copr enable copr.fedorainfracloud.org/owner/project
> > dnf copr list copr.fedorainfracloud.org/owner
> 
> This is IMHO overengineering. While there are a lot of different gits, most
> people will use only one copr. Some of them two. We (developers) maybe 3
> (staging). And that's all. Writing name of the instance to every enable/list
> is overkill.

I kind of like the idea to make those two equivalent

    $ dnf copr enable jdoe/foo # defaults to '--hub fedora'
    $ dnf copr enable fedora/jdoe/foo # the fedora/ prefix is allowed

And those too:

    $ dnf copr enable --hub redhat jdoe/foo
    $ dnf copr enable redhat/jdoe/foo

I would be against using the "full.url.example.com/jdoe/foo" variant.  But
otherwise, really nice idea (which doesn't have to be implemented right now).

xsuchy:
> [hubname-ownername-projectname]
I *think* there is some limit on the length of the repoid. @j-mracek  is it? or not?

praiskup:
On Wednesday, April 18, 2018 11:24:12 PM CEST Jakub Kadlčík wrote:
> > [hubname-ownername-projectname]
> > I think there is some limit on the length of the repoid. @j-mracek is it? or not?
> 
> If yes, we can use the filename indicator as also @praiskup suggested in
> [253#issuecomment-380072552](https://github.com/rpm-software-management/dnf-plugins-core/pull/253#issuecomment-380072552)

The 'hubname' should be the string from the config ini file.. but as I
understand it, the dnf plugin only downloads the repofile from copr
frontpage ATM;  and copr frontend has no idea how particular user
configured the 'hubname' in his /etc/dnf/plugins.d/copr.d/xxx.conf.

So basically, instead of just 'wget
hostname/coprs/jdoe/foo/repo/fedora-rawhide/jdoe-foo-fedora-rawhide.repo'
we could tweak the repo id and the repo file name after downloading.
Cut&pasted repos don't have to have the same repo ID.

clime:
Personally, I find the idea with `/etc/dnf/plugins.d/copr.d/xxx.conf` problematic because it is not clear in which order the config files should be read. In the end, you just need only a single (merged) configuration inside the plugin and if you have multiple config files, you need to determine order in which they are being read. What would happen if you have two separate config files in `/etc/dnf/plugins.d/copr.d/` each one containing a section with the same hub name but a different url? Which URL will win? We should prefix the filenames there with numbers so that the behavior is deterministic. But then again, that doesn't really make sense because any potential Copr sites that would be installing the configs are completely independent of each other and they cannot agree on any numbering (in other words, we are not making httpd plugins here). Can this be accepted as an why-not argument? If not, we can investigate it further but I don't really think the directory solution is the right one.

I think we should go for full URLs everywhere if technically possible because they are pretty much guaranteed to be unique and they are independent of a user configuration. If you then put `dnf copr enable some.copr.org/owner/project` into the Quick Enable window in your particular COPR instance, you don't need to do anything else.

I think any configuration we are suggesting in this PR should be only to create some shortcuts to save typing. It should not be an actual multi-site support implementation.

praiskup:
On Thursday, April 19, 2018 12:55:06 AM CEST Michal Novotný wrote:
> Personally, I find the idea with `/etc/dnf/plugins.d/copr.d/xxx.conf`
> problematic because it is not clear in which order the config files should be
> read.

Tooling out there usually goes with the file-alphabetical order;  and we
don't really have to define the order unless it is really needed.  We
should respect the contributor's opinion here.

> In the end, you just need only a single (merged) configuration inside the
> plugin and if you have multiple config files, you need to determine order
> in which they are being read.

No, you don't have to.  Those files should be create so they don't collide, and
the order shouldn't matter (same as *.repo files ordering doesn't matter).

> What would happen if you have two separate config files in
> `/etc/dnf/plugins.d/copr.d/` each one containing a section with the same
> hub name but a different url?  Which URL will win?

What happens if you specify two [blah] repos in /etc/yum.repos.d, or if
you specify [foo] and [bar] with the same baseurl=''?  Take an inspiration
from that.

In the end, this situation shouldn't happen at all.  We are talking about
hypothetical corner case.

Really, have a look at ... pastebinit, pybugz, yum, dnf, mock, apache,
systemd uses drop-in directories a lot! ... how they are handling config
directories, blind list for inspiration (only on my _desktop_ box):
    $ ls /etc/ -lah | grep \\.d | wc -l
    60

> We should prefix the filenames there with numbers so that the behavior
> is deterministic.

Feel free to complicate things; I don't mind as long as Fedora Copr is the
default as @dturecek did it.

> But then again, that doesn't really make sense because any potential
> Copr sites ...

Potential copr sites should identify themselves to not collide with the
two configs called (e.g.) 'redhat-copr.conf' and 'fedora-copr.conf'.  If
there will be any collisions, people will notice very soon..

> that would be installing the configs are completely independent of each
> other and they cannot agree on any numbering (in other words, we are not
> making httpd plugins here).

The config files are not automatically generated.  It is either packaged
(by Red Hat internal copr maintainers, dnf-plugins-core, or other copr
party), or it is created by the end user manually.  Those should be blamed
if there's some collision and discuss with each other.

> Can this be accepted as an why-not argument?

I don't accept it, I see no reason to not proceed.

Blocking the issue with "hey, one config is enough" is not good and I
already explained.  I don't see any other than your personal reasons to
stick with half-way solution.

> I think we should go for full URLs everywhere if technically possible
> because they are pretty much guaranteed to be unique and they are
> independent of a user configuration.
> If you then put `dnf copr enable some.copr.org/owner/project`.

Note that even fedora copr frontend moved the hostname from A to B several
times.  It would just unnecessarily lead to staled plugin configuration.

> I think any configuration we are suggesting in this PR should be only to
> create some shortcuts to save typing. It should not be an actual
> multi-site support implementation.

I don't think I understood this comment.

clime:
> Tooling out there usually goes with the file-alphabetical order; and we don't really have to define the order unless it is really needed. We should respect the contributor's opinion here.

Ye, well, we should have the order clearly defined. I don't think tools with (seemingly) random behavior are particularly user-friendly.

> What happens if you specify two [blah] repos in /etc/yum.repos.d, or if you specify [foo] and [bar] with the same baseurl=''? Take an inspiration from that.

``[foo]`` and ``[bar]`` with the same base URL is really not a problem. Two ``[blah]`` repos with different URLs is the problem. I don't know what dnf will do there and I want to avoid that situation. DNF should probably throw an error about non-unique repo ID and we should probably do the same thing in the `dnf copr` plugin about non-unique hub name. The thing is it isn't what python `ConfigParser` does by default so we would need to implement some custom logic to parse the configuration correctly.

> Potential copr sites should identify themselves to not collide with the two configs called (e.g.) 'redhat-copr.conf' and 'fedora-copr.conf'. If there will be any collisions, people will notice very soon..

Well, if you use domain names in the full copr path (`<domain>/<owner>/<project>`), then the problem is not there at all. They don't need to know about each other and it will work.

> Blocking the issue with "hey, one config is enough" is not good and I already explained. I don't see any other than your personal reasons to stick with half-way solution.

I am not really blocking the issue. You are blocking it. We are trying our best to implement what you have suggested in https://bugzilla.redhat.com/show_bug.cgi?id=1478208 with very slight improvement of using shorter param name. You then started pushing some extension to the solution without having an actual tool in hand that would make a use of it.

Anway, I just wanted to note that those should be equivalent:

    dnf copr enable --hub fedora owner/project
    dnf copr enable copr.fedorainfracloud.org/owner/project
    dnf copr enable owner/project (thanks to by default provided configuration in dnf-plugins-core package)

Hopefully, we can get domain name back to` copr.fedoraproject.org` but that's a different problem.

> Note that even fedora copr frontend moved the hostname from A to B several times. It would just unnecessarily lead to staled plugin configuration.

Well, there would be no stalled plugin configuration except what user puts in. There would be some yum.repos.d confs with IDs for the old domain. I don't think that should represent a problem. Hostname changes sometimes. The same way user configuration might change.

>  I think any configuration we are suggesting in this PR should be only to
 create some shortcuts to save typing. It should not be an actual
 multi-site support implementation.

> I don't think I understood this comment.

It means that you will be able to put `dnf copr enable <internal-copr-url>/owner/project` into the 'Quick Enable' window for your COPR instance. And `dnf copr enable` will, therefore, work for users of your instance without any additional configuration needed. That's a good thing.

> Feel free to complicate things; I don't mind as long as Fedora Copr is the default as @dturecek did it.

Well, ok, agreed. That means we can probably close the argument. The thing is that we should really first implement the solution without any "shortcuts" and then having them introduced but it's not that big deal.

praiskup:
On Thursday, April 19, 2018 9:54:58 AM CEST Michal Novotný wrote:
> We are trying our best to implement what you have suggested in
> https://bugzilla.redhat.com/show_bug.cgi?id=1478208 with very slight
> improvement of using shorter param name. You then started pushing some
> extension to the solution without having an actual tool in hand that
> would make a use of it.

Good point, and since the "initial" thought wouldn't be packageable (as
mentioned before), I already said why the extension is important.  You
just insist on hackish solution without real benefits.

I'm tired to talk to @clime here more, @dturecek, feel free to ping me on
IRC if you are are in doubts (or @AloisMahdal potentially, as OP) since
I'm representing the only users who will actually use the code frmo this
PR, as that's the only other copr instance.  I'm OK to take over the PR
work if that's needed, I'm used to fight with @clime's blocking attempts
regarding internal copr instance.

> Anway, I just wanted to note that those should be equivalent:
> 
>     dnf copr enable --hub fedora owner/project
>     dnf copr enable copr.fedorainfracloud.org/owner/project

Those can not be equivalent, since this doesn't have the "hub ID" in
config file.  So the repofile created will be "generated-but-orphaned",
and not paired with any plugin cconfiguration;  so feel free to implement
it -- but this is not what is needed actually.

>     dnf copr enable owner/project (thanks to by default provided
>                   configuration in dnf-plugins-core package)

Yes.

> It means that you will be able to put `dnf copr enable
> <internal-copr-url>/owner/project` into the 'Quick Enable' window for
> your COPR instance.

Because of the "orphaned" repo files problem, I'm -1 here.  I don't use
that "Quick Enable", and I'm OK to provide there equivalent:

  `dnf copr enable hub-id/owner/project`

or `dnf copr enable --hub hub-id owner/project`.  Anyway, this is
orthogonal thing.

xsuchy:
I just tested the length of repoid and dnf works just fine with any length (well I tried the length of 219 character)

AloisMahdal:
What about

    dnf copr add foo https://foo.my.intranet/   # a very one-time thing

which is when you bind the copr id with base URL, and drop `foo.conf` somewhere in /etc.  Avoiding conflicts is easy: plugin merely fails if there is already `foo`. Also since the above just adds `foo.conf` somewhere, you can do it by installing rpm (again, conflict solved by rpm), ie. having the `foo.conf` in some team/dept rpm.

Then on CLI, copr projects would have 2 kinds of syntax: fully qualified `<copr>/<user>/<project>` -- that would require `foo.conf` already in place, and `<user>/<project>` --- that would merely talk to whateve is the top COPR (ie. `fedora`).

    dnf copr enable amahdal/shellfu             # same as always
    dnf copr ebable fedora/amahdal/shellfu      # as above, but if you want to be darn sure
    dnf copr enable foo/bar/baz                 # enabling bar's baz on foo project
    dnf copr list foo/quux                      # what does quux offer?

This would solve the namespace problem when you have `bar/baz` at fedora copr *and* `bar/baz` at internal copr: people using more coprs would simply get used to using the more qualified id for more specific purpose.

---

@clime, the comparison with git makes sense on level of semantics of remote management commands.  Above suggestion takes inspiration from that.  On the other hand, the comparison is limited: COPR is *not* distributed peer-to-peer like git: AFAIK it's still client-server architecture where server provides builds to clients: you don't get to spawn coprs like you do git repos, and you don't get to connect copr to other copr or client system to another one.

About the `copr.d` -- IMO it's a no-brainer. Touching others' config from .spec file is a no-go and distribing own copr should be easy (after all, that's why I posted the rfe to Bugzilla in the first place.)

About the config format: AFAIK dnf/yum use plain `[foo]`.  While `[copr "foo"]` seems nice and correct, I don't t think it's a good idea to do it *in one particular plugin*.  Do it globally in DNF or don't do it.

clime:
@AloisMahdal: Thank for your comment. We really need another person to drop in sometimes (often) :grin: 

I agree very much with you and your suggestion makes lots of sense. Especially:

    dnf copr ebable fedora/amahdal/shellfu      # as above, but if you want to be darn sure

is really nice.

The thing is that I want one thing:

When somebody sets up a new COPR instance for his own purpose and installs `dnf-plugins-core`, then immediatelly what the user sees in the 'Quick Enable' window for his COPR instance should work. 

I would like:

     dnf copr add foo https://foo.my.intranet/

not needed to be run. 

That's because every new person that would want to install something from his/her COPR would need to run that command first before being able to do it.

That's the reason why I think we should be able to write and be able to work with full copr paths like:

    dnf copr enable copr.fedoraproject.org/foo/bar

In a Dockerfile I will typically want to use the full-path specification that doesn't require any setup which also makes things more readable because the person will not encounter stuff like:

    dnf copr enable mycopr/owner/project 

where `mycopr` is defined somewhere else.

In other words. I don't want to be required to set up some "context" first to be able to do something else afterwards.

At the same time, setting up that context can be made an optional feature.

I.e. we can then also have the "Alias" feature (currently implement by `hub`) 
so that you can alias "fedora" with "copr.fedoraproject.org" by putting:

    [hub "fedora"]
    domain_name = copr.fedoraproject.org

record into `/etc/dnf/plugins/copr.conf`. 

Then user would be able to do:

    dnf copr enable fedora/foo/bar

and also `dnf copr list` could output the aliased name:

    fedora/@copr/copr
    fedora/foo/bar
    internal/foo/bar

I don't mind that. I think it is quite user-friendly to allow (but not require) it.

When somebody tries to setup his/her own COPR instance, the required number of steps to make it run should be as thin as possible. So when we can take one step out (or rather not introduce it there)
of that procedure, we should do it really without any thinking. 

-----

As for the `[foo]` vs. `[copr "foo"]`. 

It would really be better to use `hub` to avoid ambiguity. copr all in lower-case means "community project".

Next, parsing the config file is in realm of a particular plugin and it should not be a problem to use even a different config file format then the ini files.

Here we are just clearly labeling things and we are overcoming the limitation of the ini files to have nested config structure. Simulating nested structures with flat structure by using a separator between keys is not uncommon. We could e.g. use `configyaml` python module so that we don't need to do that but why?

And mainly, @j-mracek already mentioned possibility to add `[main]` section and the `default_hub` configuration variable. That section should not be selectable as a hub, which means we really _should_ use the labels when `[main]` section is introduced.

----

Finally,

    dnf copr enable --hub fedora foo/bar

vs.

    dnf copr enable fedora/foo/bar


I agree that what you proposed is nicer. We could probably move to that syntax at some point.

Anyway, again, thank you a lot for your input. It is very valuable for us.

clime:
> Yep, treating the step of adding the hub as a separate explicit step is the right thing. They want to establish trust with the host. That's a very important step. So important that they either want to do it explicitly, or they want to delegate it to already-trusted admins of their corporate repo---that would be installing rpm which drops the copr.d/foo.copr.

> Enabling project, on the other hand, is important but on a different level: First, you're enabling project that is relevant to you, and second, you're establishing trust with user/team based on their coding and/or packaging competence. That's entirely different decision

I think you are just trying to introduce totally redundant step by saying that it means "establishing trust". I can just say you "establish trust" at the moment you decide to install something from some copr. That's actually the important part. Not some redundant `copr add something`.

There are other factors too. The currently proposed "dnf copr add" will require root privileges. I might want to be able run certain commands like  as a non-privileged user. We might e.g. want add (e.g. debugging) functionality that just downloads (not installs) certain copr locally, and user can then install the result into virtual machine or - all that without being a privileged user. E.g `dnf download nano` does a similar thing - it downloads something without needed root privileges and I can use that downloaded result (rpm here) for further operations that again do not require root privileges.

> Well it "should" be problem for project maintainers in a way that it's a bad, bad practice.

How it is bad practice? If anyone needs a config file format with richer expressive power, you will forbid it because other plugins don't do that? No, probably not.

clime:
Per my review, this PR is OK to be merged as implementation of RhBug:1478208.

In case, you would like a new feature to be added, please, open a new BZ at https://bugzilla.redhat.com/enter_bug.cgi?product=Copr or even code your own solution and open a PR here.

praiskup:
The current code struggles with those issues
- [ ] non-standard `[hub "???"]` syntax, please ignore @clime here
- [ ] the new configuration is not packageable, please provide config directory
- [ ] the repo file has always the same filename, regardless of the --hub specified; I would encourage you to create `_copr_<hub_id>_<user>_<project>.repo` files, instead of just `_copr_<user>_<project>.repo`
- [ ] the repo id is always `[USER-PROJECT]` regardless of the --hub specified; I'd suggest to edit the downloaded repofile, and adjust the repo_id to be named like [<HUB_ID>-<USER>-<PROJECT>] - that equals to just adding the "HUB_ID" prefix.  Other solution is to fix the repo file in Copr frontend code which is also fine (and could be done after this PR)
- [ ] dnf copr disable doesn't work reliably (because of the ^^ ambiguity)
- [ ] dnf copr list has a problem with the ambiguity ^^

This won't ever be perfect though - deal with that :-(, we won't ever have 100% backward compatibility (rename of repo id's and repo files is needed) and because of this, it would be awesome to have some `dnf copr update-config` command that would iterate through repo and adjust them according to the new layout.

dturecek:
First, thank you for all your comments. 

Here is what I would like to do:
* change the config syntax back to `[fedora]` as suggested by @j-mracek and @AloisMahdal. When using the `--hub` option we can check whether the value passed to it is not `main` so that will solve the problem of the `[main]` section being unselectable.
* add the `copr.d` config directory. As for the problem with one hub specified over multiple config files, we would just check if its there only once.
* add the hub name to the repo filename and to the repo id.
* add `add-hub` command.

Then, these two would be equivalent:
```
dnf copr enable foo/owner/project
dnf copr enable --hub foo owner/project
```
and when no hub would be specified, it would default to `https://copr.fedorainfracloud.org/`.

However, what I am unsure about is what should happen with the repo file when using this:
```
dnf copr enable copr.fedoraproject.org/foo/bar
```
Should we then create a repo file with name `_copr_copr.fedoraproject.org-foo-bar.repo` that wouldn't match any hub name added later?

I hope that this would be an okay way to do this.
Anyway, thanks again for your suggestions.

praiskup:
@dturecek, agreed with everything.  One more note:

On Tuesday, April 24, 2018 1:13:07 PM CEST dturecek wrote:
> However, what I am unsure about is what should happen with the repo file when
> using this:
> ```
> dnf copr enable copr.fedoraproject.org/foo/bar
> ```
>
> Should we then create a repo file with name
> `_copr_copr.fedoraproject.org-foo-bar.repo` that wouldn't match any hub name
> added later?

That would be unfortunate, -1 for the "orphaned" repo file.  If we implemented
this syntax in the end, it should be two-step action equivalent to:

    $ GENERATED_ID=`generate_id_based_on_url`
    $ dnf copr add-hub copr.fedoraproject.org $GENRATED_ID
    $ dnf copr enable $GENERATED_ID/foo/bar

. and I really don't think it needs to be implemented just now, same as the
'copr add-hub' command.  It's nice plus, but I would really prioritize
the problem with the same "repo ID" for different 'dnf copr enable --hub'
options.  If you have multiple repofiles with the same repoid, I doubt dnf can
handle that correctly... so it is sort of blocker.

AloisMahdal:
About the orphaned (or "anonymous") coprs, I agree with @praiskup.  Let's concentrate on making $Subj work with as little mess as possible.

(Maybe, in future, reasonable discussion could exist whether current dnf makes it possible to support such repos *without* making too much of a mess, but so far I don't see any convincing reason to even have them.)

---

Few non-blocking notes:

>     dnf copr enable foo/owner/project
>     dnf copr enable --hub foo owner/project

I'm not strongly against it, but having 2 ways it does smell a bit like YAGNI.  So unless there's immediate real user advantage in having both, I'd keep just one of them; feel free to pick whichever you like.

---

> add the hub name to the repo filename and to the repo id

Just make sure the way the data is encoded in the file is unambiguous: `_copr_<hub_id>_<user>_<project>.repo` seems to use underscore as delimiter, which can become ambiguous if underscore is in any of the three parts.  So either use different delimiter (`[.+@%]` seem filename-friendly), ban underscore from all but one elements, or check for ambiguity in active manner (fatal or not).

Similar problem could exist with slash and hyphen: eg. I have *amahdal/bisq-exchange* project, which generates *_copr_amahdal-bisq-exchange.repo* with `[amahdal-bisq-exchange]`.  Strange things could happen if hub id or username contained hyphen.  (And it might not be just matter of transition but I guess ambiguous configs could be created even on

---

By the way, the issue of transition from old to new layout safely could be attacked by including plugin-specific format version key and setting it to eg. `1`.  This way you could later have at least clean way to detect "old" format--where the key is not defined---and possibly address later transitions if problems arise.

The advantage is that don't have to go very far: merely ensuring that the key is included in all future version would let you detect and ignore old configs and instruct user to re-add them manually.

etc/dnf/plugins/copr.conf
m-blaha: Please, add "Configuration" section to doc/copr.rst with description of config options and example of config file.
Btw, I'm also voting for simpler sections naming like [fedora]. It will be also consistent with configuration of repositories.

plugins/copr.py
```diff
@@ -86,21 +86,13 @@ def set_argparser(parser):
         list_option.add_argument('--available-by-user', metavar='NAME',
                                  help=_('List available Copr repositories by user NAME'))
 
+        parser.add_argument('--hub', default='fedora',
+                            help=_('Specify a Copr hub to work with'))
+
```
m-blaha: Again, this has to be covered in documentation.

dturecek:
Changes I've made:
-  I added documentation entry for both configuration files and the `--hub` option. 
- `copr.d` directory for config files in addition to the `copr.conf` config -- the plugin ends with an error in case of multiple URLs to the same hub
- the `--hub` option specifies the hub, whose URL has to be in one of the configs
- a hub can be specified like this as well: `dnf copr enable hub/user/project`, where `hub` can be either a URL or ID that has its URL specified in one of the configs (the same way `--hub` works)
- repo file name is now `copr:<hub>:<user>:<project>.repo`, which solves the problem with `-`/`_` in usernames/projectnames

The repo ID format generated by Copr frontend will be changed in [PR 352](https://pagure.io/copr/copr/pull-request/352) to match the repo file names.

dnf-plugins-core.spec
```diff
@@ -412,6 +412,8 @@ PYTHONPATH=./plugins nosetests-%{python3_version} -s tests/
 %license COPYING
 %doc AUTHORS README.rst
 %ghost %{_var}/cache/dnf/packages.db
+%config(noreplace) %{_sysconfdir}/dnf/plugins/copr.conf
+%config(noreplace) %{_sysconfdir}/dnf/plugins/copr.d/*.conf
```
praiskup: Are there some `%{_sysconfdir}/dnf/plugins/copr.d/*.conf` files?  IMO you only want
`%dir %{_sysconfdir}/dnf/plugins/copr.d`

doc/copr.rst
```diff
@@ -60,6 +60,26 @@ Arguments (copr)
 ``search project``
     Search for a given ``project``.
 
+``enable hub/name/project``
```
praiskup: awesome!

plugins/copr.py
```diff
@@ -103,36 +104,88 @@ def set_argparser(parser):
...
-        config_file = None
+        config_files = []
         for path in self.base.conf.pluginconfpath:
-            test_config_file = '{}/{}.conf'.format(path, PLUGIN_CONF)
+            for filename in os.listdir('{0}/{1}.d'.format(path, PLUGIN_CONF)):
+                if filename.endswith('.conf'):
+                    config_file = os.path.join(path, PLUGIN_CONF + ".d", filename)
+                    config_files.append(config_file)
+
+            test_config_file = os.path.join(path, PLUGIN_CONF + ".conf")
```
praiskup: The naming `test_config_file` is a bit misleading, maybe `default_config_file`?

plugins/copr.py
```diff
@@ -103,36 +104,88 @@ def set_argparser(parser):
...
-                self.chroot_config = [distribution, releasever]
-            else:
-                self.chroot_config = [False, False]
+                config_files.append(test_config_file)
+
+        project = []
+        try:
+            project = self.opts.arg[0]
+            project = project.split("/")
+        except (ValueError, IndexError):
```
praiskup: Is the `ValueError` doesn't seem to be needed.  What about this?
```python
if len(self.opts.arg):
    project = project.split(self.opts.arg[0])
```

plugins/copr.py
```diff
@@ -103,36 +104,88 @@ def set_argparser(parser):
...
+        elif self.opts.hub:
+            copr_hub = self.opts.hub
+
+        # Copr hub specified with hub/user/project format
+        elif len(project) == 3:
+            copr_hub = project[0]  # try to find hub in config files
+            self.copr_url = "https://" + project[0]  # otherwise use it directly as URL
+
+        if config_files:
+            hub_found = False
```
praiskup: The `config_files` is always array, so the `if config_files` is not needed;  you can iterate directly.

At the same time, I'd simply break the loop with as soon as possible (related config file found).

plugins/copr.py
```diff
@@ -103,36 +104,88 @@ def set_argparser(parser):
...
+                    self.chroot_config = [distribution, releasever]
+                else:
+                    self.chroot_config = [False, False]
+
+            if self.opts.hub and not hub_found:
+                self.copr_url = self.default_url
+                print(_("Warning: No such instance '{}' in configuration file. "
+                        "Using the default one instead '{}'.").format(copr_hub, self.copr_url))
+
+        self.copr_short_url = self.copr_url
```
praiskup: It's misleading, I'd use `copr_hostname`.

praiskup:
I really like where the PR goes, thanks a lot!  Don't you have testing packages built somewhere?

clime:
As praiskup suggested in a code comment, instead of config param `domain_name`, `hostname` could be used. E.g. `localhost` is not a domain name, which I didn't realize when writing the above :(.

dturecek:
I have changed the code based on your comments, namely:

- `hub/user/project` now only works with hostname as `hub` and doesn't look through configs 
- I separated `url` in configs to `hostname`, `protocol` and `port`
- the main section of config is read only once at the beginning
- all config files are now read at once and not one by one 
- the `list` command now always shows projects as `hub/user/project`

clime:
Ok, great, there are still some small quirks in the code but they are not very problematic. Anyway, can you, please, provide a copr repo from which I could download the updated dnf-plugins-core package  so that I can test the changes locally? That would help a lot.

clime:
Thanks, will look at it and report back.

praiskup:
I really dislike that I can not do `dnf copr enable redhat/praiskup/ping` now, and I have to type `--hub redhat`.  I know @clime requested that, but that was totally convenient thing.  If the lookup was "try first the hub id and fallback to hostname", I don't see any reason to drop that.

praiskup:
The `sudo dnf copr list` shows `copr.fedorainfracloud.org/praiskup/ping` even if `praiskup/ping` was enabled by `dnf copr enable --hub redhat praiskup/ping`.

praiskup:
Otherwise very good job! :-) I really like this PR.

clime:
```
$ sudo dnf copr enable --hub copr.fedorainfracloud.org clime/rpkg-util
Warning: No such instance 'copr.fedorainfracloud.org' in configuration file. Using the default one instead 'https://copr.fedorainfracloud.org'.
```

We shouldn't be using default hub here but instead:

1) throw an error (tiny bit safer option)

or 

2) interpret `copr.fedorainfracloud.org` as a direct hostname to be used when contacting copr-frontend (and use it with `https://` protocol and 443 port)

That's because if user makes a typo on command-line in a (non-default) hub name, he/she will likely not want the operation to pick the default hub, potentially succeed and enable a completely different copr than the one they actually meant.

-----------

```
$ sudo dnf copr disable something-totally-nonexistent-that-i-didnt-even-enable-before/clime/rpkg-util
Error: No matching repo to modify: copr:something-totally-nonexistent-that-didnt-even-enable-before:clime:rpkg-util.
Repository successfully disabled.
```

Here, a success message was printed (in the end) but there was no such copr in `/etc/yum.repos.d`.

-----------

```
$ dnf copr enable --hub localhost clime/test

You are about to enable a Copr repository. Please note that this
repository is not part of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Do you want to continue? [y/N]: y
Repository successfully enabled.
```

The long message should be displayed only for Fedora Copr. It would be good to change the question  to: `Do you really want to enable localhost/clime/test? [y/N]: y` so that user knows exactly what copr is going to be enabled.

-----------

```
$ sudo dnf copr enable copr.fedoraproject.org/clime/rpkg-util

You are about to enable a Copr repository. Please note that this
...

Do you want to continue? [y/N]: y
Repository successfully enabled.

$ sudo dnf copr enable copr.fedorainfracloud.org/clime/rpkg-util

You are about to enable a Copr repository. Please note that this
...

Do you want to continue? [y/N]: y
Repository successfully enabled.

$ dnf copr list
Repository clime-rpkg-util is listed more than once in the configuration
localhost/clime/test
copr.fedorainfracloud.org/clime/rpkg-util
```

Here I enabled the same copr twice, once as `copr.fedoraproject.org/clime/rpkg-util` and the second time as `copr.fedorainfracloud.org/clime/rpkg-util/`. This produced two different files with the same content, which lead to `Repository clime-rpkg-util is listed more than once in the configuration`. It would be better if we actually construct the resulting filename from ID of the returned copr dnf config. So if frontend returns:

```
[copr:localhost:clime:test]
name=Copr repo for test owned by clime
```

We should read that returned repo ID, just attach `.repo` suffix to it and use that as a filename for the file to be stored at `/etc/yum.repos.d`.

This avoids the ambiguity related to multiple DNS records pointing to the same copr-frontend machine.

-----------

I haven't found any other obvious problems.

-----------

Now to the review of @praiskup:

> The sudo dnf copr list shows copr.fedorainfracloud.org/praiskup/ping even if praiskup/ping was enabled by dnf copr enable --hub redhat praiskup/ping

I suspect that's because the internal copr still returns the old ID format? (`[praiskup-ping]`) I think the behavior is then to assume the Fedora instance because we don't have the required information to determine that it was some other Copr instance.

> I really dislike that I can not do dnf copr enable redhat/praiskup/ping now, and I have to type --hub redhat. I know @clime requested that, but that was totally convenient thing. If the lookup was "try first the hub id and fallback to hostname", I don't see any reason to drop that.

I am probably ok with the hostname fallback if both `--hub <hub_name>` and `<hub_name>/<owner>/<copr>` behave the same way (that would be what's described in the point 2) in the first reported issue in this comment).

Right now, we have `--hub <hub_name>` and `<hostname>/<owner>/<copr>`. The first expression with `--hub` switch references something that has been configured in the hub configs. The second variant uses hostname directly. I think this option is slightly more rigid.

I am ok with the switch to the first variant `--hub <hub_name>` and `<hub_name>/<owner>/<copr>` given that both syntaxes behave the same way and fallback to interpreting `<hub_name>` as hostname if `<hub_name>` could not be matched in the system hub configuration. The output of `dnf copr list` command should stay the same as it is now, however.

But maybe we could also first try to introduce the rigid variant and later if we really find it convenient implement the "hostname fallback". I think that could be a very slightly safer approach here.

praiskup:
NIce review, thanks!  Only few remarks:
@clime wrote
> The long message should be displayed only for Fedora Copr. 

That message needs to be displayed for any copr, since Fedora Project doesn't take responsibility for any of them (not only Fedora copr).

> This avoids the ambiguity related to multiple DNS records pointing to the same copr-frontend machine.

The solution to this problem should be to correctly resolve the DNS, same as web-ui does that.
If there's no http redirect, and the server is "broken" (communicates on two hostnames), dnf
plugin should expect that those are two different servers.

> I suspect that's because the internal copr still returns the old ID format?

Dunno, at least the repo file looks like `copr:copr.devel.redhat.com:praiskup:ping.repo` which seems to be correct, and the config file contains correct "hostname" (copr.devel.redhat.com).  So I think dnf plugin should be aware of the different hostname.

> I am probably ok with the hostname fallback if both --hub <hub_name> and <hub_name>/<owner>/<copr> behave the same way (that would be what's described in the point 2) in the first reported issue in this comment).

That makes sense to me, +1.

dturecek:
Thanks for your reviews.

> ```
> $ sudo dnf copr disable something-totally-nonexistent-that-i-didnt-even-enable-before/clime/rpkg-util
> Error: No matching repo to modify: copr:something-totally-nonexistent-that-didnt-even-enable-before:clime:rpkg-util.
> Repository successfully disabled.
> ```
> Here, a success message was printed (in the end) but there was no such copr in /etc/yum.repos.d.

This is a bug. However, the success message was printed because a repository *was* sucessfully disabled -- you had `clime/rpkg-util` copr in `/etc/yum.repos.d`.

> Dunno, at least the repo file looks like copr:copr.devel.redhat.com:praiskup:ping.repo which seems to be correct, and the config file contains correct "hostname" (copr.devel.redhat.com). So I think dnf plugin should be aware of the different hostname.

That's because the repo filename uses the hostname from the config file. This will be solved by constructing the filename from the repo ID as @clime suggested.

plugins/copr.py
```diff
@@ -217,16 +279,25 @@ def _list_installed_repositories(self, directory, enabled_only, disabled_only):
...
 
-            copr_name = copr.split('-', 1)
-            msg = copr_name[0] + '/' + copr_name[1]
+            if re.match("^copr:", copr):
+                copr_name = copr.rsplit(':', 2)
+                copr_hostname = copr_name[0].split(':', 1)[1]
+                msg = copr_hostname + '/' + copr_name[1] + '/' + copr_name[2]
+            else:
+                copr_name = copr.split('-', 1)
+                msg = self.default_hostname + '/' + copr_name[0] + '/' + copr_name[1]
```
praiskup: Here the 'self.default_hostname' might be wrong, because the `copr` can come from `copr:copr.devel.redhat.com:user:project.repo`, and still we print the default_hostname (fedoraproject.org).  That's because we are loosing the context about the original filename.

Please consider this patch on top of yours please.

```patch
--- /tmp/old.py 2018-08-06 10:27:13.125289239 +0200
+++ /tmp/new.py 2018-08-06 10:52:17.475963068 +0200
@@ -274,17 +274,20 @@
             raise dnf.exceptions.Error(
                 _('Unknown subcommand {}.').format(subcommand))
 
-    def _list_installed_repositories(self, directory, enabled_only, disabled_only):
-        parser = ConfigParser()
 
-        for root, dir, files in os.walk(directory):
-            for file_name in files:
-                match_new = re.match("^copr:" + self.copr_hostname, file_name)
-                match_old = self.copr_url == self.default_url and re.match("^_copr_", file_name)
-                match_any = re.match("^copr:|^_copr_", file_name)
+    def _list_repo_file(self, directory, file_name, enabled_only, disabled_only):
+        match_new = re.match("^copr:" + self.copr_hostname, file_name)
+        match_old = self.copr_url == self.default_url and re.match("^_copr_", file_name)
+        match_any = re.match("^copr:|^_copr_", file_name)
+
+        if self.opts.hub:
+            if not (match_new or match_old):
+                return
+        elif not match_any:
+            return
 
-                if self.opts.hub and (match_new or match_old) or not self.opts.hub and match_any:
-                    parser.read(directory + '/' + file_name)
+        parser = ConfigParser()
+        parser.read(directory + '/' + file_name)
 
         for copr in parser.sections():
             enabled = parser.getboolean(copr, "enabled")
@@ -295,6 +298,10 @@
                 copr_name = copr.rsplit(':', 2)
                 copr_hostname = copr_name[0].split(':', 1)[1]
                 msg = copr_hostname + '/' + copr_name[1] + '/' + copr_name[2]
+            elif re.match("^copr:", file_name):
+                copr_name = copr.split('-', 1)
+                hostname = file_name.split(':')[1]
+                msg = hostname + '/' + copr_name[0] + '/' + copr_name[1]
             else:
                 copr_name = copr.split('-', 1)
                 msg = self.default_hostname + '/' + copr_name[0] + '/' + copr_name[1]
@@ -302,6 +309,11 @@
                 msg += " (disabled)"
             print(msg)
 
+    def _list_installed_repositories(self, directory, enabled_only, disabled_only):
+        for root, dir, files in os.walk(directory):
+            for file_name in files:
+                self._list_repo_file(directory, file_name, enabled_only, disabled_only)
+
     def _list_user_projects(self, user_name):
         # http://copr.fedorainfracloud.org/api/coprs/ignatenkobrain/
         api_path = "/api/coprs/{}/".format(user_name)
```

praiskup:
On Monday, August 6, 2018 9:51:13 AM CEST dturecek wrote:
> That's because the repo filename uses the hostname from the config file. This
> will be solved by constructing the filename from the repo ID as @clime
> suggested.

This basically means that any not-yet-updated copr frontend will break the
listing.  Consider the patch I proposed, which makes it more independent on
frontend version.

plugins/copr.py
clime: Filenames should play no role in the logic. Just repo IDs as the single source of thruth. So I need to be -1 here.

clime:
> This basically means that any not-yet-updated copr frontend will break the listing. Consider the patch I proposed, which makes it more independent on frontend version.

Yes, but there is just one another instance that I know of right now and that can be updated, no?

We talked about this with @dturecek and decided that only repo IDs should play role even though I understand that it is a slight inconvenience if you already have some old coprs from internal instance enabled (you can reenable them though to fix it). Filenames should not be taken into account - a user can copy repo config from web UI and paste it manually into an arbitrarily named config file. It should still work.

praiskup:
On Monday, August 6, 2018 11:07:05 AM CEST Michal Novotný wrote:
> Filenames should play no role in the logic. Just repo IDs as the single source
> of thruth. So I need to be -1 here.

Filenames already play a huge role here, so I don't think it makes sense to have
a buggy implementation backed up by this argument.

praiskup:
On Monday, August 6, 2018 11:12:21 AM CEST Michal Novotný wrote:
> Yes, but there is just one another instance that I know of right now and that
> can be updated, no?

I wish it was that easy with copr; I'm doing my best for quite some time.

> We talked about this with @dturecek and decided that only repo IDs should play
> role even though I understand that it is a slight inconvenience if you already
> have some old coprs from internal instance enabled (you can reenable them
> though to fix it). Filenames should not be taken into account - a user can
> copy repo config from web UI and paste it manually into an arbitrarily named
> config file. It should still work.

As I said, filename plays a role, we are only matching `_copr_*` and `copr:*`
files.  So "arbitrarily named" files won't just work.

You know my preference (it's just tooo trivial to -1 it), but I won't fight more
for it if you prefer broken output over sligth compat 'if' statement.

dturecek:
> Filenames should not be taken into account - a user can copy repo config from web UI and paste it manually into an arbitrarily named config file.

Well, arbitrary names might not work as the list command only reads files that have either `_copr_` or `copr:` prefix.

However, even if we'd use the patch, there would still be the problem with coprs that were enabled in the past as they will be listed with `copr.fedorainfracloud.org` hub.

dturecek:
Since I will be constructing the filename from repo ID, there's a problem with potential Copr frontends that wouldn't use the updated repo ID format. What I'd like to do is that all downloaded repo files with old ID format would be saved to file with old filename `_copr_user-project.repo`. 

Then, the list command can display a note next to these coprs, stating that the hub name might not be correct. Something like this:

```
$ dnf copr list
copr.fedorainfracloud.org/user/project-with-old-repo-id *
copr.fedorainfracloud.org/user1/project-with-new-repo-id
copr.fedorainfracloud.org/user2/project-with-old-repo-id *
* These coprs contain old repo-ID format and the hubs displayed might be incorrect. Re-enable them to solve this.
```

plugins/copr.py
```diff
@@ -103,37 +107,89 @@ def set_argparser(parser):
...
+            copr_plugin_config.read(default_config_file)
+            if copr_plugin_config.has_option('main', 'distribution') and\
+                    copr_plugin_config.has_option('main', 'releasever'):
                 distribution = copr_plugin_config.get('main', 'distribution')
                 releasever = copr_plugin_config.get('main', 'releasever')
                 self.chroot_config = [distribution, releasever]
             else:
                 self.chroot_config = [False, False]
 
+        for filename in os.listdir('{0}/{1}.d'.format(config_path, PLUGIN_CONF)):
```
xsuchy: rather use os.path.join?

plugins/copr.py
```diff
@@ -103,37 +107,89 @@ def set_argparser(parser):
...
+            copr_hub = self.opts.hub
+            copr_plugin_config.read(sorted(config_files, reverse=True))
+            if copr_hub in copr_plugin_config.sections():
+                hostname = protocol = port = None
+                try:
+                    hostname = copr_plugin_config.get(copr_hub, 'hostname')
+                    protocol = copr_plugin_config.get(copr_hub, 'protocol')
+                    port = copr_plugin_config.get(copr_hub, 'port')
+                except NoOptionError:
+                    pass
```
xsuchy: What happens when a protocol is not defined in config but a port is defined?

clime:
> Filenames already play a huge role here, so I don't think it makes sense to have a buggy implementation backed up by this argument.

Filenames might play some role now related to the support of deprecated config format, that's probably fine. But long-term, they should play no role.

praiskup:
On Monday, August 6, 2018 3:46:40 PM CEST Michal Novotný wrote:
> Filenames might play some role now related to the support of deprecated config
> format, that's probably fine. But long-term, they should play no role.

What's now proposed as new thing will be likely deprecated in future (I for
example really don't understand why we create separate repo file for each copr
repostiory, we could - similarly to subscription-manager - maintain on precisely
named repo file, or at least one repo file per hostname).  It is just too much
code and maintenance with the dir traversal...

What I really dislike is any attempt to make the logic dependant on the format
of Repo ID.  Since we won't have control of all copr instances in future, it
would be really naive to expect that all the instances in the wild will follow
our rules WRT Repo-ID naming.

clime:
```
$ cat copr:copr.fedorainfracloud.org:clime:rpkg-util.repo 
[copr:copr.fedorainfracloud.org:clime:rpkg-util]
name=Copr repo for rpkg-util owned by clime
baseurl=https://copr-be.cloud.fedoraproject.org/results/clime/rpkg-util/fedora-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://copr-be.cloud.fedoraproject.org/results/clime/rpkg-util/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
$ dnf copr list
copr.fedorainfracloud.org/clime/rpkg-util
$ sudo mv copr\:copr.fedorainfracloud.org\:clime\:rpkg-util.repo foobar.repo
$ dnf copr list
$
``` 

I've just noticed this issue. If I just rename a valid new-format copr config file, it disappears from listing. That should not happen because filename of repo config file should play no role. @dturecek do you think you could look at that?

clime:
Note that I was in `/etc/yum.repos.d` dir, when doing the `mv` (that's not visible from the log).

xsuchy:
>  We should only list repositories from precisely named repo files, the rest of files is supposed to be manually added - thus not maintained by dnf copr plugin.

+1 Any manual alternation and you are on your own.

clime:
We need only a single source of truth: repo ID.

It needs to be possible that users just copy the repo config from web UI into an arbitrarily named file and it will work for them. There are tutorials on youtube where users show exactly this use-case and it is something that any beginner can easily do on his/her own even without dnf-plugins-core installed.

This is therefore a feature that we are required to have.

clime:
> What's now proposed as new thing will be likely deprecated in future (I for example really don't understand why we create separate repo file for each copr repostiory, we could - similarly to subscription-manager - maintain on precisely named repo file, or at least one repo file per hostname). It is just too much code and maintenance with the dir traversal...

Yup, that's true - "It is just too much code and maintenance with the dir traversal..." That's why we should fetch the repo list (meaning a python list) directly from dnf in the plugin and take into account only those entries that start with 'copr:', '_copr:' or similar well-recognized prefix.

> What I really dislike is any attempt to make the logic dependant on the format of Repo ID. Since we won't have control of all copr instances in future, it would be really naive to expect that all the instances in the wild will follow our rules WRT Repo-ID naming.

Not sure what you mean by naive. The repo ID generation is part of copr-frontend logic. If anyone deploys Copr, he/she will get the expect repo ID format. That's something we can rely on.

praiskup:
Nope ;-) If we want to rely on 'repo ID' format, we have to maintain it ouselves **here in dnf plugins core** (and not rely on the version of any random copr-frontend in the wild).  Period.

- I miss one compat "hack" (but I can submit a PR later)
- since we are doing a rather big change, I'd much rather appreciate if we had only one repo file per copr frontend instance right now - to not do compat for compat in future - say `/etc/yum.repos.d/<hostname>.repo`, I can have a look right after this PR

I think this is basically good to go.

dturecek:
So made the last few changes based on your suggestions.

In the end, I decided to go as @praiskup suggested in the case repo file has repo ID with the old format as I don't see why we should list the default hub even though we have the information about the actual one.

plugins/copr.py
```diff
@@ -212,25 +272,48 @@ def run(self):
...
+            else:
+                copr_name = copr.split('-', 1)
+                msg = self.default_hostname + '/' + copr_name[0] + '/' + copr_name[1]
             if not enabled:
                 msg += " (disabled)"
             print(msg)
 
+    def _list_installed_repositories(self, directory, enabled_only, disabled_only):
+        for root, dir, files in os.walk(directory):
+            for file_name in files:
```
clime: Instead of walking the directory and getting the filenames, you can simply use `self.base.repos` structure, which is a dictionary consisting of `copr_id: <repo object>` pairs and which is already provided by dnf itself. Just select repos that do have a key starting with `_copr:` (or `copr:` but having them prefixed with '_' is quite suitable here) and you can later recognize in `_list_repo_file` method whether they are enabled or not just by reading `<repo object>.enabled` attribute. This is much more simple and faster than parsing the files manually from the plugin and it also solves the above-mentioned problem with the `dnf copr list` command not correctly listing all the enabled copr repos when user just manually copy-pastes the repo configs from web UI.

praiskup:
On Wednesday, August 8, 2018 4:20:11 PM CEST Alois Mahdal wrote:
> Guys, I'm sorry I'm so late, but I'm a bit confused by the hostname fallback.

If you do 'dnf copr enable SOMETHING/jdoe/foo', and the `SOMETHING` is not
id of preconfigured "hub" the code will treat SOMETHING as HOSTNAME (+
default port).

> Doesn't relying on hostnames assume that hostname=hub?

The hostname only partly identifies hub, you are right.  That said, hub is
hostname + port + protocol.  And if we have only hostname we have to
expect some sane defaults in port/protocol.

OT: doesn't 'protocol=https' and 'port=443' seem to be redundant?

> Ie. there won't ever be two hubs per hostname?

Could be, but it's not likely.  I mean, IMO it's OK **by default** expect that
there's one hub on one hostname with reasonable defaults.  But think I
understand you;  the thing is that neither port nor protocol is baked into
repo-ID nor repo file name.  That _is_ a problem.

> Why not require hub definition explicitly like I suggested months ago:
> 
>     dnf copr addhub foo https://..../
>     dnf copr enable foo/joe/utils
>     dnf copr disable foo/joe/utils
>     dnf copr rmhub foo
>
> which would mean that there's no confusion about hub identity, file
> naming is much easier.

The 'addhub'/'rmhub' would be useful, but it is something on top of what's
really blocking us.  I still believe that the configuration file will be
maintained inside RPM (e.g. for internal copr in rhcopr.noarch.rpm) per
copr instance.  In the worst case, the file can be added by administrator
manually.

That said, I'd be +1 here, but not necessarily in this PR.  This PR is
taking too much time, really :-(

> IIRC the only "argument" against it was that @clime's one about the
> instruction given to user; that the user would have to copy/paste two
> commands instead of one, which is IMO pretty much moot.

TBH, I'm neither fan of 'dnf copr enable <hostname>/<user>/<project>'
syntax, but as long as the "hub" (or say "hub ID") takes precedence, I'm
fine to accept a compromise (if somebody believes it is useful).  I
believe that nobody will use it anyway.

Related to this compromise:  I'm neither fan of putting "hostname" into
repo ID and repo filename.  I'd much rather see a 'hub ID' there instead.
Hostnames _are_ changed from time to time (it already happened in Fedora
copr case).  If hostname of copr service changes, we have to change
hostname in `/etc/dnf/plugins/copr.conf` which also means that all the
copr repo files remain **orphaned**.  And we should have an option to
semi-automatically recover from such de-synchronization (e.g.  by `dnf
copr sync-repofiles` in future).  Of course, that would require us to have
inseparably paired hub-config files with repo files - which in turn means
that we have to have hubID somehow baked into the repo file.

So for consistency reasons, I agree with @AloisMahdal - I'm +1 so the
`<HUB>` in `--hub <HUB>` or `<HUB>/<OWNER>/<GROUP>` only means `hubID` (not
hostname).

plugins/copr.py
clime: It obviously does help because we can just use the dictionary keys from `self.base.repos` as the source for hub ID.

clime: `.repofile` attribute can be read also (at the end of the method) but only for backward-compatibility reasons to support the old copr repo files that did not have copr ID in the required form (which would say: "hey, I am a copr repo") but they had a filename starting with `_copr_`. We can still list those copr repos but also display some kind of warning to stderr that will tell users that they can re-enable them to get rid of the warning. Later, we can drop this backward-compatibility logic (and warning) and keep only the one that relies on repo IDs. The reason for relying only on repo IDs is that we should be able to have all the (enabled) copr repos placed in just a single file and `dnf copr list` should still list them all. Placing all copr repos into a single file was already suggested here earlier but this is not something we need to be doing in the plugin. At the same time, `dnf copr list` should still work even if they are placed in the same file, otherwise what command displays is not a correct information about the system.

praiskup: Hmpfs, `dnf copr` should only operate with files added by `dnf copr`.  Doing otherwise would not only be kludge, but also it would be giving info that user didn't ask for.  E.g. if I write `dnf copr list`, I want to see what I added by `dnf copr enable` and nothing else.  Otherwise, neither repo ID nor repo file name has additional semantic.  Yes, there would be technically correct ways to achieve what you wish but I don't think it's really desired.

xsuchy: > if I write dnf copr list, I want to see what I added by dnf copr enable and nothing else. 

+1

When I created this plugin originally. I decided to use `_copr` file prefix as no sane human will decide to use `_` as a prefix. I thought that this will immediately strike everybody that "do not touch this, this is kind of private file".  If you wget something or manually edit some file, then you are on your own. It may work or not. Not our business.

clime:
Having the hostname fallback is important for operations that will not need root privileges. E.g. `dnf copr fetch <hostname>/<owner>/<project>`, which would download all the repository data for the current unprivileged user so that he/she can make an image from downloaded results locally, which also does not require root privileges these days. User can then switch into the image and use and test the rpms built in Copr. It's also not actually a fallback. It is just that you can optionally alias shorter hub name with a certain longer hostname (and you can also change default host and port that way but we can later add command line arguments for that as well so that unprivileged user can alter those things). This argument was already mentioned in this thread. You may argue that this problem could be solved by having per-user hub configuration read e.g. from user's home directory but that still seems a little bit cumbersome when a user can just copy-paste the copr identifier from a Copr frontend page (e.g. from the Quick Enable box) and use that on command line. The main point here is that we need to work primarily with hostnames because frontend is responsible for generating the repo files and it can only know reliably its hostname, not some locally defined hub name. We cannot easily move repo generation to client side because only frontend knows where the actual repo on backend is placed and the type of the repo (currently it's just always `rpm-md`). The whole hub configuration thing is therefore not actually something that would be required for the plugin operation, it is just a nice to have feature. Domain name changes might bring a little bit of trouble in the sense that a user will be then able to have the same copr enabled twice (once for old copr-frontend domain name, the second time for new). I don't think that's that big deal but we may even provide some command like `sync-all`, which would reenable every single copr meaning that it would even replace the "old" copr IDs with new ones (containing the new domain name). This will be possible if the old domain still continues to operate by redirecting to a new one (at least for some time) because you can read the origin for a given copr repo config from the ID itself, which means you are able to construct URL where the new repo config can be fetched from the ID itself. I mostly just summarized what has been all already mentioned in this thread.

praiskup:
I dislike all the hostname vision, sorry.  What's more important now in thIs PR, it makes everything brittle.

Expecting that copr frontend generates valid repo ID is also very bad design issue.  We have no control over that, except that we provide some default upstream template.

praiskup:
Let me please offer a help with this PR;  I'm OK to take over this (keeping the original author with original changes + additional patch) as this might feel a bit tiring.

AloisMahdal:
KISS
====

Look:

    dnf copr addhub foo some://hub.url
    dnf copr enable foo/joe/barn
    dnf copr disable foo/joe/barn
    dnf copr rmhub foo

Each of the above commands have completely clear meaning and the whole thing is consistent and easy to script and delegate.


Filename/ID mapping
===================

This has already been partially solved by dnf:  `_copr_USER-PROJECT.repo`.  This would now change to `_copr_HUB-USER-PROJECT.repo`.

    _copr_netvor-jats.repo => _copr_fedora-netvor-jats.repo


Migration
---------

The migration is not actually that hard: once plugin is upgraded, we could:

 *  warn admin that the old .repo files will be abandoned, then let them rot,
 *  provide them with tool to do the renaming,
 *  rename files for them (even in rpm scriptlet),
 *  warn them (on every `dnf copr` invocation) that there are old ambiguous files and they should be renamed,
 *  actually accept the old naming and assume that the `HUB` here is the default hub.


URL rot (if hub url changes)
============================

If `foo` hub URL changed to, let's say `some://hub.url:8080`

    dnf copr rmhub foo              # Y/N to remove old coprs
    dnf copr addhub foo some://hub.url:8080
    dnf copr enable foo/joe/barn
    ...

Note that baseurl's of repos may, or may not be related to hub URL, but that problem is not created by multiple-hub support.  Removing/re-adding repo is already sort-of the clean thing to do.


Other things unrelated to this PR
=================================

Now, it looks like you want to have some mechanism to get COPR/hub data like repos for an image:

 *  sure, you could have `dnf copr fetch <whatever>`.
 *  sure, you could have it working on user level (after all, it's just a set of HTTP requests),
 *  sure, you could have it --- ideally in different PR, though.

There are some anciliary decisions that you would need to make, but they don't need to block this PR (or bz1478208, for that matter).  For example, are ids recognized?  Are /etc settings merged to users'?

(But then again, I'd be still curious about real use case of this.  If someone wants to make image from repos, sure, but what makes you think they will want to use COPR repos, and if they will why would they want to treat them differently and use "dedicates" fetch utility?)


Bad ideas
=========

Sorry, but no matter what kind of "informed assumption" you are trying to make, **no, hostname does not identify copr**.  That's what URLs are for, and URL has many parts: protocol, host (FQDN? non-TLD, eg. `foo`?, IPv4 addtess? IPv6 address?), port name, path, query.  So trying to supplement id with hostnames means you would have to hard-code almost all of the above, which sounds like pretty much a usability crime.

praiskup:
Agreed with @AloisMahdal, except that `addhub` and `rmhub` commands are not that
important at this point -- for the internal copr instance we will have
hub configuration packaged in special and maintained RPM (so any users will
receive automatic updates of that config file).

dturecek:
I have fixed some errors and added a message when there's an old repo file.
As for me, this is now good to go, as per the discussions both here and during the meetings.

Just to summarize the behavior a little bit:
- We use repo ID provided by frontend to determine the hub.
- If for some reason the repo ID provided by frontend has an old format, we construct the repo filename from the hub provided to the copr plugin.
- When listing a project with an old repo file (`_copr_user-project.repo`), the default hub is assumed and a message about this is printed.
- <hub> is first looked for in the config, and if not found used as is, both in `--hub` and `hub/user/project`.

I have a build here: https://copr.fedorainfracloud.org/coprs/dturecek/dnf-plugins-core/

doc/copr.rst
```diff
@@ -60,6 +60,28 @@ Arguments (copr)
...
+    Enable the ``name/project`` Copr repository from the specified Copr ``hub``,
+    Hub is be specified by its hostname (eg. `copr.fedoraproject.org`).
```
praiskup: s/.../Hub is specified by it's ID or hostname/

praiskup:
Wow, I gave it a try - and I must say that (having the `/etc/dnf/plugins/copr.d/rhel.conf` configuration file on my box) the usability is really awesome.  I wish we had this long time before.

There are things that we can solve later:
- if there were multiple copr frontends on one IP address, we should handle the hostname case (different ports)
- `dnf copr list` could show `hub ID` instead of full hostname (if that particular hostname is configured) to make the output a bit smaller
- we could move per-hub repos into single flie
- and ensure unique format of repo ID
.. but **none** of those ^^ is work for this PR.  Good job done, thanks @dturecek.

doc/copr.rst
```diff
@@ -60,6 +60,28 @@ Arguments (copr)
...
+
+``--hub Copr``
+    Specify a Copr hub to use. Default is the Fedora Copr: ``https://copr.fedoraproject.org``.
+
+--------------------
+Configuration (copr)
+--------------------
+
+``/etc/dnf/plugins/copr.conf``
+``/etc/dnf/plugins/copr/*.conf``
```
xsuchy: /copr.d/ should be here.

PR #254, opened 2018-04-06
Do not sink all rpm's stderr (RhBug:1547897)

m-blaha:
Redirect rpm library error output do dnf.rpm log.

https://bugzilla.redhat.com/show_bug.cgi?id=1547897


j-mracek:
@rh-atomic-bot r+

PR #255, opened 2018-04-11
WIP repoclosure enhancements

m-blaha:
- remove --check option
- return --lookaside option
- option for json output

https://bugzilla.redhat.com/show_bug.cgi?id=1561435


Conan-Kudo:
@m-blaha Why remove `--check`?

m-blaha:
@Conan-Kudo I was thinking about making it more consistent with the rest of dnf and (for me) more straightforward - lookaside repos are specified explicitly (not implicitly as a diference of sets --repoid - --check). Using --repoid in other dnf commands means you want to work only with these specific repos. Here the actual work was done on --check repos.

PR #258, opened 2018-05-04
repoclosure on non-existing package ends with error (RhBug:1571622)

m-blaha:
When the repoclosure command gets non-existing package in the --pkg option,
it shows an error message and returns error code.

https://bugzilla.redhat.com/show_bug.cgi?id=1571622

### DIFF
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -105,7 +105,10 @@ class RepoClosureCommand(dnf.cli.Command):
             pkglist_q = self.base.sack.query().filter(empty=True)
             errors = []
             for pkg in self.opts.pkglist:
-                pkg_q = to_check.filter(name=pkg)
+                subj = dnf.subject.Subject(pkg)
+                pkg_q = to_check.intersection(
+                    subj.get_best_query(self.base.sack, with_nevra=True,
+                                        with_provides=False, with_filenames=False))
                 if pkg_q:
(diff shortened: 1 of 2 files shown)


plugins/repoclosure.py
```diff
@@ -103,6 +103,9 @@ def _get_unresolved(self, arch=None):
 
         if self.opts.pkglist:
             to_check = to_check.filter(name=self.opts.pkglist)
```
j-mracek: Believe that each of argument should be checked separately. Otherwise the command like "dnf repoclosure --pkg dnf --pkg notMe" will exit with 0, even if package  "notMe" is not present.

plugins/repoclosure.py
```diff
@@ -102,16 +102,27 @@ def _get_unresolved(self, arch=None):
             to_check = self.base.sack.query().available()
 
         if self.opts.pkglist:
-            to_check = to_check.filter(name=self.opts.pkglist)
+            pkglist_q = self.base.sack.query().filter(empty=True)
+            errors = []
+            for pkg in self.opts.pkglist:
+                pkg_q = to_check.filter(name=pkg)
```
j-mracek: According to doc:
``--pkg <pkg``
    Check closure for this package only.

pkg could be not only pkg name. What about to use a subject with_nevra only?

j-mracek:
It looks good

PR #261, opened 2018-05-10
copr: fix bz#1576736

clime:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1576736.

Details about name mangling: https://docs.python.org/3/reference/expressions.html?highlight=name%20mangling#atom-identifiers

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -36,13 +36,13 @@ import re
 # If that fails, attempt to import the deprecated implementation
 # from the platform module.
 try:
-    from distro import linux_distribution as linux_distro
+    from distro import linux_distribution
 except ImportError:
     try:
-        from platform import linux_distribution as linux_distro
+        from platform import linux_distribution
     except ImportError:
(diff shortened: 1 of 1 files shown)


xsuchy:
:+1:

plugins/copr.py
```diff
@@ -36,13 +36,13 @@
 # If that fails, attempt to import the deprecated implementation
 # from the platform module.
 try:
-    from distro import linux_distribution as __linux_distro
+    from distro import linux_distribution as linux_distro
```
m-blaha: How about removing this `as linux_distro` import and rename the function on the line 45 to `def linux_distribution()`?

j-mracek: @clime @xsuchy I am neutral here. What do you think?

clime: I made the changes as requested.

m-blaha:
@rh-atomic-bot r+

PR #262, opened 2018-05-16
Add provide of yum-utils

j-mracek:
Due to conflict of dnf-2.8.9+ with yum, it is nice to provide yum-utils due to
increase compatibility and not breaking work-flows.

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -138,6 +138,9 @@ reposync commands. Additionally provides generate_completion_cache passive plugi
 
 %package -n dnf-utils
 Conflicts:      yum-utils < 1.1.31-513
+%if 0%{?rhel} != 7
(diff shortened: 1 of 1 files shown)


dnf-plugins-core.spec
```diff
@@ -138,6 +138,9 @@ reposync commands. Additionally provides generate_completion_cache passive plugi
 
 %package -n dnf-utils
 Conflicts:      yum-utils < 1.1.31-513
+%if 0%{?rhel} != 7
```
ignatenkobrain: this is also true for Fedora, so I don't think that this is correct.

dmach: It was actually meant that way.
AFAIK most of the python2 based software is going away in Fedora,
we'd also like to remove yum3 stack from Fedora 29.

PR #264, opened 2018-05-18
Bump version to 3.0.0

dmach:


dmach:
Includes patches from PR#250, we should merge that first.

PR #267, opened 2018-06-15
Make progressbar of download packages sensitive to -q option

j-mracek:
Additionally it allows to download packages from url

Requires: https://github.com/rpm-software-management/dnf/pull/1110


m-blaha:
tested locally

PR #268, opened 2018-06-29
Revert "Reduce query creation for leave plugin about 18x"

esmil:
This reverts commit 786b1c5e7180eb6b9d5af8245be69f69d59cad0e.

This optimization seems to be broken in latest rawhide.

### DIFF
--- a/plugins/leaves.py
+++ b/plugins/leaves.py
@@ -51,15 +51,11 @@ class LeavesCommand(dnf.cli.Command):
             rdepends.append([])
 
         for i, pkg in enumerate(packages):
-            req_list = []
             for req in pkg.requires:
                 sreq = str(req)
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
It would be nice if you could explain an issue.

esmil:
Well if you read the four lines reverted it changes the code to lookup installed packages that resolve a list of dependencies rather than lookup up each dependency individually. That doesn't work. Which means that the dependency graph doesn't have all right edges, and dnf leaves ends up listing almost all installed packages rather the leaves.

esmil:
I mean you can try leaves plugin on any rawhide and compare the results with and without this revert and see the difference.

j-mracek:
Thanks for testing and patch. The new set is really incorrect

esmil:
Btw. this also seems to be broken in Fedora 29. Hope you get to fix it before it is released.

PR #270, opened 2018-07-11
Do not distribute dnf-utils subpackage for RHEL 7

m-blaha:


dnf-plugins-core.spec
```diff
@@ -15,6 +15,12 @@
 %bcond_without python2
 %endif
 
+%if ! 0%{?rhel} || 0%{?rhel} > 7
```
Conan-Kudo: This check is wrong. It should be inverted:

```spec
%if 0%{?rhel} && 0%{?rhel} < 8
%bcond_with dnfutils
%else
%bcond_without dnfutils
%endif
```

m-blaha: Why is it wrong? Your condition is just negation with switched branches. Or am I missing something?

PR #283, opened 2018-09-16
Rename package-spec to pkg-spec for consistency

sedrubal:
The versionlock plugin was the only plugin that used `package-spec`. All
other plugins used `pkg-spec`.

This makes it hard to search for it.

### DIFF
--- a/doc/versionlock.rst
+++ b/doc/versionlock.rst
@@ -51,31 +51,31 @@ operations like `repoquery`, `list`, `info`, etc.
 Synopsis
 --------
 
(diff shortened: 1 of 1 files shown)


sedrubal:
ah, it isn't that easy. The DNF docs use `package-spec` more often. I think it is more important to link to the corresponding explanation on this page (https://dnf.readthedocs.io/en/latest/command_ref.html#specifying-packages-label) every time such a keyword is used. But therefore I fear I can't help...

PR #287, opened 2018-09-21
Download only packages with unique NEVRAs (RhBug:1612874)

j-mracek:
Downloading multiple packages with same NEVRA into same location results
in fail during checksum check. Now only one package with same NEVRA will
be downloaded.

https://bugzilla.redhat.com/show_bug.cgi?id=1612874


m-blaha:
all tests passed.

PR #297, opened 2018-10-16
copr: fix enabling Rawhide repository (RhBug:1628888)

dturecek:
This fixes the second issue in [Bug1628888](https://bugzilla.redhat.com/show_bug.cgi?id=1628888) when copr didn't recognize a Rawhide repo and tried to use Fedora 30 instead.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -35,8 +35,10 @@ import re
 # If that fails, attempt to import the deprecated implementation
 # from the platform module.
 try:
-    from distro import linux_distribution
+    from distro import linux_distribution, os_release_attr
 except ImportError:
+    def os_release_attr(_):
+        return ""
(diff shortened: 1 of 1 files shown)


plugins/copr.py
```diff
@@ -35,8 +35,10 @@
 # If that fails, attempt to import the deprecated implementation
 # from the platform module.
 try:
-    from distro import linux_distribution
+    from distro import linux_distribution, os_release_attr
 except ImportError:
+    def os_release_attr():
```
xsuchy: this should be:
  def os_release_attr(_):
otherwise, you will get:
TypeError: os_release_attr() takes 0 positional arguments but 1 was given

xsuchy:
Thank you.

PR #298, opened 2018-10-17
Revert "spec: bcond: dnfutils -> yumcompatibility"

dmnks:
I missed the fact that, for Fedora <= 29, dnf-utils was not built any
more with the original commit.  My mistake.  Let's revert it now as a
quick fix and consider a better fix later.

This reverts commit 9d07c0da599448e963e214ef50160b8b1f164b96.


j-mracek:
It works. Thanks.

PR #299, opened 2018-10-24
[config-manager] Add --enable/--disable options (RhBug:1613045)

Kratochvilova:
Requires https://github.com/rpm-software-management/dnf/pull/1248

https://bugzilla.redhat.com/show_bug.cgi?id=1613045
It is not possible to add --enable/--disable options directly to the plugin's argparser, because the common options are resolved first and these options would be ambiguous.
Hence, the options are added to the main parser, although only used with the config-manager command.
(This workaround is similar to the one for --destdir option.)

Version bump, since change in dnf is required.


j-mracek:
@rh-atomic-bot r+

j-mracek:
Thanks for nice job.

PR #302, opened 2018-11-06
copr: fix disabling repos with old filenames and repo IDs (RhBug:1644563)

dturecek:
This gets rid of the traceback and fixes disabling repos with the old format.


xsuchy:
+1

PR #308, opened 2018-11-29
[versionlock] Accept more NEVRA forms in the deletion code (RhBug:165…

dhgutteridge:
…4529)

https://bugzilla.redhat.com/show_bug.cgi?id=1654529


dmach:
Looks good to me, it only adds couple more patterns to the list.

PR #319, opened 2019-01-16
Not download src without --source (RhBug:1666648)

j-mracek:
"dnf download dnf" will not download src rpm with this patch.

It makes behavior similar with yumdowloader.

https://bugzilla.redhat.com/show_bug.cgi?id=1666648


j-mracek:
@rh-atomic-bot try

m-blaha:
tested locally

PR #320, opened 2019-01-23
[builddep] Report all rpm errors (RhBug:1663619,1658292)

m-blaha:
https://bugzilla.redhat.com/show_bug.cgi?id=1663619

### DIFF
--- a/plugins/builddep.py
+++ b/plugins/builddep.py
@@ -21,42 +21,16 @@
 
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from dnfpluginscore import _, logger, rpm_logger
+from dnfpluginscore import _, logger
 
 import argparse
 import dnf
 import dnf.cli
 import dnf.exceptions
 import dnf.rpm.transaction
-import functools
-import logging
+import dnf.yum.rpmtrans
 import os
 import rpm
-
-
-class redirect_rpm_logging(object):
-    def __init__(self):
-        self.sink = None
-
-    def __call__(self, func):
-        @functools.wraps(func)
-        def inner(*args, **kwds):
-            with self:
-                return func(*args, **kwds)
-        return inner
-
-    def __enter__(self):
-        for handler in rpm_logger.handlers:
-            if isinstance(handler, logging.FileHandler):
-                rpm.setLogFile(handler.stream)
-                break
-        else:
-            self.sink = open('/dev/null', 'w')
-            rpm.setLogFile(self.sink)
-
-    def __exit__(self, exc_type, exc, exc_tb):
-        if self.sink:
-            self.sink.close()
 
 
 @dnf.plugin.register_command
@@ -91,6 +65,10 @@ class BuildDepCommand(dnf.cli.Command):
         ptype.add_argument('--srpm', action='store_true',
                             help=_('treat commandline arguments as source rpm'))
 
+    def pre_configure(self):
+        if not self.opts.rpmverbosity:
+            self.opts.rpmverbosity = 'error'
+
     def configure(self):
         demands = self.cli.demands
         demands.available_repos = True
@@ -107,8 +85,8 @@ class BuildDepCommand(dnf.cli.Command):
                     self.base.repos.enable_source_repos()
                     break
 
-    @redirect_rpm_logging()
     def run(self):
+        rpmlog = dnf.yum.rpmtrans.RPMTransaction(self.base)
         # Push user-supplied macro definitions for spec parsing
         for macro in self.opts.define:
             rpm.addMacro(macro[0], macro[1])
@@ -127,6 +105,8 @@ class BuildDepCommand(dnf.cli.Command):
                 else:
                     self._remote_deps(pkgspec)
             except dnf.exceptions.Error as e:
+                for line in rpmlog.messages():
+                    logger.error(_("RPM: {}").format(line))
                 logger.error(e)
                 pkg_errors = True
 
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -1,4 +1,4 @@
-%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 4.2.1}
+%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 4.2.8}
 %global dnf_plugins_extra 2.0.0
 %global hawkey_version 0.34.0
 


pmatilai:
I think it'd be better to just always let rpm's messages through, or at least always capture them in a way that can be replayed in case of error. Asking user to rerun with -v just to get the error messages that always is there is a bit meh.

The rpm message sink introduced in commit 2d3dc762aeb56fc4a44db5bc9003856ad32c88dc was IMO always a rather strange solution to https://bugzilla.redhat.com/show_bug.cgi?id=1074461, I only provided the /dev/null sink as an example of controlling rpm's log, not as the right thing to do.

pmatilai:
One crude but simple possibility would be opening a temporary file for the rpm log (instead of /dev/null) and dumping out in case of parse errors.

lukash:
@pmatilai actually, from what it says in the bz: https://bugzilla.redhat.com/show_bug.cgi?id=1663619, the user is already getting an error message from rpm: "can't parse specfile". The problem is the message is not descriptive enough. The best solution would be if rpm threw an exception that was descriptive enough.

I find the user re-running with the -v option to get additional details valid, I don't think we necessarily need to dump the (perhaps unrelated) verbose output in case of an error if the user didn't ask for it. What we really need are error messages that are descriptive enough.

Is improving the "can't parse specfile" message something you could do perhaps? :smile:

plugins/builddep.py
```diff
@@ -107,35 +103,38 @@ def configure(self):
...
-                logger.error(e)
-                pkg_errors = True
-
-        # Pop user macros so they don't affect future rpm calls
-        for macro in self.opts.define:
-            rpm.delMacro(macro[0])
-
-        if pkg_errors:
-            raise dnf.exceptions.Error(_("Some packages could not be found."))
+        with redirect_rpm_logging(self.opts.verbose) as rpmlog:
```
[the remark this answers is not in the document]
lukash: This and the Flake8 error below should obviously be fixed in this commit and not in the subsequent one :smirk:

plugins/builddep.py
```diff
@@ -21,38 +21,33 @@
...
 import dnf
 import dnf.cli
 import dnf.exceptions
 import dnf.rpm.transaction
-import functools
-import logging
-import os
-import rpm
 
+from dnfpluginscore import _, logger, rpm_logger
```
lukash: The rearrangement of imports should go into a separate commit. e.g. imagine needing to revert this in the future and someone else adding/removing imports in between. Not only you'd revert the rearrangement unnecessarily, you'd also have an unnecessary conflict on the revert...

plugins/builddep.py
```diff
@@ -191,8 +191,9 @@ def _spec_deps(self, spec_fn):
         try:
             spec = rpm.spec(spec_fn)
         except ValueError as ex:
-            msg = _("Failed to open: '%s', not a valid spec file: %s") % (
-                    spec_fn, ex)
+            self.rpm_error_occured = True
+            msg = _("Failed to open: '%s', not a valid spec file: %s.") % (
+                    spec_fn, str(ex).strip())
```
lukash: FWIW I'd change the message string to:
```
"Failed to open spec file '%s': %s"
```

The "not a valid spec file" is redundant and it's just simpler this way.

lukash:
@m-blaha I'd welcome some description of the changes in the commit message. Along the lines of "Adding the verbose argument to the redirect_rpm_logging class and not redirecting stderr to /dev/null if it's true. Due to it being parametrized, the class cannot be a decorator anymore, change it to a context manager and use a with block." It would help people not as quick understanding the code as myself while reading the commit :smile:

pmatilai:
@lukash , https://bugzilla.redhat.com/show_bug.cgi?id=1663619 just doesn't include the rpm output of the same thing, https://bugzilla.redhat.com/show_bug.cgi?id=1547897 is more useful for understanding the issue. But in short, compare these two:

```
# dnf builddep /home/build/rpmbuild/SPECS/systemd.spec 
Last metadata expiration check: 9:40:09 ago on Wed 21 Feb 2018 15:32:36 UTC.
Failed to open: '/home/build/rpmbuild/SPECS/systemd.spec', not a valid spec file: can't parse specfile
```

```
[pmatilai🎩︎sopuli systemd]$ rpmspec -q systemd.spec 
error: Unable to open /home/pmatilai/rpmbuild/SOURCES/triggers.systemd: No such file or directory
error: query of specfile systemd.spec failed, can't parse
```

In rpm the detailed information goes to rpm log, the "can't parse" thing is just a yes/no result code.

pmatilai:
FWIW, dnf showing uninformative/misleading errors instead of those from rpm is actually a common theme, for example https://bugzilla.redhat.com/show_bug.cgi?id=1677199 is exactly the same thing in a different setting. And unless the base approach to rpm errors is changed one way or the other, there'll be others. So instead of trying to find some minor tweak to this particular case, I'd suggest looking at the broader picture.

lukash:
@pmatilai I'm afraid I don't understand your point... In the comparison you presented the two outputs contain basically the same information to me. I assume the detailed information can be found in the log for `rpmspec` and is redirected to `/dev/null` for 'dnf builddep`.

> FWIW, dnf showing uninformative/misleading errors instead of those from rpm is actually a common theme...

Well, I'm not going to argue that's not true, I don't really know. But what I see in this case is `dnf` actually forwarding the rather unhelpful message `rpm` gives it.

I agree we should be looking at the broader picture, but at the same time if you want helpful error messages, you need to fix them one by one. It seems to be surprisingly hard to do this right. I spend significant time and effort in my reviews trying to make people write helpful and complete error messages. No way around it I guess.

For the broader picture then, is your idea to solve this to blanket-print the rpm output log in case of an error? Again, I'm rather unfamiliar with the specifics here, namely the actual content of the log output... But it does seem to me to be a somewhat lazy solution of "here, have a log dump of everything we've done and look up the error for yourself". I might be wrong. But I still prefer a single error string clearly describing the error... And then the `-v` for explicit verbose output. I do realize the parsing of the spec file in particular can print multiple errors, so perhaps multiline output is actually desirable and a single error string not sufficient? How about putting the first parsing error to the error string?

pmatilai:
> I'm afraid I don't understand your point... In the comparison you presented the two outputs contain basically the same information to me. I assume the detailed information can be found in the log for rpmspec and is redirected to /dev/null for 'dnf builddep`.

Um, what? Look closer. The rpmspec case shows the exact reason *why* it fails (a related file not found), allowing the user to understand and address the situation, dnf builddep output does not. Which is the whole point here. An error message that doesn't say *why* it failed is rather useless. In rpm, the why is practically always in the log, and that's so deeply rooted that there's little we can do about it.

You say you're not familiar with the contents of rpm log output, I'd suggest starting there - you can't really make a informed decisions if you don't know what you're talking about. Rpm does not log warning and error-level messages for fun, and dnf really, really should let *all* error-level messages from rpm to come through to the user (and preferably warnings too) one way or the other.

lukash:
@m-blaha regardless of the discussion with Panu, unless there are objections to this particular PR, I think it's an improvement that makes sense and want to merge it if you address my comments...

One more question, this is specific to the builddep plugin, do we have nothing similar for other commands and do we not have similar problems there too? Thanks.

pmatilai:
The python bindings to the rpm log are certainly spartan and don't give you much to work with, but like you noted one possibility is grabbing the log (which dnf already does...) and on errors, display any error-level messages since the last time you did that. That way you can control exactly when to output and where, and you could eg prefix the lines with ```rpm:``` or something like that to make it clear where its coming from. Out of sync is not an issue because there *wont be* much output there.

Further down the road, as libdnf is in C++ it has access to various things that aren't exported to python, including custom logging callbacks.

pmatilai:
One more thing...
> You can't just pass through all the messages - it would be noisy and potentially mess with dnf's progress bars etc.

There's still this strange assumption that rpm spews out tons of gibbedygoo into the log during its operation, it doesn't. This is not like running gnome/gtk applications from a terminal you know :smile:  Rpm is mostly of the old-school unix style where no news is good news - rpm never logs warnings or errors without a good reason, and wiping those under the carpet is disservice to the user.

m-blaha:
Sorry for delay. I thought about this PR again, and I think Panu is right that hiding rpm's errors and asking user to add -v is a bit cumbersome. On top of that in verbose mode the output of rpm is not even going to dnf.rpm.log. I'm marking this as "work in progress".

pmatilai:
> Just wondering, this looks like quite a complex solution to a situation which we should be having in a lot of places, but it's just applied here in a single plugin... Is this just one special case?

This is by no means an isolated case, it happens all over the place so maybe it should be dealt with in dnf core really.

lukash:
I just happened to come across https://bugzilla.redhat.com/show_bug.cgi?id=1677199.

Should we use this approach in dnf and provide it as an API for plugins too, to handle all the cases in a unified way?

m-blaha:
Thanks for pointing this bug out. It certainly is worth investigating.

m-blaha:
I created PR solving the bz1677199. If this got merged, the same solution could be used also for builddep bug: https://github.com/rpm-software-management/dnf/pull/1430

m-blaha:
Rewritten to use for catching rpm log messages the same RPMTransaction() object as in base.do_transaction()

m-blaha:
> Are you sure you want to translate this string?
> 
> and should it not be:
> 
> ```python
> _("RPM: {}").format(line)
> ```
> 
> instead of:
> 
> ```python
> _("RPM: {}".format(line))
> ```

Thanks for catching this!
The translation is there merely for consistency - all error messages are translated. And may be for some languages like Japanese it could make some sense.

PR #322, opened 2019-02-21
Remove dependency on iniparse

jrohel:
Plugins get parser from DNF.
Don't rely on iniparse module and don't import it.


lukash:
It would be nice to actually mention the reason for removing it in the commit message.

j-mracek:
Excellent

PR #327, opened 2019-02-28
Smooth upgrade path for yum-utils

dmnks:
Changes the `dnf-utils` subpackage name to `yum-utils` for Fedora 31, ensuring the removal of the original `yum-utils` as well as consistency with RHEL.

For more details, please see the commit messages.

[Similar change for the dnf-yum -> yum transition.](https://github.com/rpm-software-management/dnf/pull/1337)

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -21,6 +21,11 @@
 %else
 %bcond_without dnfutils
 %endif
+%global utils_subpackage_name yum-utils
+%if (0%{?fedora} && 0%{?fedora} < 31) || (0%{?rhel} && 0%{?rhel} < 8)
+# Avoid name conflict with yum-utils < 4
+%global utils_subpackage_name dnf-utils
+%endif
 
 Name:           dnf-plugins-core
 Version:        4.0.7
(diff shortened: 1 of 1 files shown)


dnf-plugins-core.spec
```diff
@@ -159,7 +159,9 @@ Additionally provides generate_completion_cache passive plugin.
 
 %if %{with dnfutils}
 %package -n dnf-utils
-Conflicts:      yum-utils < 1.1.31-513
+# we need to specify the version since we would conflict with ourselves
+# otherwise (see the Provides below)
+Conflicts:      yum-utils <= 1.1.31-519
```
remicollet: Could be simpler to 
Conflicts: yum-utils < 1.1.32

remicollet: If retired from F31, rather

```
%if 0%{?fedora} >= 31
Obsoletes: yum-utils < 1.1.32
Provides: yum-utils = %{version}
%else
Conflicts: yum-utils < 1.1.32
%endif
```

carlwgeorge: https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages

The guidelines suggest just bumping the release one higher, not the version. 
 I think `yum-utils < 1.1.31-520` is the right choice.

dmnks:
This change would affect RHEL-8 so we have to discuss that internally first. I've closed this PR for the time being.

PR #331, opened 2019-04-29
Fix download of src when it is not the latest

j-mracek:
Additionally it add support for epoch in nevra provided with --source
option. Before patch the epoch was ignored when --source option was
used.


j-mracek:
The issue was discovered in https://github.com/rpm-software-management/ci-dnf-stack/pull/535

j-mracek:
@rh-atomic-bot try

Conan-Kudo:
@rh-atomic-bot retry

Conan-Kudo:
@rh-atomic-bot r+

PR #332, opened 2019-05-07
[reposync] Enable timestamp preserving for downloaded data (RhBug:1688537)

m-blaha:
Added new --remote-time command line switch. With this switch on
the reposync command will try to set the timestamps of the downloaded
files to those on the remote side.

https://bugzilla.redhat.com/show_bug.cgi?id=1688537

Requires https://github.com/rpm-software-management/libdnf/pull/725

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -1,6 +1,6 @@
 %{?!dnf_lowest_compatible: %global dnf_lowest_compatible 4.2.1}
 %global dnf_plugins_extra 2.0.0
-%global hawkey_version 0.7.0
+%global hawkey_version 0.34.0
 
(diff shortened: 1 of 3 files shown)


jrohel:
@rh-atomic-bot r+

jamescassell:
Thanks for adding this!  It seems to be effective for all but the *.zck repodata... if I `curl -LRO` the `*.zck` repodata file, it keeps the server date, so it should be possible to fix that here, too. (or in librepo; not sure where exactly the fix would be)

m-blaha:
@jamescassell  would you mind filing a new bug on dnf-plugins-core component for this feature? https://bugzilla.redhat.com/enter_bug.cgi?product=Fedora

PR #334, opened 2019-05-15
Set the cost of _dnf_local repo to 500, to make it preferred to normal repos

hedayat:
Currently, _dnf_local repo is only preferred by name, but it'll not always work; e.g. when a repository name starts with capital letters. 
Instead of relying on alphabetical preference, which (I think) is not even guaranteed, we set its cost to 500 so that it is preferred to normal repositories which has the cost of 1000.


Conan-Kudo:
@rh-atomic-bot r+

PR #336, opened 2019-05-16
copr: support 'dnf copr enable fedora-stg/user/project'

praiskup:


j-mracek:
Please could you enhance the commit message? The important part is why you want to make the change or what's the benefit of additional server location there.

praiskup:
Sorry, I forgot about this, it was rather blind shot...

We discussed this in copr team, and we (team members) would find this useful;  but it is probably not useful for general audience.  If you disagree, let me know - I'd make the commit nicer.

PR #341, opened 2019-06-06
needs-restarting: add --reboothint option

dmnks:
This is the original implementation from the needs-restarting script in
yum-utils, with a few tweaks (e.g. no versions are printed to keep
things simple).

Original RHEL-7 request:
https://bugzilla.redhat.com/show_bug.cgi?id=1192946

New RHEL-8 request:
https://bugzilla.redhat.com/show_bug.cgi?id=1639468

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -35,6 +35,12 @@ import re
 import stat
 
 
+# For which package updates we should recommend a reboot
+# Mostly taken from https://access.redhat.com/solutions/27943
+NEED_REBOOT = ['kernel', 'glibc', 'linux-firmware', 'systemd', 'dbus',
+               'dbus-broker', 'dbus-daemon']
+
+
 def list_opened_files(uid):
     for (pid, smaps) in list_smaps():
         try:
@@ -169,7 +175,10 @@ class NeedsRestartingCommand(dnf.cli.Command):
     @staticmethod
     def set_argparser(parser):
         parser.add_argument('-u', '--useronly', action='store_true',
-                        help=_("only consider this user's processes"))
+                            help=_("only consider this user's processes"))
+        parser.add_argument('-r', '--reboothint', action='store_true',
+                            help=_("only report whether a reboot is required "
+                                   "(exit code 1) or not (exit code 0)"))
 
     def configure(self):
         demands = self.cli.demands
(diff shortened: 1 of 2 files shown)


plugins/needs_restarting.py
```diff
@@ -35,6 +35,11 @@
 import stat
 
 
+# For which package updates we should recommend a reboot
+# Taken from https://access.redhat.com/solutions/27943
+NEED_REBOOT = ['kernel', 'glibc', 'linux-firmware', 'systemd', 'dbus']
```
Conan-Kudo: `dbus-broker` and `dbus-daemon` are also probably ones that belong on this list...

plugins/needs_restarting.py
```diff
@@ -180,6 +188,27 @@ def run(self):
...
 
+        if self.opts.reboothint:
+            need_reboot = set()
+            installed = self.base.sack.query().installed()
+            for pkg in installed.filter(name=NEED_REBOOT):
+                if pkg.installtime > process_start.boot_time:
+                    need_reboot.add(pkg.name)
+            if need_reboot:
+                print('Core libraries or services have been updated '
+                      'since boot-up:')
```
m-blaha: All these static strings should be marked for translation using _()

dmnks: Good point, thanks

plugins/needs_restarting.py
dmnks: Those are not mentioned explicitly on the KB page (only "dbus" is), but yeah, they seem to be quite essential so I'll include them. In any case, we should think about a more flexible way in the future as having a hard-coded list doesn't scale well :)

dmnks: Also, I'd like to revisit this plugin later as it deserves more love and could be more useful and user-friendly. But that's for later.

Conan-Kudo:
@dmnks A more flexible solution might be something like what `urpmi` does: it reads for packages in the transaction that contain the Provides statement: `should-restart = system`. That lets it scale a lot better.

dmnks:
That's really nice, I like that.

The only downside I can see is that this would require an action from packagers. However, there's just a handful of them at the moment so it shouldn't be too hard. And it's also their domain of expertise (whether to reboot or not), after all :)

It's also worth mentioning that RH errata have been shipping (for a few years) with a special flag (in the updateinfo data) with a similar meaning, i.e. whether a reboot is required. Neither yum or dnf have been adapted to read that flag so far. I can see the benefit here, though: sometimes, you may only need reboots after a specific set of packages is updated (as opposed to one specific package). You can't model that with the simple Provides flag.

Nevertheless, a lot of interesting thoughts here. Let's see where we can take this in the near future.

Conan-Kudo:
@dmnks It may be worth it to support both models. The issue with updateinfo is that it's only read when updateinfo is specifically requested (that is, as part of an advisory request in dnf).

dmnks:
Right. I agree; both ways should be combined somehow (perhaps accessible via a single API query or something).

plugins/needs_restarting.py
dmnks: Note to self: the plugin could also be renamed to "ps" for simplicity (and consistency with zypper).

PR #343, opened 2019-06-26
Refactor and fix the behavior of debuginfo-install

lukash:
Basically rewrites the plugin to:
* Install the reuqested version if specified on the command line

* Install the version of the installed base package if it's installed

* Install the latest available debuginfo version of packages otherwise

* Properly install both debuginfo and debugsource packages where
available

* Print informative messages for packages that do not have debuginfo

* More effectively query and install the packages in one step

https://bugzilla.redhat.com/show_bug.cgi?id=1586059
https://bugzilla.redhat.com/show_bug.cgi?id=1586084

---

Requires this: https://github.com/rpm-software-management/dnf/pull/1425
The tests are here: https://github.com/rpm-software-management/ci-dnf-stack/pull/566

### DIFF
--- a/plugins/debuginfo-install.py
+++ b/plugins/debuginfo-install.py
@@ -2,7 +2,7 @@
 # Install the debuginfo of packages and their dependencies to debug this package.
 #
 # Copyright (C) 2014 Igor Gnatenko
-# Copyright (C) 2014 Red Hat
+# Copyright (C) 2014-2019 Red Hat
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions of
@@ -22,8 +22,7 @@
 from dnfpluginscore import _, logger
 
 import dnf
-import dnf.cli
-import dnf.subject
+from dnf.package import Package
 
 class DebuginfoInstall(dnf.Plugin):
     """DNF plugin supplying the 'debuginfo-install' command."""
@@ -46,8 +45,7 @@ class DebuginfoInstall(dnf.Plugin):
 
         if autoupdate:
             # allow update of already installed debuginfo packages
-            dbginfo = dnf.sack._rpmdb_sack(self.base).query().filter(
-                                                name__glob="*-debuginfo")
+            dbginfo = dnf.sack._rpmdb_sack(self.base).query().filterm(name__glob="*-debuginfo")
             if len(dbginfo):
                 self.base.repos.enable_debug_repos()
 
@@ -57,9 +55,13 @@ class DebuginfoInstallCommand(dnf.cli.Command):
     aliases = ("debuginfo-install",)
     summary = _('install debuginfo packages')
 
-    dbgdone = []
-    packages = None
-    packages_available = None
+    def __init__(self, cli):
+        super(DebuginfoInstallCommand, self).__init__(cli)
+
+        self.available_debuginfo_missing = set()
+        self.available_debugsource_missing = set()
+        self.installed_debuginfo_missing = set()
+        self.installed_debugsource_missing = set()
 
     @staticmethod
     def set_argparser(parser):
@@ -74,59 +76,173 @@ class DebuginfoInstallCommand(dnf.cli.Command):
         self.base.repos.enable_debug_repos()
 
     def run(self):
-        self.packages = self.base.sack.query()
-        self.packages_available = self.packages.available()
         errors_spec = []
 
-        for pkgspec in self.opts.package:
-            package_query = sorted(dnf.subject.Subject(pkgspec).get_best_query(
-                self.cli.base.sack).filter(arch__neq='src'), reverse=True)
-            if not package_query:
-                msg = _('No match for argument: %s')
-                logger.info(msg, self.base.output.term.bold(pkgspec))
-                errors_spec.append(pkgspec)
-            for pkg in package_query:
-                self._di_install(pkg)
+        debuginfo_suffix_len = len(Package.DEBUGINFO_SUFFIX)
+        debugsource_suffix_len = len(Package.DEBUGSOURCE_SUFFIX)
 
-        if not self.dbgdone:
-            logger.info(_("No debuginfo packages available to install"))
+        for pkgspec in self.opts.package:
+            solution = dnf.subject.Subject(pkgspec).get_best_solution(self.base.sack,
+                                                                      with_src=False)
+
+            query = solution["query"]
+            if not query:
+                logger.info(_('No match for argument: %s'), self.base.output.term.bold(pkgspec))
+                errors_spec.append(pkgspec)
+                continue
+
+            package_dict = query.available()._name_dict()
+            # installed versions of packages have priority, replace / add them to the dict
+            package_dict.update(query.installed()._name_dict())
+
+            # Remove debuginfo packages if their base packages are in the query.
+            # They can get there through globs and they break the installation
+            # of debug packages with the same version as the installed base
+            # packages. If the base package of a debuginfo package is not in
+            # the query, the user specified a debug package on the command
+            # line. We don't want to ignore those, so we will install them.
+            # But, in this case the version will not be matched to the
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version themselves etc.
+            for name in list(package_dict.keys()):
+                if name.endswith(Package.DEBUGINFO_SUFFIX):
+                    if name[:-debuginfo_suffix_len] in package_dict:
+                        package_dict.pop(name)
+                if name.endswith(Package.DEBUGSOURCE_SUFFIX):
+                    if name[:-debugsource_suffix_len] in package_dict:
+                        package_dict.pop(name)
+
+            # attempt to install debuginfo and debugsource for the highest
+            # listed version of the package (in case the package is installed,
+            # only the installed version is listed)
+            for pkgs in package_dict.values():
+                first_pkg = pkgs[0]
+
+                # for packages from system (installed) there can be more
+                # packages with different architectures listed and we want to
+                # install debuginfo for all of them
+                if first_pkg._from_system:
+                    # we need to split them by architectures and install the
+                    # latest version for each architecture
+                    arch_dict = {}
+
+                    for pkg in pkgs:
+                        arch_dict.setdefault(pkg.arch, []).append(pkg)
+
+                    for package_arch_list in arch_dict.values():
+                        pkg = package_arch_list[0]
+
+                        if not self._install_debug_from_system(pkg.debug_name, pkg):
+                            if not self._install_debug_from_system(pkg.source_debug_name, pkg):
+                                self.installed_debuginfo_missing.add(str(pkg))
+
+                        if not self._install_debug_from_system(pkg.debugsource_name, pkg):
+                            self.installed_debugsource_missing.add(str(pkg))
+
+                    continue
+
+                # if the package in question is -debuginfo or -debugsource, install it directly
+                if first_pkg.name.endswith(Package.DEBUGINFO_SUFFIX) \
+                        or first_pkg.name.endswith(Package.DEBUGSOURCE_SUFFIX):
+
+                    self._install(pkgs)  # pass all pkgs to the solver, it will pick the best one
+                    continue
+
+                # if we have NEVRA parsed from the pkgspec, use it to install the package
+                if solution["nevra"] is not None:
+                    if not self._install_debug(first_pkg.debug_name, solution["nevra"]):
+                        if not self._install_debug(first_pkg.source_debug_name, solution["nevra"]):
+                            self.available_debuginfo_missing.add(
+                                "{}-{}".format(first_pkg.name, first_pkg.evr))
+
+                    if not self._install_debug(first_pkg.debugsource_name, solution["nevra"]):
+                        self.available_debugsource_missing.add(
+                            "{}-{}".format(first_pkg.name, first_pkg.evr))
+
+                    continue
+
+                # if we don't have NEVRA from the pkgspec, pass nevras from
+                # all packages that were found (while replacing the name with
+                # the -debuginfo and -debugsource variant) to the solver, which
+                # will pick the correct version and architecture
+                if not self._install_debug_no_nevra(first_pkg.debug_name, pkgs):
+                    if not self._install_debug_no_nevra(first_pkg.source_debug_name, pkgs):
+                        self.available_debuginfo_missing.add(
+                            "{}-{}".format(first_pkg.name, first_pkg.evr))
+
+                if not self._install_debug_no_nevra(first_pkg.debugsource_name, pkgs):
+                    self.available_debugsource_missing.add(
+                        "{}-{}".format(first_pkg.name, first_pkg.evr))
+
+        if self.available_debuginfo_missing:
+            logger.info(
+                _("Could not find debuginfo package for the following available packages: %s"),
+                ", ".join(sorted(self.available_debuginfo_missing)))
+
+        if self.available_debugsource_missing:
+            logger.info(
+                _("Could not find debugsource package for the following available packages: %s"),
+                ", ".join(sorted(self.available_debugsource_missing)))
+
+        if self.installed_debuginfo_missing:
+            logger.info(
+                _("Could not find debuginfo package for the following installed packages: %s"),
+                ", ".join(sorted(self.installed_debuginfo_missing)))
+
+        if self.installed_debugsource_missing:
+            logger.info(
+                _("Could not find debugsource package for the following installed packages: %s"),
+                ", ".join(sorted(self.installed_debugsource_missing)))
 
         if errors_spec and self.base.conf.strict:
             raise dnf.exceptions.PackagesNotAvailableError(_("Unable to find a match"),
                                                            pkg_spec=' '.join(errors_spec))
 
-    def _dbg_available(self, dbgname, package, match_evra):
-        if match_evra:
-            return self.packages_available.filter(
-                name=dbgname,
-                epoch=int(package.epoch),
-                version=str(package.version),
-                release=str(package.release),
-                arch=str(package.arch))
-        else:
-            return self.packages_available.filter(
-                name=dbgname,
-                arch=str(package.arch))
+    def _install_debug_from_system(self, debug_name, pkg):
+        query = self.base.sack.query().filter(name=debug_name,
+                                              epoch=pkg.epoch,
+                                              version=pkg.version,
+                                              release=pkg.release,
+                                              arch=pkg.arch)
 
-    def _di_install(self, package):
-        for dbgname in [package.debug_name, package.source_debug_name]:
-            if dbgname in self.dbgdone:
-                break
-            if self._dbg_available(dbgname, package, True):
-                di = "{0}-{1}:{2}-{3}.{4}".format(
-                    dbgname,
-                    package.epoch,
-                    package.version,
-                    package.release,
-                    package.arch)
-                self.base.install(di)
-            elif self._dbg_available(dbgname, package, False):
-                di = "{0}.{1}".format(dbgname, package.arch)
-                self.base.install(di)
-            else:
-                continue
-            self.dbgdone.append(dbgname)
-            break
-        else:
-            logger.info(
-                _("Could not find debuginfo for package: %s") % package)
+        if query:
+            self._install(query)
+            return True
+
+        return False
+
+    def _install_debug(self, debug_name, base_nevra):
+        kwargs = {}
+
+        # if some part of EVRA was specified in the argument, add it as a filter
+        if base_nevra.epoch is not None:
+            kwargs["epoch__glob"] = base_nevra.epoch
+        if base_nevra.version is not None:
+            kwargs["version__glob"] = base_nevra.version
+        if base_nevra.release is not None:
+            kwargs["release__glob"] = base_nevra.release
+        if base_nevra.arch is not None:
+            kwargs["arch__glob"] = base_nevra.arch
+
+        query = self.base.sack.query().filter(name=debug_name, **kwargs)
+
+        if query:
+            self._install(query)
+            return True
+
+        return False
+
+    def _install_debug_no_nevra(self, debug_name, pkgs):
+        query = self.base.sack.query().filterm(
+            nevra_strict=["{}-{}.{}".format(debug_name, p.evr, p.arch) for p in pkgs])
+        if query:
+            self._install(query)
+            return True
+
+        return False
+
+    def _install(self, pkgs):
+        selector = dnf.selector.Selector(self.base.sack)
+        selector.set(pkg=pkgs)
+        self.base.goal.install(select=selector, optional=not self.base.conf.strict)
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -1,4 +1,4 @@
-%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 4.2.8}
+%{?!dnf_lowest_compatible: %global dnf_lowest_compatible 4.2.10}
 %global dnf_plugins_extra 2.0.0
 %global hawkey_version 0.34.0
 %global yum_utils_subpackage_name dnf-utils


plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version himself etc.
+            for na in list(packages.keys()):
+                if na[0].endswith(Package.DEBUGINFO_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGINFO_SUFFIX)], na[1]) in packages:
+                        packages.pop(na)
+                if na[0].endswith(Package.DEBUGSOURCE_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGSOURCE_SUFFIX)], na[1]) in packages:
+                        packages.pop(na)
```
lukash: This is super-ugly. Help!

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            # They can get there through globs and they break the installation
+            # of debug packages with the same version as the installed base
+            # packages. If the base package of a debuginfo package is not in
+            # the query, the user specified a debug package on the command
+            # line. We don't want to ignore those, so we will install them.
+            # But, in this case the version will not be matched to the
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version himself etc.
+            for na in list(packages.keys()):
```
j-mracek: I would recommend to use a query instead with query = query.union(query2) or difference or so on functions.

lukash: How exactly would you use it in this case? I am doing a check for each package I'm iterating through. An alternative could be to make a list of packages that match the condition and filter those out - but the list can be potentially very long, and it will be a list of (name, architecture) pairs. Is this what you mean?

lukash:
I've added dependency on dnf-4.2.9 to the spec file.

plugins/debuginfo-install.py
pkratoch: So, the root cause why you are doing it is that ``debuginfo-install foo-debuginfo`` installs the newest version and does not match the version of ``foo``. And in the specific case of ``debuginfo-install foo foo-debuginfo`` (or similarly with globs), this fixes the problem. Right?

I get that you cannot just naively check the name without the suffix... but you are actually doing that here. So why not just check the existence of the base package, not just its presence in the packages? Would it be too slow? I really don't know, so sorry if you already thought about this.

I am thinking about doing one query for all the base names, installing these and then just checking that all the debuginfo names are in the packages_installed; and if not, installing them, too. Also, I don't know what @j-mracek had in mind, so maybe he has some good idea.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+        # if the package is installed and none of EVR was specified in the
+        # argument, use the installed package's EVR (install debuginfo with the
+        # same version as the installed package)
+        if package._from_system and \
+                "epoch" not in kwargs and \
+                "version" not in kwargs and \
+                "release" not in kwargs:
+            kwargs["epoch"] = package.epoch
+            kwargs["version"] = package.version
+            kwargs["release"] = package.release
```
pkratoch: When I have ``foo-1.0-1`` installed and ``foo-1.0-2`` available, then ``debuginfo-install foo-1.0`` installs ``foo-1.0-2``. Can it also match the installed package if the provided attributes are the same as of the installed package? That behavior seems better to me. For example like this:
```
        if package._from_system and \
                ("epoch" not in kwargs or kwargs["epoch"] == package.epoch) and \
                ("version" not in kwargs or kwargs["version"] == package.version) and \
                ("release" not in kwargs or kwargs["release"] == package.release):
            kwargs["epoch"] = package.epoch
            kwargs["version"] = package.version
            kwargs["release"] = package.release
```

plugins/debuginfo-install.py
lukash: > So, the root cause why you are doing it is that `debuginfo-install foo-debuginfo` installs the newest version and does not match the version of `foo`. And in the specific case of `debuginfo-install foo foo-debuginfo` (or similarly with globs), this fixes the problem. Right?

Yes.

> I get that you cannot just naively check the name without the suffix... but you are actually doing that here. So why not just check the existence of the base package, not just its presence in the packages? Would it be too slow? I really don't know, so sorry if you already thought about this.
> 
> I am thinking about doing one query for all the base names, installing these and then just checking that all the debuginfo names are in the packages_installed; and if not, installing them, too. Also, I don't know what @j-mracek had in mind, so maybe he has some good idea.

I'm not sure how this would be better? I still need to separate the packages to "base" and "debuginfo" by checking for the suffix, and then I'd have to do this in two steps, which I think would also result in not installing a newer version of e.g. `foo-debuginfo` with `dnf debuginfo-install foo-debuginfo` if an older version of `foo-debuginfo` was already installed? But maybe I'm just not getting it...

If you are thinking of performance overhead of using the NA (name, architecture) dicts, I think it's not a concern, I need to use the dicts anyway (to squash the installed and available package lists) and then the lookup is just O(1). The number of packages also shouldn't be huge.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            # of debug packages with the same version as the installed base
+            # packages. If the base package of a debuginfo package is not in
+            # the query, the user specified a debug package on the command
+            # line. We don't want to ignore those, so we will install them.
+            # But, in this case the version will not be matched to the
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version himself etc.
+            for na in list(packages.keys()):
+                if na[0].endswith(Package.DEBUGINFO_SUFFIX):
```
j-mracek: Providing a constant by Package.DEBUGINFO_SUFFIX is different way that was used so far for name of system repo or so on.

plugins/debuginfo-install.py
lukash: Good idea, I'll do it.

plugins/debuginfo-install.py
lukash: I'm not sure what you mean :confused:

j-mracek: What about to put DEBUGINFO_SUFFIX outside of package class

lukash: Why?

j-mracek: Debuginfo surfix is identical for package and for repo.

j-mracek: Anyway add local variable debuginfo_surfix = Package.DEBUGINFO_SUFFIX. I believe that it is a better to not call package object to get static value.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            # version themselves etc.
+            for na in list(packages.keys()):
+                if na[0].endswith(Package.DEBUGINFO_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGINFO_SUFFIX)], na[1]) in packages:
+                        packages.pop(na)
+                if na[0].endswith(Package.DEBUGSOURCE_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGSOURCE_SUFFIX)], na[1]) in packages:
+                        packages.pop(na)
+
+            if not packages:
```
j-mracek: The if could be moved several lines up and could end with "continue" line, because there is no packages.key() and no package.values(). Pop up statement still keep non-debuginfo packages in the dict.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            # But, in this case the version will not be matched to the
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version themselves etc.
+            for na in list(packages.keys()):
+                if na[0].endswith(Package.DEBUGINFO_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGINFO_SUFFIX)], na[1]) in packages:
+                        packages.pop(na)
+                if na[0].endswith(Package.DEBUGSOURCE_SUFFIX):
+                    if (na[0][:-len(Package.DEBUGSOURCE_SUFFIX)], na[1]) in packages:
```
j-mracek: The len of suffix could be calculated once.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            kwargs["version"] = base_nevra.version
+        if base_nevra.release is not None:
+            kwargs["release"] = base_nevra.release
+        if base_nevra.arch is not None:
+            kwargs["arch"] = base_nevra.arch
+
+        # if the package is installed and none of EVR was specified in the
+        # argument, use the installed package's EVR (install debuginfo with the
+        # same version as the installed package)
+        if package._from_system and \
```
j-mracek: I believe there is a problem with arch value in case that secondary architecture is installed.
foo.i686 installed
I believe that with argument foo, debuginfo for x86_64 will be installed. I am proposing to add:
```
if package._from_system:
    kwargs["arch"] = package.arch
```

j-mracek: Also globs in version, release, arch will make the function non-functional.
Keuwords could be replaced like "release__glob".
I am also not sure, how the if will handle it.
Probably conditions ```                 ("epoch" not in kwargs or kwargs["epoch"] == package.epoch) and \
                ("version" not in kwargs or kwargs["version"] == package.version) and \
                ("release" not in kwargs or kwargs["release"] == package.release): ```
could be removed, because installed package already matched version, release or so on in previous query.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+                ("version" not in kwargs or kwargs["version"] == package.version) and \
+                ("release" not in kwargs or kwargs["release"] == package.release):
+            kwargs["epoch"] = package.epoch
+            kwargs["version"] = package.version
+            kwargs["release"] = package.release
+
+        query = self.packages_available.filter(name=debug_name, **kwargs)
+
+        if query:
+            s = dnf.selector.Selector(self.base.sack)
```
j-mracek: Please could you rename s to sltr or something else? It will look better.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
                 errors_spec.append(pkgspec)
-            for pkg in package_query:
-                self._di_install(pkg)
 
-        if not self.dbgdone:
-            logger.info(_("No debuginfo packages available to install"))
+            # attempt to install debuginfo and debugsource for the highest
+            # listed version of the package (in case the package is installed,
+            # only the installed version is listed)
+            for pkg in packages.values():
```
j-mracek: I would recommend to add support when 'foo-debuginfo' argument is used. Debuginfo command not only search for packages, but also enebles debuginfo repositories. What about the change like this?
```
            for pkgs in packages.values():
                package = pkgs[0]
                pkg_name = package.name
                if pkg_name.endswith(Package.DEBUGINFO_SUFFIX) or pkg_name.endswith(Package.DEBUGSOURCE_SUFFIX):
                    sltr = dnf.selector.Selector(self.base.sack)
                    sltr.set(pkg=pkgs)
                    self.base.goal.install(select=sltr, optional=not self.base.conf.strict)
                else:
                    if not self._install(package, package.debug_name, solution["nevra"]):
                        self.debuginfo_missing.add(str(package))

                    if not self._install(package, package.debugsource_name, solution["nevra"]):
                        self.debugsource_missing.add(str(package))
```

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
-                self.base.install(di)
-            else:
-                continue
-            self.dbgdone.append(dbgname)
-            break
-        else:
-            logger.info(
-                _("Could not find debuginfo for package: %s") % package)
+    def _install(self, package, debug_name, base_nevra):
+        if debug_name in self.packages_installed:
```
j-mracek: Is this condition really necessary? When I use argument foo-1 and foo-2 and both version will exist, the code will install packages for the first argument. The second argument will be silently skipped. By removal of that condition it will fail with depsolver error.

plugins/debuginfo-install.py
lukash: Hmm, sorry, still not entirely clear.. so in stead of:

```python
class Package:
    DEBUGINFO_SUFFIX = "-debuginfo"
    # ...
```

You want me to put the constant straight into the module like:
```python
DEBUGINFO_SUFFIX = "-debuginfo"

class Package:
    # ...
```

At least that's what I think makes sense and I can do that... But it doesn't seem exactly what you're suggesting?

plugins/debuginfo-install.py
pkratoch: There is a support for ``foo-debuginfo`` arguments. Note that Lukas changed/added the ``debug_name`` and ``debugsource_name`` functions in https://github.com/rpm-software-management/dnf/pull/1425 so that it handles the suffixes better.

Debuginfo repositories are enabled in ``self.configure()``. Or what did you mean?

plugins/debuginfo-install.py
lukash: The complexity of len() in python is O(1), there is no calculation being done, but I wonder if the interpreter will optimize out the function calls or calls the function again and again multiple times. Brief research shows it does some optimizations of constants (which this is), but there is a function call so I'm not sure it can figure out it is a constant. Unless someone knows better if it does optimize, I'll store the lengths in variables.

plugins/debuginfo-install.py
j-mracek: Thanks Pavla I didn't know that. But anyway the proposed change will spare another query, because the search result should be identical.

plugins/debuginfo-install.py
lukash: > Also globs in version, release, arch will make the function non-functional.
> Keuwords could be replaced like "release__glob".
> I am also not sure, how the if will handle it.

Good observation. I've just tested it, it will not find the correct package when a glob is specified. However, what is a glob supposed to do here? The way it's supposed to work is the user specifies a version of the `-debuginfo` to install. If they specify a glob, which version should it install? Like the highest one matching the glob? I'm not sure we want to support this at all?

> Probably conditions ("epoch" not in kwargs or kwargs["epoch"] == package.epoch) and \ ("version" not in kwargs or kwargs["version"] == package.version) and \ ("release" not in kwargs or kwargs["release"] == package.release):
could be removed, because installed package already matched version, release or so on in previous query.

You're right, I think it can actually be removed and the tests do pass with it gone... :slightly_smiling_face:

plugins/debuginfo-install.py
lukash: This is one of the few remnants of the original code :slightly_smiling_face: It's clearly supposed to prevent reinstalling the same packages many times over. However, the `self.base.goal` already does that (and better), so I'll drop this.

plugins/debuginfo-install.py
lukash: For this three-line scope and clear definition of what `s` is on the right side of `=` I think one letter variables are fine, but I've renamed it to `selector`.

plugins/debuginfo-install.py
j-mracek: When glob is used in version i would suggest:
1. In set is installed package use version of installed package
2. Let solver to decide - mostly the highest version
You only need to change kwargs["version"] to kwargs["version__glob"], kwargs["release"] to kwargs["release__glob"], and kwargs["arch"] to kwargs["arch__glob"]. parsing epoch do not support globs.
The rest of the code will handle it correctly.

plugins/debuginfo-install.py
```diff
@@ -79,54 +83,95 @@ def run(self):
...
+            return True
+
+        kwargs = {}
+
+        # if some part of EVRA was specified in the argument, add it as a filter
+        if base_nevra.epoch is not None:
```
j-mracek: base_nevra could be None. Please try argument like "/etc/dnf/dnf.conf", or provide, or NEVRA when "-" or ":" will be replaced by "*"

plugins/debuginfo-install.py
j-mracek: Yes, that what I suggested. The question is in which module in libdnf.

plugins/debuginfo-install.py
j-mracek: Perfect

j-mracek:
Please reference also this bug - https://bugzilla.redhat.com/show_bug.cgi?id=1532378

plugins/debuginfo-install.py
lukash: I clearly need it for `package` here. IIUC you're saying the suffix is the same for repositories, but that is sort of a coincidence and it's also just a convention? We shouldn't try to link the two I think. I'd put it in `package`. I kind of liked it in the class, since that's the one working with the suffixes and it kinda feels it belongs there...

lukash:
The resolved comments were addressed in the latest commit.

The secondary architecture use-case and the non-existent nevra will require big changes to make them work though.

lukash:
I've pushed a newer version that deals with different architectures of installed packages. I'm now using `query._name_dict()` instead of `query._na_dict()`, as that didn't work for the multiple architecture case. But the `_name_dict()` still has a problem, because the package lists in its values can contain packages with different achitectures and sometimes the wrong architecture is first in the list, and it can:
1. be logged in case `-debuginfo` wasn't found, confusing the user
2. in case NEVRA is None (as @j-mracek notes above), the only information we have is this package, and it is wrong

I've written tests for these two cases here, they're marked with @xfail: https://github.com/rpm-software-management/ci-dnf-stack/pull/566

The problems are also marked with TODOs in the current code.

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,126 @@ def configure(self):
...
-            break
-        else:
-            logger.info(
-                _("Could not find debuginfo for package: %s") % package)
+    def _install_debug_from_system(self, package, debug_name):
+        query = self.packages_available.filter(name=debug_name,
+                                               epoch=package.epoch,
+                                               version=package.version,
+                                               release=package.release,
+                                               arch=package.arch)
```
j-mracek: Please replace self.packages_available by self.base.sack.query() to prevent reinstalling of packages

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,126 @@ def configure(self):
...
+        if base_nevra.epoch is not None:
+            kwargs["epoch__glob"] = base_nevra.epoch
+        if base_nevra.version is not None:
+            kwargs["version__glob"] = base_nevra.version
+        if base_nevra.release is not None:
+            kwargs["release__glob"] = base_nevra.release
+        if base_nevra.arch is not None:
+            kwargs["arch__glob"] = base_nevra.arch
+
+        query = self.packages_available.filter(name=debug_name, **kwargs)
```
j-mracek: Please could you replace self.packages_available by self.base.query to prevent reinstalling installed packages

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,125 @@ def configure(self):
...
+            # packages. If the base package of a debuginfo package is not in
+            # the query, the user specified a debug package on the command
+            # line. We don't want to ignore those, so we will install them.
+            # But, in this case the version will not be matched to the
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version themselves etc.
+            for name in list(package_dict.keys()):
+                if name.endswith(Package.DEBUGINFO_SUFFIX):
+                    if name[:-debuginfo_suffix_len] in package_dict:
```
j-mracek: Please add condition for testing arch. This could be very tricky because foo-debuginfo could be in dist with a single architecture and foo could have also a single or multiple arch packages in dict.

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,125 @@ def configure(self):
...
+            # listed version of the package (in case the package is installed,
+            # only the installed version is listed)
+            for packages in package_dict.values():
+                package = packages[0]
+
+                if packages[0]._from_system:
+                    # for packages from system (installed) there can be more
+                    # packages with different architectures listed and we want
+                    # to install debuginfo for all of them
+                    for package in packages:
```
j-mracek: I would recommend a different approach. It is little bit complicated therefore I provide a diff
```
diff --git a/plugins/debuginfo-install.py b/plugins/debuginfo-install.py
index 4b53565..28739d1 100644
--- a/plugins/debuginfo-install.py
+++ b/plugins/debuginfo-install.py
@@ -116,14 +116,18 @@ class DebuginfoInstallCommand(dnf.cli.Command):
             # listed version of the package (in case the package is installed,
             # only the installed version is listed)
             for packages in package_dict.values():
+                arch_dict = {}
+
                 package = packages[0]
 
-                if packages[0]._from_system:
+                if package._from_system:
                     # for packages from system (installed) there can be more
                     # packages with different architectures listed and we want
                     # to install debuginfo for all of them
-                    for package in packages:
-                        if not self._install_debug_from_system(package, package.debug_name):
+                    for pkg in packages:
+                        arch_dict.setdefault(pkg.arch, []).append(pkg)
+                    for pkgs_per_arch in arch_dict.values():
+                        if not self._install_debug_from_system(pkgs_per_arch, package.debug_name):
                             self.debuginfo_missing.add(str(package))
 
                         if not self._install_debug_from_system(package, package.debugsource_name):
@@ -131,7 +135,6 @@ class DebuginfoInstallCommand(dnf.cli.Command):
 
                 else:
                     # if the first package is not from the system, we just want its name
-                    package = packages[0]
                     if package.name.endswith(Package.DEBUGINFO_SUFFIX) \
                             or package.name.endswith(Package.DEBUGSOURCE_SUFFIX):
 
@@ -159,12 +162,14 @@ class DebuginfoInstallCommand(dnf.cli.Command):
             raise dnf.exceptions.PackagesNotAvailableError(_("Unable to find a match"),
                                                            pkg_spec=' '.join(errors_spec))
 
-    def _install_debug_from_system(self, package, debug_name):
-        query = self.base.sack.query().filter(name=debug_name,
-                                              epoch=package.epoch,
-                                              version=package.version,
-                                              release=package.release,
-                                              arch=package.arch)
+    def _install_debug_from_system(self, per_arch_packages, debug_name):
+        base_query = self.base.sack.query().filterm(arch=per_arch_packages[0].arch).apply()
+        query = self.base.sack.query().filterm(empty=True)
+        for package in per_arch_packages:
+            query.union(self.base.sack.query().filter(name=debug_name,
+                                                      epoch=package.epoch,
+                                                      version=package.version,
+                                                      release=package.release))
 
         if query:
             self._install(query)
(END)

```

j-mracek: Also part
```
        base_query = self.base.sack.query().filterm(arch=per_arch_packages[0].arch).apply()
        query = self.base.sack.query().filterm(empty=True)
        for package in per_arch_packages:
            query.union(base_query.filter(name=debug_name,
                                          epoch=package.epoch,
                                          version=package.version,
                                          release=package.release))
```
could be replaced by:
```
nevras=[]
        for pkg in per_arch_packages:
            nevras.append("{}-{}:{}-{}.{}".format(
                debug_name, pkg.epoch, pkg.version, pkg.release, pkg.arch))
        query = self.base.sack.query().filterm(nevra_strict=nevras)
```
The second option would be more efficient.

lukash: @j-mracek first, I'd really like to stress that I _need_ some sort of description of your proposed solutions. Most of the time I have to seriously try hard to figure out what you actually mean...

It seems you've got some sort of error in your patch, the `base_query` you create in `_install_debug_from_system` is never used. Even with that fixed though, I don't understand how your solution practically differs from what my code is doing...

And, on top of that, this doesn't address neither of the two problems I've described above :slightly_smiling_face::


> <snip> But the `_name_dict()` still has a problem, because the package lists in its values can contain packages with different achitectures and sometimes the wrong architecture is first in the list, and it can:
> 
> 1. be logged in case `-debuginfo` wasn't found, confusing the user
> 
> 2. in case NEVRA is None (as @j-mracek notes above), the only information we have is this package, and it is wrong

Please do see the tests in the CI repo, they demonstrate well what doesn't work. The two tests in question are marked with @xfail: [rpm-software-management/ci-dnf-stack#566](https://github.com/rpm-software-management/ci-dnf-stack/pull/566)

plugins/debuginfo-install.py
lukash: Not sure I understand you, but if you mean I should test if the `-debuginfo` package in the dict has the same architecture as the base package I'm looking up here, I don't think it's needed.

I don't think we really care about the architecture now, we are just detecting a case when both `-debuginfo` and base package matched a single input pkgspec. And if it did, we want to remove the `-debuginfo` package, becase it is superfluous here and breaks the correct handling of installing the debuginfo for the base package later, which will take care of the architectures too...

It's a bit complicated but I think there is no problem here, in case you still think it is, can you think of a test case I could add for it?

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,125 @@ def configure(self):
...
+                            or package.name.endswith(Package.DEBUGSOURCE_SUFFIX):
+
+                        self._install((package, ))
+                        continue
+
+                    if solution["nevra"] is None:
+                        print("TODO We have this package and it may be the wrong one: %s", package)
+
+                    if not self._install_debug(package.debug_name, solution["nevra"]):
+                        self.debuginfo_missing.add(str(package)) # TODO check this is the correct package being logged - could be different architecture etc.
```
j-mracek: To report latest packages for all arch it can be resolve by:
```
    def _latest_arch_representant(self, packages):
        """Packages should be sorted list"""
        latest_per_arch = [packages[0]]
        arch = packages[0].arch
        for pkg in packages[1:]:
            if arch != pkg.arch:
                latest_per_arch.append(pkg)
                arch = pkg.arch
        return latest_per_arch
```
and self.debuginfo_missing.add(str(package)) could be replaced by:
```
for pkg in _latest_arch_representant(packages):
    self.debuginfo_missing.add(str(pkg))
```

plugins/debuginfo-install.py
lukash: So the problem actually was with installonly packages. I've added a test to the tests PR and I've fixed the code to install only the latest version for each architecture, based on the above suggestion, but I've simplified it a bit, installing only the latest package for each architecture.

plugins/debuginfo-install.py
```diff
@@ -74,59 +75,133 @@ def configure(self):
...
+                    # if the first package is not from the system, we just want its name
+                    package = packages[0]
+                    if package.name.endswith(Package.DEBUGINFO_SUFFIX) \
+                            or package.name.endswith(Package.DEBUGSOURCE_SUFFIX):
+
+                        self._install((package, ))
+                        continue
+
+                    if solution["nevra"] is None:
+                        print("TODO We have this package and it may be the wrong one: %s", package)
```
j-mracek: What about to use a similar construct like for installed packages?
```
                        if not self._install_debug_no_nevra(packages, package.debug_name):
                            # same problem with logging and relevant arch
                            self.debuginfo_missing.add(str(package))

                        if not self._install_debug_no_nevra(packages, package.debugsource_name):
                            self.debugsource_missing.add(str(package))
```
```
    def _install_debug_no_nevra(self, packages, debug_name):
        """The function is identical with proposed change of _install_debug_from_system"""
        nevras = []
        for pkg in packages:
            nevras.append("{}-{}:{}-{}.{}".format(
                debug_name, pkg.epoch, pkg.version, pkg.release, pkg.arch))
        query = self.base.sack.query().filterm(nevra_strict=nevras)
        if query:
            self._install(query)
            return True

        return False
```

lukash: I've used this approach, just simplified a bit the `nevras` composition.

plugins/debuginfo-install.py
lukash: I've split the _Could not find ..._ messages instead and for installed packages I print out the full NEVRA, for available packages I print out NEVR, omitting the architecture, of which there can be multiple and it can be confusing listing the same package multiple times with different architectures.

lukash:
Pushed a new version which I hope is more or less final (apart from perhaps moving the `DEBUGINFO_SUFFIX` and `DEBUGSOURCE_SUFFIX` constants?). I've made some attempt to restructure the code a bit better as well.

@j-mracek please go through the unresolved conversations and see if you still find any upstanding issues (feel free to resolve the ones where you don't).

plugins/debuginfo-install.py
j-mracek: Please make change according to your best opinion

plugins/debuginfo-install.py
j-mracek: Probably you all right here.

plugins/debuginfo-install.py
j-mracek: ok

plugins/debuginfo-install.py
```diff
@@ -46,8 +45,7 @@ def config(self):
 
         if autoupdate:
             # allow update of already installed debuginfo packages
-            dbginfo = dnf.sack._rpmdb_sack(self.base).query().filter(
-                                                name__glob="*-debuginfo")
+            dbginfo = dnf.sack._rpmdb_sack(self.base).query().filter(name__glob="*-debuginfo")
```
j-mracek: Please replace `filter` by `filterm`. `filter` requrns new query, `filterm` modifies the present query.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+        debuginfo_suffix_len = len(Package.DEBUGINFO_SUFFIX)
+        debugsource_suffix_len = len(Package.DEBUGSOURCE_SUFFIX)
+
         for pkgspec in self.opts.package:
-            package_query = sorted(dnf.subject.Subject(pkgspec).get_best_query(
-                self.cli.base.sack).filter(arch__neq='src'), reverse=True)
-            if not package_query:
-                msg = _('No match for argument: %s')
-                logger.info(msg, self.base.output.term.bold(pkgspec))
+            solution = dnf.subject.Subject(pkgspec).get_best_solution(self.base.sack)
```
j-mracek: Please add one key word:
`solution = dnf.subject.Subject(pkgspec).get_best_solution(self.base.sack, with_src=False)`
In some cases when glob is used in combination with parser it ciuld result that it returns only src packages and line below will remove them. I don't remember any example.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+
+            query = solution["query"].filterm(arch__neq='src')
```
j-mracek: Can be removed, see above.

plugins/debuginfo-install.py
lukash: Why? Note I just changed the formatting here sort of ad-hoc... I'm not entirely sure what the code does, but it creates a `dbginfo` variable and stores the new query in there. So I don't see how using `filterm` would be correct here.

plugins/debuginfo-install.py
lukash: Will do.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+
+            package_dict = query.available()._name_dict()
+            # installed versions of packages have priority, replace / add them to the dict
+            package_dict.update(query.installed()._name_dict())
+
+            if not package_dict:
```
j-mracek: Please could you move the code 
```

            if not query:
                logger.info(_('No match for argument: %s'), self.base.output.term.bold(pkgspec))
                errors_spec.append(pkgspec)
                continue
```
before creation of dicts. It is more efficient and transparent.

lukash: Will do.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+                        if not self._install_debug_from_system(package.debugsource_name, package):
+                            self.installed_debugsource_missing.add(str(package))
+
+                    continue
+
+                # if the package in question is -debuginfo or -debugsource, install it directly
+                if first_package.name.endswith(Package.DEBUGINFO_SUFFIX) \
+                        or first_package.name.endswith(Package.DEBUGSOURCE_SUFFIX):
+
+                    self._install((first_package, ))
```
j-mracek: I would recommend to use all packages that we got like self._install(packages). To keep more options to solver is beneficial, because user could use arguments that have only limited overlap. Like `foo-debuginfo` and `foo-debuginfo-1-5`. Using all packages from the query is not harmful.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+
+        if query:
+            self._install(query)
+            return True
+
+        return False
+
+    def _install_debug_no_nevra(self, debug_name, packages):
+        query = self.base.sack.query().filterm(
+            nevra_strict=["{}-{}.{}".format(debug_name, p.evr, p.arch) for p in packages])
```
j-mracek: Please could you rename p to pkg. This is more common name in the project

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+                        self.available_debuginfo_missing.add(
+                            "{}-{}".format(first_package.name, first_package.evr))
+
+                    if not self._install_debug(first_package.debugsource_name, solution["nevra"]):
+                        self.available_debugsource_missing.add(
+                            "{}-{}".format(first_package.name, first_package.evr))
+
+                    continue
+
+                # inf we don't have NEVRA from the pkgspec, pass nevras from
```
j-mracek: Typo if not inf

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+
+                        if not self._install_debug_from_system(package.debug_name, package):
+                            self.installed_debuginfo_missing.add(str(package))
+
+                        if not self._install_debug_from_system(package.debugsource_name, package):
+                            self.installed_debugsource_missing.add(str(package))
+
+                    continue
+
+                # if the package in question is -debuginfo or -debugsource, install it directly
```
j-mracek: question? Did you mean query?

plugins/debuginfo-install.py
lukash: I would actually rather prefer a single-letter variable here. The scope is limited to this single line, so there is no doubt what it represents and the single letter makes the line shorter and easier to read. I think single letters are common in constructs like list comprehension and good to use there...

lukash: I could, though, rename:
`packages -> pkgs`
`package -> pkg`

Throughout the code. I does save quite a few characters overall and if it's sort of a convention...

plugins/debuginfo-install.py
m-blaha: You create new query by dnf.sack._rpmdb_sack(self.base).query(), then .filter() method creates and returns new query based on the original one. filterm() does the modification in-place. You would spare one query creation.

plugins/debuginfo-install.py
j-mracek: Proposed change `packages -> pkgs` `package -> pkg` is fine.

plugins/debuginfo-install.py
lukash: No, it's a phrase, meaning the package we are working with. I think it's fine to use it here...

plugins/debuginfo-install.py
lukash: Right, I didn't realize `filterm()` also returns the query. Thanks.

plugins/debuginfo-install.py
lukash: After some discussion with @m-blaha I'll keep the constants inside the class.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+            # installed version of the base package, as that would require
+            # another query and is further complicated if the user specifies a
+            # version themselves etc.
+            for name in list(package_dict.keys()):
+                if name.endswith(Package.DEBUGINFO_SUFFIX):
+                    if name[:-debuginfo_suffix_len] in package_dict:
+                        package_dict.pop(name)
+                if name.endswith(Package.DEBUGSOURCE_SUFFIX):
+                    if name[:-debugsource_suffix_len] in package_dict:
+                        package_dict.pop(name)
```
j-mracek: I believe that after the latest changes `for name in list(package_dict.keys()):` loop could be removed completely. If we will provide all possibility to solver (list of all packages and not only the latest) there is no need to reduce data set.

lukash: This won't work for the case when there is e.g. package `bar-2.0` installed and `bar-{debuginfo,debugsource}-2.0` don't exist. Expected behavior is no `bar-{debuginfo,debugsource}` aren't installed, but if I remove this piece of code, it will install the latest versions of `bar-{debuginfo,debugsource}` available.

There is a testcase with this scenario (called _"multiple packages, a glob that doesn't find debuginfo for the installed version"_) in the CI PR, which breaks if I try this (even if I pass all the `-debuginfo` packages to the solver instead of just the first one).

j-mracek: Then it must be resolved somewhere else.

plugins/debuginfo-install.py
```diff
@@ -74,59 +76,170 @@ def configure(self):
...
+            logger.info(
+                _("Could not find debuginfo package for the following installed packages: %s"),
+                ", ".join(sorted(self.installed_debuginfo_missing)))
+
+        if self.installed_debugsource_missing:
+            logger.info(
+                _("Could not find debugsource package for the following installed packages: %s"),
+                ", ".join(sorted(self.installed_debugsource_missing)))
 
         if errors_spec and self.base.conf.strict:
```
j-mracek: Why missing installed_debugsource_missing does not trigger an error? Then we can remove one loop above (`for name in list(package_dict.keys()):`).

lukash: I don't understand what you mean...

j-mracek: What about to replace it by:
`if (errors_spec or self.installed_debugsource_missing) and self.base.conf.strict:`

I suggest that this will resolve the issue when debuginfo is present but not in version of installed package and argument like (foo*) is used.

lukash: But how will it resolve the issue? This is just reporting an error at the end, how can it resolve the issue?

Also you've picked one of the four lists of missing packages, so I just don't see how that makes sense?

(I've pushed fixes for the minor comments you had)

plugins/debuginfo-install.py
j-mracek: > This won't work for the case when there is e.g. package `bar-2.0` installed and `bar-{debuginfo,debugsource}-2.0` don't exist. Expected behavior is no `bar-{debuginfo,debugsource}` are installed, but if I remove this piece of code, it will install the latest versions of `bar-{debuginfo,debugsource}` available.

In this case I believe that it should fail.

> 
> There is a testcase with this scenario (called _"multiple packages, a glob that doesn't find debuginfo for the installed version"_) in the CI PR, which breaks if I try this (even if I pass all the `-debuginfo` packages to the solver instead of just the first one).

j-mracek: Please provide detailed example including set packages installed, available and argument used for command.

lukash: The referenced test case is just that...

plugins/debuginfo-install.py
pkratoch: @j-mracek,  you have probably missed that based on the discussion in [bug 1586059](https://bugzilla.redhat.com/show_bug.cgi?id=1586059), there should be error only when the pkgspec doesn't match anything. If -debuginfo/-debugsource package is missing, it's not an error, because not all packages must have debuginfo. See [comment 15](https://bugzilla.redhat.com/show_bug.cgi?id=1586059#c15) for a reasoning.

lukash: (I somehow thought the discussion is still about the `for name in list(package_dict.keys()):` "hack" way earlier in the code, also discussed in the thread above. Thanks @pkratoch)

lukash:
I've found out that the feature of trying to install `-debuginfo` package from the base (parent) package in case it is not found for a subpackage (e.g. `debuinfo install foo-subpackage` will try to install `foo-debuginfo` if `foo-subpackage-debuginfo` is not found), which I've removed originally, is required, because e.g. `glibc` does have the debuginfo set up this way.

Luckily it wasn't too hard to add, although it does add one more twist to the plot...

I've pushed a fixup commit for now, so that it's clear what I've changed, I'll squash it later.

plugins/debuginfo-install.py
j-mracek: @lukash Please could you consider the proposed change?

lukash: Done.

lukash:
I think all issues are resolved and this is ready to merge (along with the tests).

j-mracek:
@rh-atomic-bot r+

j-mracek:
@rh-atomic-bot retry

lukash:
@rh-atomic-bot retry

j-mracek:
@rh-atomic-bot retry

pkratoch:
The tests passed locally, merging.

PR #345, opened 2019-07-01
Remove unneeded function from mock

j-mracek:
One of the functions emulate function that was even removed from dnf.


m-blaha:
@rh-atomic-bot r+

PR #346, opened 2019-07-12
[reposync] Add --urls option (RhBug:1686602)

m-blaha:
This is compatibility with yum-utils version of the reposync.
When the --urls switch is used, packages are not downloaded from
repositories, only their URLs are printed.

https://bugzilla.redhat.com/show_bug.cgi?id=1686602


m-blaha:
Corresponding test: https://github.com/rpm-software-management/ci-dnf-stack/pull/575

jamescassell:
Would be great if this also printed the metadata URLs when `--download-metadata` is passed, rather than the behavior of printing the URLs for packages, but downloading the metadata.

jamescassell:
Also, this seems to output `rsync://` URLs on Fedora default repos, which is likely not what `dnf` would use to download the objects.

m-blaha:
@jamescassell I added commit to not use rsync:// urls in --urls output.

m-blaha:
Requires https://github.com/rpm-software-management/dnf/pull/1455 and https://github.com/rpm-software-management/libdnf/pull/770

plugins/reposync.py
```diff
@@ -181,25 +199,22 @@ def get_pkglist(self, repo):
             query.filterm(arch=self.opts.arches)
         return query
 
-    def download_packages(self, repo):
+    def download_packages(self, repo, pkglist):
```
pkratoch: The ``repo`` argument is no longer used; it could be removed, too.

m-blaha: Thanks, removed.

plugins/reposync.py
```diff
@@ -109,10 +112,40 @@ def run(self):
...
+                                       if md_type in mdl]
+                    if group_locations:
+                        for group_location in group_locations:
+                            url = repo.remote_location(group_location)
+                            if url:
+                                print(url)
+                                break
+                        else:
+                            msg = _("Failed to get mirror for the group file.")
+                            logger.warning(msg)
```
pkratoch: We discussed this offline, but I want to mention it here as well to keep better track of it - this code actually works with different comps locations than the getcomps method.

m-blaha: Yes, the getcomps method uses cached metadata - I can add demand `freshest_metadata` to make sure that metadata in cache are really up-to-date regardless the expiration status of the repository. It is not useful to download the whole metadata set again and we do not have method to download only comps metadata - we have only possibility to download default metadata set (primary, comps and filelists), full metadata (used by reposync --download-metadata) or  we can add some extra metadata to the default set (for example "other" metadata containing the changelogs).

But when printing the URL of the comps metadata, the cached version is not useful -  we need to find out original url from where the file was downloaded to the cache.

PR #347, opened 2019-07-17
Behaviour of "--setopt" in config-manager plugin (RhBug:1702678)

jrohel:
The PR changes behaviour of `--setopt` in config-manager plugin.

`--setopt:` Fix crash with `--save --dump` 
Adds globs support to `--setopt repoid` (`--setopt=*-debuginfo.key=value`).
` --setopt=key=value` is not applied to repositories config
`--setopt` and empty list of repositories (RhBug:1702678)
`--setopt`: Add check for existence of input repositories

Solves: https://bugzilla.redhat.com/show_bug.cgi?id=1702678
It is more compatible with YUM now.

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -90,8 +90,6 @@ class ConfigManagerCommand(dnf.cli.Command):
                 self.base.conf.write_raw_configfile(dnf.const.CONF_FILENAME, 'main', sbc.substitutions, modify)
             if self.opts.dump:
                 print(self.base.output.fmtSection('main'))
-                for name, val in modify.items():
-                    sbc._set_value(name, val)
                 print(self.base.conf.dump())
 
         if self.opts.set_enabled or self.opts.set_disabled:
@@ -120,9 +118,6 @@ class ConfigManagerCommand(dnf.cli.Command):
                 self.base.conf.write_raw_configfile(repo.repofile, repo.id, sbc.substitutions, repo_modify)
             if self.opts.dump:
                 print(self.base.output.fmtSection('repo: ' + repo.id))
-                for name, val in repo_modify.items():
-                    if repo._has_option(name):
-                        repo._set_value(name, val)
                 print(repo.dump())
 
     def add_repo(self):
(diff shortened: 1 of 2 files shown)


plugins/config_manager.py
```diff
@@ -78,6 +78,31 @@ def run(self):
...
+        not_matching_repos_id = set()  # IDs of not matched repositories
+
+        def match_repos(key, add_matching_repos):
+            matching = self.base.repos.get_matching(key)
+            if not matching:
+                not_matching_repos_id.add(name)
+            elif add_matching_repos:
+                matching_repos.extend(matching)
+
+        if self.opts.crepo:
```
dmnks: I'm thinking that if we extended `matching_repos` with repoids from `repo_setopts` unconditionally (that is, regardless of whether `self.opts.crepo` is empty or not), we could

* simplify the code quite a bit (the `else` clause could be removed and `match_repos()` could be simplified if not removed as well) and
* make the behavior compatible with the original `yum-config-manager` in RHEL-7.

The difference currently is in the way we treat these repoids: here, we use them to **initialize the default** argument list, whereas in RHEL-7, we use them simply to **extend** the final argument list. The rationale behind the latter is that once you use a repoid-flavoured `--setopt`, you expect that to be applied, no matter what other repos you list as arguments.

In the nutshell, `--setopt=c.gpgcheck=0 a b` would currently not alter `c`. This limitation is not documented in the man page (as updated in this PR). As an alternative to what I proposed above, we could just add a note note to the `--setopt` section about this limitation, but I'm afraid it would just make it all more confusing to the reader.

dmnks:
The behavior is pretty consistent now and should be easier for the users to understand.

With this update, the only differences from the original `yum-config-manager` are the following:

* `--setopt=c.gpgcheck=0 a b` won't touch the repo `c`. This is, however, consistent with the overall concept of the explicit section list that can be used to override the default section list (the man page, also updated in this PR, describes that well).
* The default section list is initialized using only enabled repos in the original `yum-config-manager` implementation. Here, we select all the repos, even disabled ones. This shouldn't be a big deal, though, and can be fixed later if needed.

Everything else LGTM. Merging. Thanks!

PR #348, opened 2019-07-30
[spec] Add conflict with yum-plugin-copr

j-mracek:
It is required because both provide a same file.

file /usr/share/man/man8/yum-copr.8.gz from install
of yum-plugin-copr-1.1.31-519.fc30.noarch conflicts with file
from package dnf-utils-4.0.7-1.fc30.noarch

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -166,7 +166,8 @@ Additionally provides generate_completion_cache passive plugin.
 %if %{with yumutils}
 %package -n %{yum_utils_subpackage_name}
 %if "%{yum_utils_subpackage_name}" == "dnf-utils"
-Conflicts:      yum-utils < 1.1.31-513
+Conflicts:      yum-utils < 1.1.31-520
+Conflicts:      yum-plugin-copr < 1.1.31-520
 %if 0%{?rhel} != 7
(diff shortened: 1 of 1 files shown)


dnf-plugins-core.spec
```diff
@@ -167,6 +167,7 @@ Additionally provides generate_completion_cache passive plugin.
 %package -n %{yum_utils_subpackage_name}
 %if "%{yum_utils_subpackage_name}" == "dnf-utils"
 Conflicts:      yum-utils < 1.1.31-513
+Conflicts:      yum-plugin-copr < 1.1.32
```
Conan-Kudo: Change this to `Conflicts:      yum-plugin-copr < 1.1.31-520`

j-mracek: Any minor release of that plugin will result that I have to create new patch. We have to always think about future and not only about present situation. I know that there will be not any yum-utils < 1.1.32. I also know that dnf-utils will conflict with yum-utils. But for this time I am going to change it.

Conan-Kudo:
@j-mracek Just either rename or delete the man page in the yum-utils package in Fedora. It's already retired for F31.

Conan-Kudo:
@rh-atomic-bot r+

PR #350, opened 2019-08-08
copr: specify arch when requesting repo

praiskup:
The current copr frontend code simply ignores additional path
specifiers, so the overall output isn't affected.

We'd like though, if it will be enabled in the corresponding
copr project, diverge slightly the repo files for some
architectures.  Namely to support multilib:
https://pagure.io/copr/copr/issue/1

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -287,6 +287,10 @@ Do you really want to enable {0}?""".format('/'.join([self.copr_hostname,
         elif not match_any:
             return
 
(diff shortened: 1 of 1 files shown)


xsuchy:
This is on hold until the Copr server part will be merged and released.

praiskup:
> This is on hold until the Copr server part will be merged and released.

Merge is enough (so we all agreed that the new format of URL is correct).  Otherwise the code in this PR is compatible with old copr frontend URL (old frontend ignores the `&arch=x86_64` argument) so this can be easily released.

PR #351, opened 2019-08-11
[spec] Generate yum-utils(1) instead of symlinking (RhBug:1676418)

dmnks:
This ensures that the man page actually says "yum-utils" instead of
"dnf-utils" in the beginning.


pkratoch:
LGTM. Tried on RHEL 8, Fedora 30 and rawhide.

PR #355, opened 2019-09-05
[spec] Fix man page conflicts with old YUM plugins

dmnks:


dmnks:
Not needed (already fixed in a better way: https://github.com/rpm-software-management/dnf-plugins-core/pull/333)

PR #358, opened 2019-09-09
[doc] Skip creating and installing migrate documentation for Python 3+

Ionic:
The migrate plugin is Python-2-only, and so should be its documentation.

Note: this will not remove references in other files (really just index.rst). Doing so would be non-trivial and is probably not worth the effort. A broken link should be red flag enough for users not to try to use a plugin. Certainly better than shipping documentation for something that won't be available anyway IMHO.


Conan-Kudo:
@Ionic Could you please rebase this PR, then I can go ahead and get it merged.

Ionic:
Should be good to go again.

Conan-Kudo:
@rh-atomic-bot r+

PR #359, opened 2019-09-10
copr: fix disable if copr instance has non-default port

praiskup:
With config like
$ cat /etc/dnf/plugins/copr.d/test.conf
[local]
hostname = localhost
protocol = http
port = 55555

The 'dnf copr disable ...' failed before because `self.copr_hostname`
looks like 'something:55555'.


praiskup:
@FrostyX @xsuchy ptal

FrostyX:
+1

PR #360, opened 2019-09-11
[doc] Add notes that general DNF options are also available (RhBug:16…

pkratoch:
…89645)

The plugin commands also accept all general DNF options and it should
be mentioned in each plugin's documentation.

I also removed the "Besides these reposync specific options" bit for
reposync in order to unify the message format with other plugins (and
not all plugin commands have options specific to them).

https://bugzilla.redhat.com/show_bug.cgi?id=1689645


doc/copr.rst
```diff
@@ -96,6 +102,12 @@ Arguments (playground)
 ``upgrade``
     Upgrade the Playground repository settings (same as ``disable`` and then ``enable``).
 
+--------------------
+Options (playground)
+--------------------
+
+The ``dnf playground`` command also accepts all general DNF options. See `Options` in :manpage:`dnf(8)` for details.
```
lukash: Saying _also_ when there are no other options seems a bit weird, I'd leave it out if the _Options_ section is otherwise empty.

FWIW if it was me, I'd put the same somewhat shorter message at the beginning of the _Options_ section:
> All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.

Which eliminates the slight differences and makes it the first thing the user reads. But it's bikeshedding a bit :slightly_smiling_face: (feel free to leave it this way)

pkratoch: > Saying _also_ when there are no other options seems a bit weird, I'd leave it out if the _Options_ section is otherwise empty.

I can leave out the "also".

> Which eliminates the slight differences and makes it the first thing the user reads.

But we don't want it to be the first thing the user reads, do we? I think in most cases, the general options don't make much sense with the plugin command, whereas the plugins options are more likely to be relevant.

lukash: > But we don't want it to be the first thing the user reads, do we? I think in most cases, the general options don't make much sense with the plugin command, whereas the plugins options are more likely to be relevant.

True, but it's a very short message. And while the general options themselves most likely aren't very relevant, the fact they are applicable is, IMO. So I'd say get this important fact across while we have the user's attention and then get on with the options. If the user is looking for a plugin-specific option, this single line won't be much of a distraction.

Anyway, I'm not sure, I'll leave it up to you :slightly_smiling_face:

pkratoch: > True, but it's a very short message. And while the general options themselves most likely aren't very relevant, the fact they are applicable is, IMO. So I'd say get this important fact across while we have the user's attention and then get on with the options. If the user is looking for a plugin-specific option, this single line won't be much of a distraction.

Fair point and I am inclined to change it. How about the reposync message, though? Have the note about the ``--repoid`` option at the beginning, too? Split it up? I don't like neither of these variants. Any other idea?

m-blaha: Just FYI, mention of --repoid came from fixing the https://bugzilla.redhat.com/show_bug.cgi?id=1659328 bug. There are also examples in reposync doc and some of those use the --repoid option. May this is enough and --repoid part could be left out from the message.

PR #363, opened 2019-10-03
[versionlock] Add --raw option (RhBug:1645564)

m-blaha:
Versionlock add/exclude commands are now able to insert into
versionlock.list file patterns without converting it into NEVRAs. It is
usable for example for locking package to versions not available yet:

$ dnf version lock --raw add 'bash-5.*'
Adding versionlock on: bash-5.*

This command will keep the bash package on major version 5 and dnf will
still consume any future updates as far as they do not bump the major
version.

Without --raw switch used, the versionlock would take only currently
available packages into account:

$ dnf versionlock add 'bash-5.*'
Adding versionlock on: bash-0:5.0.7-1.fc30.*

https://bugzilla.redhat.com/show_bug.cgi?id=1645564

### DIFF
--- a/doc/versionlock.rst
+++ b/doc/versionlock.rst
@@ -117,6 +117,13 @@ Options
 
 All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.
 
+``--raw``
+    Do not resolve ``<package-name-spec>`` to NEVRAs to find specific version to lock to. Instead ``<package-name-spec>`` are used as they are. This enables locking to not yet available versions of the package.
+    For example you may want to keep the `bash` package on major version 5 and consume any future updates as far as they keep the major version::
+
+        $ dnf versionlock add --raw 'bash-5.*'
+        Adding versionlock on: bash-5.*
+
 -------------
 Configuration
 -------------
--- a/doc/versionlock.rst
+++ b/doc/versionlock.rst
@@ -51,31 +51,61 @@ operations like `repoquery`, `list`, `info`, etc.
 Synopsis
 --------
 
-``dnf versionlock [add|exclude|list|delete|clear] [<package-spec>]``
+``dnf versionlock [options] [add|exclude|list|delete|clear] [<package-name-spec>]``
 
 ---------
 Arguments
 ---------
 
-``<package-spec>``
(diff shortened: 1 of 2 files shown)


doc/versionlock.rst
```diff
@@ -117,6 +117,13 @@ Options
 
 All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.
 
+``--raw``
+    Do not parse ``<package-name-spec>`` to NEVRAs to find specific version to lock to. Instead ``<package-name-spec>`` are used as they are. This enables locking to versions of package not yet available.
```
lukash: This reads a bit better IMO:
> ... This enables locking to not yet available versions of the package.

doc/versionlock.rst
```diff
@@ -117,6 +117,13 @@ Options
...
+    For example you want to keep the `bash` package on major version 5 and consume any future updates as far as they keep the major version::
```
lukash: > For example you **may** want to ...

m-blaha:
Thanks @lukash, fixed.

lukash:
@m-blaha I think we could use tests in `ci-dnf-stack` for the `--raw` switch. `versionlock` tests are there already, so adding this should be easy.

m-blaha:
Yes, I'll add the tests.

lukash:
Waiting for tests...

m-blaha:
Here is PR with the tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/681

doc/versionlock.rst
```diff
@@ -117,6 +117,13 @@ Options
 
 All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.
 
+``--raw``
+    Do not parse ``<package-name-spec>`` to NEVRAs to find specific version to lock to. Instead ``<package-name-spec>`` are used as they are. This enables locking to not yet available versions of the package.
```
lukash: Okay, one more minor thing, can you please replace _"parse"_ with _"resolve"_ here?
> Do not ** resolve** ``<package-name-spec>`` to NEVRAs ...

I think it better describes what's going to happen.

PR #366, opened 2019-10-24
Add new plugin post-transaction-actions (RhBug:967264)

jrohel:


plugins/post-transaction-actions.py
```diff
@@ -0,0 +1,158 @@
...
+                       dnf.transaction.PKG_REINSTALL: "reinstall",
+                       dnf.transaction.PKG_REINSTALLED: "reinstalled",
+                       dnf.transaction.PKG_REMOVE: "remove",
+                       dnf.transaction.PKG_UPGRADE: "upgrade",
+                       dnf.transaction.PKG_UPGRADED: "upgraded"}
+
+        try:
+            action = action_dict[ts_item.action]
+        except KeyError:
+            action = "unknown - %s" % ts_item.action
```
m-blaha: This could be shortened using .get() method and default:
```
action = action_dict.get(ts_item.action, "unknown - %s" % ts_item.action)
```

plugins/post-transaction-actions.py
```diff
@@ -0,0 +1,158 @@
...
+            all_ts_items.append(ts_item)
+
+        commands_to_run = set()
+        for (a_key, a_state, a_command) in action_tuples:
+            if a_state == "forward":
+                ts_items = forward_ts_items
+            elif a_state == "backward":
+                ts_items = backward_ts_items
+            elif a_state == "any":
+                ts_items = all_ts_items
```
m-blaha: You do not need all_ts_items list at all, you can use `ts_items = forward_ts_items + backward_ts_items` (in case the order of items is not important)

plugins/post-transaction-actions.py
```diff
@@ -0,0 +1,158 @@
...
+                                        universal_newlines=True)
+                stdout, stderr = proc.communicate()
+                if stdout:
+                    self.logger.info(_('post-transaction-actions: %s') % (stdout))
+                if stderr:
+                    self.logger.error(_('post-transaction-actions: %s') % (stderr))
+            except Exception as e:
+                self.logger.error(_('post-transaction-actions: Bad Command "%s": %s') %
+                                  (command, e))
+            proc = None
```
m-blaha: Why?

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The ``enabled`` is used to enable/disable plugin. If it is omitted default value "1" is used.
+The ``actiondir`` sets path to directory with action files. Action file must have extension ".action".
+If ``actiondir`` is omitted default value "/etc/dnf/plugins/post-transaction-actions.d/" is used.
+
+------------------
+Action file format
+------------------
+
+Empty lines and lines starts with '#' character are ignored.
```
lukash: > Empty lines and lines **that start** with **a** '#' are ignored.

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,90 @@
...
+Each non-comment line in these files contains the following three fields: action_key, transaction_state,
+command. The fields are separated by ':' character.
```
lukash: > Each non-comment line defines an action and consists of three items separated by colons: ``action_key:transaction_state:command``.

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,90 @@
...
+
+line format:
+  ``action_key:transaction_state:command``
```
lukash: I'd leave the line format (the two above lines) out, I've merged that with the previous paragraph ...

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,90 @@
...
+
+The following variables are allowed to be passed to any ``command``:
+ * ``${name}``, ``$name`` - package name
+ * ``${arch}``, ``$arch`` - package arch
+ * ``${ver}``, ``$ver`` - package version
+ * ``${rel}``, ``$rel`` - package release
+ * ``${epoch}``, ``$epoch`` - package epoch
+ * ``${repoid}``, ``$repoid`` - package repository id
+ * ``${state}``, ``$state`` - "downgrade", "downgraded", "install", "obsolete", "obsoleted",
+                              "reinstall", "reinstalled", "remove", "upgrade", "upgraded"
```
lukash: ... and I'd do this in similar fashion as the config option listing, like:
```
``action_key``
    <here I think it needs better description of what it is actually matched against - are those NEVRAs, can they be in the shortened forms. The paths to files are for file provides? Does it match against regular package provides as well?>

    <I'm also debating whether action_key is a good name - seems to be more like a package filter...

``transaction_state``
    ...

``command``
    ...
```

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,90 @@
...
+
+An example action file:
+
+.. code-block:: none
+
+   # log all packages (state, nevra, repo) in transaction into.
```
lukash: Seems this sentence ends abruptly, should be like:
> ... in transaction into **a file**.

plugins/post-transaction-actions.py
jrohel: Yes. It looks better. I will change it.

plugins/post-transaction-actions.py
jrohel: Using all_ts_items is optimization. The `all_ts_items` is prepared once. Instead of making sum of lists in each loop iteration.

plugins/post-transaction-actions.py
jrohel: It is not needed. I will remove it.

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+   The following variables are supported:
+      * ``${name}``, ``$name`` - package name
+      * ``${arch}``, ``$arch`` - package arch
+      * ``${ver}``, ``$ver`` - package version
+      * ``${rel}``, ``$rel`` - package release
+      * ``${epoch}``, ``$epoch`` - package epoch
+      * ``${repoid}``, ``$repoid`` - package repository id
+      * ``${state}``, ``$state`` - "downgrade", "downgraded", "install", "obsolete", "obsoleted",
+                                   "reinstall", "reinstalled", "remove", "upgrade", "upgraded"
```
lukash: So for this part, I'd first do:
> Any shell command. The following variables in the command will be substituted:
>    (list of variables)

And then in a next paragraph, I'd describe the actual behavior a bit more thoroughly, my take:
> The shell command will be evaluated for each package that matched the `package_filter`. However, after variable substitution, any duplicate commands will be removed and each command will only be executed once per transaction. The order of execution of the commands may differ from the order of packages in the transaction.

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,89 @@
...
+   The order of execution of the commands may differ from the order of the packages in the transaction.
+
+   The following variables are supported:
+      * ``${name}``, ``$name`` - package name
+      * ``${arch}``, ``$arch`` - package arch
+      * ``${ver}``, ``$ver`` - package version
+      * ``${rel}``, ``$rel`` - package release
+      * ``${epoch}``, ``$epoch`` - package epoch
+      * ``${repoid}``, ``$repoid`` - package repository id
+      * ``${state}``, ``$state`` - "downgrade", "downgraded", "install", "obsolete", "obsoleted",
```
lukash: I'd add a short description before the list of values, like:
> ``${state}``, ``$state`` - the change of package state in the transaction: "downgrade", "downgraded", ...

etc/dnf/plugins/post-transaction-actions.conf
```diff
@@ -0,0 +1,3 @@
+[main]
+enabled = 1
+actiondir = /etc/dnf/plugins/post-transaction-actions.d/
```
lukash: A question, why are we shipping a config file with values that are the default anyway?

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,89 @@
...
+``package_filter``
+   A (glob-like) filtering rule aplied on the package NEVRA (also in the shortened forms) or
+   package files.
+
+``transaction_state``
+   Filters packages according to their state in the transaction.
+
+   * ``forward`` - packages that appeared on the system (includes reinstall packages)
+   * ``backward`` - packages that got removed from the system
+   * ``any`` - all packages
```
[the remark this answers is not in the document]
m-blaha: The original yum plugin used states: `update, install, remove, any`. But `update` state was only a subset of `install` so I understand why @jrohel wants to make it unambiguous.
Although we are using forward/backward terminology internally a lot, I have to admit I do not like it and consider it obscure.

etc/dnf/plugins/post-transaction-actions.conf
m-blaha: Consistency? Explicit is better then implicit? All other plugins do it this way.

lukash: Hmm.. okay :slightly_smiling_face:

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,89 @@
...
+    Path to the directory with action files. Action files must have the ".action" extension.
+    Default value is "/etc/dnf/plugins/post-transaction-actions.d/".
+
+------------------
+Action file format
+------------------
+
+Empty lines and lines that start with a '#' character are ignored.
+Each non-comment line defines an action and consists of three items separated by colons:
+package_filter, transaction_state, command.
```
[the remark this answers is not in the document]
jrohel: Yes. I will change it.

doc/post-transaction-actions.rst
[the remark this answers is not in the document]
jrohel: OK. We discussed about that in team. I will replace the word "forward" with "incoming" and "backward" with "outgoing".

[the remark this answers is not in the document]
jrohel: So, another rename. I hope the last one. To "in" and "out".

doc/post-transaction-actions.rst
```diff
@@ -0,0 +1,91 @@
...
+    Path to the directory with action files. Action files must have the ".action" extension.
+    Default value is "/etc/dnf/plugins/post-transaction-actions.d/".
+
+------------------
+Action file format
+------------------
+
+Empty lines and lines that start with a '#' character are ignored.
+Each non-comment line defines an action and consists of three items separated by colons:
+package_filter:transaction_state:command.
```
lukash: I'd put the line format in backticks, unless there's a reason not to?
```
``package_filter:transaction_state:command``.
```

m-blaha:
@rh-atomic-bot r+

dnf-plugins-core.spec
```diff
@@ -485,6 +513,7 @@ PYTHONPATH=./plugins nosetests-%{python3_version} -s tests/
 %{python2_sitelib}/dnf-plugins/download.*
 %{python2_sitelib}/dnf-plugins/generate_completion_cache.*
 %{python2_sitelib}/dnf-plugins/needs_restarting.*
+%{python2_sitelib}/dnf-plugins/post-transaction-actions.*
```
m-blaha: This line should not be here. It would make post-transaction-actions plugin part of dnf-plugins-core package. We want it as separate subpackage.

dnf-plugins-core.spec
```diff
@@ -510,6 +539,7 @@ PYTHONPATH=./plugins nosetests-%{python3_version} -s tests/
 %{python3_sitelib}/dnf-plugins/download.py
 %{python3_sitelib}/dnf-plugins/generate_completion_cache.py
 %{python3_sitelib}/dnf-plugins/needs_restarting.py
+%{python3_sitelib}/dnf-plugins/post-transaction-actions.py
```
m-blaha: This line should not be here.

dnf-plugins-core.spec
```diff
@@ -524,6 +554,7 @@ PYTHONPATH=./plugins nosetests-%{python3_version} -s tests/
 %{python3_sitelib}/dnf-plugins/__pycache__/download.*
 %{python3_sitelib}/dnf-plugins/__pycache__/generate_completion_cache.*
 %{python3_sitelib}/dnf-plugins/__pycache__/needs_restarting.*
+%{python3_sitelib}/dnf-plugins/__pycache__/post-transaction-actions.*
```
m-blaha: This line should not be here.

PR #367, opened 2019-10-25
Adjust output yum/dnf according to used binary (RhBug:1746474)

j-mracek:
https://bugzilla.redhat.com/show_bug.cgi?id=1746474

Requires: https://github.com/rpm-software-management/dnf/pull/1514
Requires CI - rpm-software-management/ci-dnf-stack#668

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -23,6 +23,7 @@ from dnfpluginscore import _, logger, P_
 import dnf
 import dnf.cli
 import dnf.pycomp
+import dnf.util
 import fnmatch
 import os
 import re
@@ -33,7 +34,8 @@ import shutil
 class ConfigManagerCommand(dnf.cli.Command):
 
     aliases = ['config-manager']
-    summary = _('manage dnf configuration options and repositories')
+    summary = _('manage {prog} configuration options and repositories').format(
+        prog=dnf.util.MAIN_PROG)
 
     @staticmethod
     def set_argparser(parser):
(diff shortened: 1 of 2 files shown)


plugins/config_manager.py
```diff
@@ -167,7 +169,7 @@ def add_repo(self):
             else:
                 # just url to repo, create .repo file on our own
                 repoid = sanitize_url_to_fs(url)
-                reponame = 'created by dnf config-manager from %s' % url
+                reponame = 'created by {0} config-manager from {1}'.format(dnf.util.MAIN_PROG, url)
```
lukash: Out of curiosity, why use named placeholder on the previous string and numbered placeholders on this one?

Also, this string isn't translated.. do we want to translate it perhaps?

j-mracek: I believe that we don't want to translated. This information is stored in .repo files and names should be same generated by same command. Is something wrong in substitution that I used? Because the string is not translated the substitution format is not so important.

lukash: I see, you're right, we don't want this translated I think.

And nothing wrong about it, just looking at this commit the formatting placeholders are inconsistent, but I don't really mind.

j-mracek: Ok I changed it to `'created by {} config-manager from {}'.format(dnf.util.MAIN_PROG, url)`

m-blaha:
@rh-atomic-bot r+

PR #368, opened 2019-11-08
[reposync] Apply only modular excludes (RhBug:1750273)

m-blaha:
This is a regression, reposync always used to apply excludes and
includes. It only needs to ignore modular dependencies to be able to
synchronize non-enabled streams.

https://bugzilla.redhat.com/show_bug.cgi?id=1750273

### DIFF
--- a/plugins/reposync.py
+++ b/plugins/reposync.py
@@ -204,7 +204,7 @@ class RepoSyncCommand(dnf.cli.Command):
         return True
 
     def get_pkglist(self, repo):
-        query = self.base.sack.query(flags=hawkey.IGNORE_EXCLUDES).available().filterm(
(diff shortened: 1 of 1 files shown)


pkratoch:
Looks good. I wrote some tests for this patch and assigned you for review: https://github.com/rpm-software-management/ci-dnf-stack/pull/696

pkratoch:
Now I noticed the commit message is a bit misleading, because you are applying only non-modular excludes (or rather ignoring only modular excludes). Can you perhaps replace "Apply" with "Ignore"?

m-blaha:
Thanks, good catch. Fixed.

pkratoch:
@rh-atomic-bot r+

PR #369, opened 2019-11-08
[doc] move manpages for plugins to "dnf-PLUGIN" (RhBug:1706386)

lukash:
Moves the man pages for the plugins from "dnf.plugin.PLUGIN" to
"dnf-PLUGIN" to follow the established convention for man pages of
binaries with commands.

In addition to replacing ".plugin." for a "-" in the man page filenames,
it also replaces a "_" with a "-" for the config-manager and
needs-restarting plugins.

https://bugzilla.redhat.com/show_bug.cgi?id=1706386

Same thing for `dnf-plugins-extras`: https://github.com/rpm-software-management/dnf-plugins-extras/pull/169


m-blaha:
Just small notice - https://dnf.readthedocs.io/en/latest/command_ref.html contains in See also part text that also should be changed:
dnf.plugin.*(8), assorted DNF plugins that might be installed on the system.
But the PR itself LGTM.

m-blaha:
Also shouldn't there be also consistent man pages for built-in commands? We can begin (as Michal suggests in the BZ) with dnf-<command>(8) being redirect to dnf(8).
But given that all these man pages would be only redirects, I'm not 100% convinced that such a change is for good.

lukash:
> Just small notice - https://dnf.readthedocs.io/en/latest/command_ref.html contains in See also part text that also should be changed:
> dnf.plugin.*(8), assorted DNF plugins that might be installed on the system.
> But the PR itself LGTM.

Good point. What do we do about it? Changing it to:
> dnf-*(8), assorted DNF plugins that might be installed on the system.

Doesn't seem great :slightly_frowning_face: 

> Also shouldn't there be also consistent man pages for built-in commands? We can begin (as Michal suggests in the BZ) with dnf-(8) being redirect to dnf(8).

How can I redirect a man page? With a symlink? Also, do you really mean redirecting `dnf-` with the dash at the end? What's that useful for?

> But given that all these man pages would be only redirects, I'm not 100% convinced that such a change is for good.

I've thought about it... unless I'm missing a simple way to handle this, we'd have to redirect (have a symlink?) each command explicitly, which would be a whole lot (taking into account aliases etc.) and I'm afraid we'd be adding more room for inconsistencies, which we already have plenty of between the built-in `--help` and man pages. Instead I'd consider going the way of git and systemd with splitting the man page per command and displaying the man pages on `--help`. In some future version of dnf, e.g. 5 :slightly_smiling_face:

lukash:
@m-blaha I've created https://github.com/rpm-software-management/dnf/pull/1528 for DNF.. have a look, before merging, maybe we should also add version dependencies to make sure the location of man pages and the reference in the DNF man page is always consistent?

m-blaha:
@rh-atomic-bot r+

m-blaha:
@rh-atomic-bot retry

m-blaha:
The tests that have failed were expected to do so. Merging.

PR #370, opened 2019-11-14
[versionlock] Use demands.plugin_filtering_enabled

m-blaha:
The original demands.resolving is used as a fall back if the
plugin_filtering_enabled is not set.

Requires: https://github.com/rpm-software-management/dnf/pull/1530
For testing you can use the copr repo: https://copr.fedorainfracloud.org/coprs/mblaha/versionlockdemand/


j-mracek:
@rh-atomic-bot r+

PR #371, opened 2019-11-14
[builddep] Accept --skip-broken switch (RhBug:1628634)

m-blaha:
Using --skip-broken makes builddep command tollerant to uninstallable
packages. All other (installable) build dependencies are going to be
installed.

https://bugzilla.redhat.com/show_bug.cgi?id=1628634

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/689

### DIFF
--- a/doc/builddep.rst
+++ b/doc/builddep.rst
@@ -39,6 +39,11 @@ All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for det
 ``--srpm``
     Treat arguments as source rpm files.
 
+``--skip-unavailable``
+    Skip build dependencies not available in repositories. All available build dependencies will be installed.
+
+Note that `builddep` command does not honor the `--skip-broken` option, so there is no way to skip uninstallable packages (e.g. with broken dependencies).
+
 --------
 Examples
 --------
--- a/plugins/builddep.py
+++ b/plugins/builddep.py
@@ -59,6 +59,8 @@ class BuildDepCommand(dnf.cli.Command):
         parser.add_argument('-D', '--define', action='append', default=[],
                             metavar="'MACRO EXPR'", type=macro_def,
                             help=_('define a macro for spec file parsing'))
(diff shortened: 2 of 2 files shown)


plugins/builddep.py
```diff
@@ -115,7 +115,10 @@ def run(self):
             rpm.delMacro(macro[0])
 
         if pkg_errors:
-            raise dnf.exceptions.Error(_("Some packages could not be found."))
+            msg = _("Some packages could not be found.")
+            if not self.opts.skip_broken:
+                msg += " " + _("Try to add '--skip-broken' to skip uninstallable packages.")
```
j-mracek: I would recommend to add this hint in documentation with warning but keep original statement here. I suggest that this is a workaround that could lead in unpredictable behavior during builds.

doc/builddep.rst
```diff
@@ -39,6 +39,11 @@ All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for det
 ``--srpm``
     Treat arguments as source rpm files.
 
+``--skip-unavailable``
+    Skip build dependencies not available in repositories. All available build dependencies will be installed.
+
+Note that `builddep` command does not honor the `--skip-broken` option so there is no way how to skip uninstallable packages (e.g. with broken dependencies).
```
j-mracek: Not important note: What about to replace `Note that` by `Note:`?

j-mracek:
LGTM!

doc/builddep.rst
lukash: FWIW I think saying _"Note that"_ is fine :slightly_smiling_face: 

But I'd throw a comma in there and leave out the _"how"_:
> Note that `builddep` command does not honor the `--skip-broken` option**,** so there is no **way to** skip uninstallable packages

m-blaha: Thanks Lukas

j-mracek:
@rh-atomic-bot r+

PR #373, opened 2019-11-25
Fix: config_manager respect config file location during save

jrohel:
DNF has the `--config` argument. It defines path to the configuration
file. The config_manager plugin used the configuration from the path
but the changes were allways saved into /etc/dnf/dnf.conf.
The PR fixes it. The changes are saved to the original
configuration file.


pkratoch:
All tests passed locally, so I am merging this.
Test for this feature: https://github.com/rpm-software-management/ci-dnf-stack/pull/711

PR #376, opened 2019-12-11
[versionlock] Prevent conflicting/duplicate entries (RhBug:1782052)

dhgutteridge:
Don't allow the same package NEVRA form to be recorded multiple times. It's either contradictory (if both add and exclude actions were submitted), or simply redundant.

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -33,6 +33,7 @@ NOT_READABLE = _('Unable to read version lock configuration: %s')
 NO_LOCKLIST = _('Locklist not set')
 ADDING_SPEC = _('Adding versionlock on:')
 EXCLUDING_SPEC = _('Adding exclude on:')
+EXISTING_SPEC = _('Package already locked:')
 DELETING_SPEC = _('Deleting versionlock for:')
 NOTFOUND_SPEC = _('No package found for:')
 NO_VERSIONLOCK = _('Excludes from versionlock plugin were not applied')
@@ -155,13 +156,21 @@ class VersionLockCommand(dnf.cli.Command):
                 cmd = self.opts.subcommand
 
         if cmd == 'add':
-            _write_locklist(self.base, self.opts.package, self.opts.raw, True,
-                            "\n# Added locks on %s\n" % time.ctime(),
-                            ADDING_SPEC, '')
+            entry = _search_locklist(self.opts.package)
+            if entry == '':
+                _write_locklist(self.base, self.opts.package, self.opts.raw, True,
+                                "\n# Added lock on %s\n" % time.ctime(),
+                                ADDING_SPEC, '')
+            else:
+                logger.info("%s %s", EXISTING_SPEC, entry)
         elif cmd == 'exclude':
-            _write_locklist(self.base, self.opts.package, self.opts.raw, False,
-                            "\n# Added exclude on %s\n" % time.ctime(),
-                            EXCLUDING_SPEC, '!')
+            entry = _search_locklist(self.opts.package)
+            if entry == '':
+                _write_locklist(self.base, self.opts.package, self.opts.raw, False,
+                                "\n# Added exclude on %s\n" % time.ctime(),
+                                EXCLUDING_SPEC, '!')
+            else:
+                logger.info("%s %s", EXISTING_SPEC, entry)
         elif cmd == 'list':
             for pat in _read_locklist():
                 logger.info(pat)
@@ -200,6 +209,16 @@ def _read_locklist():
     except IOError as e:
         raise dnf.exceptions.Error(NOT_READABLE % e)
     return locklist
+
+
+def _search_locklist(package):
+    found = ''
+    locked_specs = _read_locklist()
+    for ent in locked_specs:
+        if _match(ent, package):
+            found = ent
+            break
+    return found
 
 
 def _write_locklist(base, args, raw, try_installed, comment, info, prefix):
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -33,7 +33,8 @@ NOT_READABLE = _('Unable to read version lock configuration: %s')
 NO_LOCKLIST = _('Locklist not set')
 ADDING_SPEC = _('Adding versionlock on:')
 EXCLUDING_SPEC = _('Adding exclude on:')
-EXISTING_SPEC = _('Package already locked:')
+EXISTING_SPEC = _('Package already locked in equivalent form:')
+CONFLICTING_SPEC = _('Package already %s')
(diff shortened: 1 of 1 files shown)


m-blaha:
I created basic test scenarios for this patch: https://github.com/rpm-software-management/ci-dnf-stack/pull/732

plugins/versionlock.py
```diff
@@ -155,13 +156,21 @@ def run(self):
...
-            _write_locklist(self.base, self.opts.package, self.opts.raw, True,
-                            "\n# Added locks on %s\n" % time.ctime(),
-                            ADDING_SPEC, '')
+            entry = _search_locklist(self.opts.package)
+            if entry == '':
+                _write_locklist(self.base, self.opts.package, self.opts.raw, True,
+                                "\n# Added lock on %s\n" % time.ctime(),
+                                ADDING_SPEC, '')
+            else:
+                logger.info("%s %s", EXISTING_SPEC, entry)
```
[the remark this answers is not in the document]
dhgutteridge: Sorry, yes, you're right, it really should be an error if the action requested conflicts with what's already recorded. I've made some changes to return an error in the case of conflict, and to be more specific in general. (That means the output messages will now differ from the sample test cases you'd already set.) There's a distinction between the solution being "exactly the same pattern" and the user entering a different argument that ends up resolving into the same pattern. So that's why I've used the phrase "recorded in equivalent form" in the non-error condition.  (And then, of course, the pattern recorded is displayed in the message.)

I assumed you want to use dnf.exceptions.Error, since logger.error doesn't set ERRNO on exit. (And I tried to match the Pythonic idiom used elsewhere in the code base, which values terseness.)

vlours:
Hey @m-blaha @dhgutteridge,

I'm currently working on an Ansible module for yum/dnf versionlock. I discovered this "bug" during my tests. I'm glad someone is already fixing it.

But in the meantime, I found a second bug related to the dnf '-q' option (quiet). 
When using this option with versionlock, **ALL outputs are dismissed**. You even miss the add/exclude/delete actions.

#### Samples
```[root@5a94f6a929cb ~]# dnf -q versionlock clear
[root@5a94f6a929cb ~]# dnf -q versionlock add python
[root@5a94f6a929cb ~]# dnf -q versionlock list
[root@5a94f6a929cb ~]# dnf versionlock list
Last metadata expiration check: 1 day, 22:05:15 ago on Wed 18 Dec 2019 05:01:04 AM UTC.
python-0:2.7.5-86.el7.*
[root@5a94f6a929cb ~]# dnf -q versionlock exclude python
[root@5a94f6a929cb ~]# dnf -q versionlock exclude python
[root@5a94f6a929cb ~]# dnf versionlock list
Last metadata expiration check: 1 day, 22:05:39 ago on Wed 18 Dec 2019 05:01:04 AM UTC.
python-0:2.7.5-86.el7.*
!python-0:2.7.5-86.el7.*
!python-0:2.7.5-86.el7.*
[root@5a94f6a929cb ~]# dnf -q versionlock delete python
[root@5a94f6a929cb ~]# dnf versionlock list
Last metadata expiration check: 1 day, 22:05:49 ago on Wed 18 Dec 2019 05:01:04 AM UTC.
```

_Yum's plugin is working fine with the '-q' option and the duplicate entries_
```[root@5a94f6a929cb ~]# yum -q versionlock clear
[root@5a94f6a929cb ~]# yum -q versionlock add python
Adding versionlock on: 0:python-2.7.5-86.el7
[root@5a94f6a929cb ~]# yum -q versionlock list
0:python-2.7.5-86.el7.*
[root@5a94f6a929cb ~]# yum -q versionlock exclude python
Adding exclude on: 0:python-2.7.5-86.el7
[root@5a94f6a929cb ~]# yum -q versionlock exclude python
[root@5a94f6a929cb ~]# yum -q versionlock list
0:python-2.7.5-86.el7.*
!0:python-2.7.5-86.el7.*
[root@5a94f6a929cb ~]# yum -q versionlock delete python
Deleting versionlock for: 0:python-2.7.5-86.el7.*
Deleting versionlock for: !0:python-2.7.5-86.el7.*
[root@5a94f6a929cb ~]# yum -q versionlock list
[root@5a94f6a929cb ~]#
```

As there is no 'Issues' option available in this repo, I was wondering if one of you can have a look on this point.

#### Environment
Tested with: in a Centos 7.7 docker image with:
- dnf-4.0.9.2-1.el7_6.noarch
- python2-dnf-plugin-versionlock-4.0.2.2-3.el7_6.noarch
- yum-3.4.3-163.el7.centos.noarch
- yum-plugin-versionlock-1.1.31-52.el7.noar

m-blaha:
@vlours good catch. I filed a bug https://bugzilla.redhat.com/show_bug.cgi?id=1785563 based on your comment to have the issue tracked.

plugins/versionlock.py
```diff
@@ -155,13 +157,25 @@ def run(self):
...
-            _write_locklist(self.base, self.opts.package, self.opts.raw, True,
-                            "\n# Added locks on %s\n" % time.ctime(),
-                            ADDING_SPEC, '')
+            (entry, entry_cmd) = _search_locklist(self.opts.package)
+            if entry == '':
+                _write_locklist(self.base, self.opts.package, self.opts.raw, True,
+                                "\n# Added lock on %s\n" % time.ctime(),
+                                ADDING_SPEC, '')
+            elif cmd != entry_cmd:
+                raise dnf.exceptions.Error(CONFLICTING_SPEC % ('excluded: ' + entry))
```
m-blaha: This will not work with translations - the 'excluded' (and the 'locked' string below) will remain untranslated. I would prefer to have two complete strings (to make it clearer for translators) - 'Package {} is already locked' and 'Package {} is already excluded'. Or may be even more explicit - 'Cannot lock already excluded package {}'. Please rephrase it as you need (I'm not native speaker).

plugins/versionlock.py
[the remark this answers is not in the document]
m-blaha: Yes, raising dnf.exceptions.Error is exactly what I meant.

plugins/versionlock.py
dhgutteridge: Thanks, I've redone this as you suggested.

dhgutteridge: One other minor detail: I didn't apply lstrip('!') to entry in the error message when it's an exclusion record. Personally, I prefer to see a raw version of an entry string. But you may find that less user-friendly.

m-blaha:
@dhgutteridge Thanks for your work!

m-blaha:
@rh-atomic-bot r+

m-blaha:
@rh-atomic-bot retry

m-blaha:
@rh-atomic-bot retry

PR #377, opened 2020-01-03
[config-manager] Allow use of --set-enabled without arguments (RhBug:…

pkratoch:
…1679213)

Since config-manager was enhanced to also modify repositories specified
by repoids in the --setopt option, it should no longer be required to
specify repoids as arguments for --set-enabled.

As a consequence, "config-manager --set-enabled" without any other
argument will exit with 0 and have no effect (same as "--set-disabled").

https://bugzilla.redhat.com/show_bug.cgi?id=1679213

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -67,11 +67,6 @@ class ConfigManagerCommand(dnf.cli.Command):
 
     def run(self):
         """Execute the util action here."""
-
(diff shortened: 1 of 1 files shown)


pkratoch:
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/749

PR #378, opened 2020-01-07
Fix: don't open stdin if versionlock is missing

inknos:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1785563
Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/783

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -179,7 +179,7 @@ class VersionLockCommand(dnf.cli.Command):
                 logger.info("%s %s", EXISTING_SPEC, entry)
         elif cmd == 'list':
             for pat in _read_locklist():
-                logger.info(pat)
+                print(pat)
         elif cmd == 'clear':
             with open(locklist_fn, 'w') as f:
                 # open in write mode truncates file
@@ -192,7 +192,7 @@ class VersionLockCommand(dnf.cli.Command):
             with os.fdopen(out, 'w', -1) as out:
                 for ent in locked_specs:
                     if _match(ent, self.opts.package):
-                        logger.info("%s %s", DELETING_SPEC, ent)
+                        print("%s %s" % (DELETING_SPEC, ent))
                         count += 1
                         continue
                     out.write(ent)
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@inknos Can you please fix the authorship of your first commit? It looks like it has your computer's username and hostname instead of your correct identity.

pkratoch:
There are actually two fixes, right? One is that the messages are printed on stdout instead of logged and that fixes the bug 1785563.

The other one is to error out if ``locklist_fn`` is not set. The problem with this is that when I have versionlock installed and I delete the ``locklist`` option from ``versionlock.conf``, then I get an error even when I run ``dnf upgrade``. Which I don't think is desirable. We should distinguish between running the versionlock command (in that case, it should error out) and running some other command (in that case, there should be just a warning, such as "versionlock enabled, but locklist not set"). 

Also, we don't check if ``locklist_fn`` is set in the ``_write_locklist`` (running ``dnf versionlock add libdnf`` prints that it adds libdnf, but it actually doesn't have any effect), so if you could add it, too, that would be great.

plugins/versionlock.py
```diff
@@ -192,6 +192,8 @@ def run(self):
 def _read_locklist():
     locklist = []
     try:
+        if not locklist_fn:
+            raise dnf.exceptions.Error(NO_LOCKLIST)
```
inknos: These lines just check if the filename is valid. If you are not checking the file name you can try to open a `NaN` which will reault in opening a input stream.  
> The problem with this is that when I have versionlock installed and I delete the locklist option from versionlock.conf, then I get an error even when I run dnf upgrade. Which I don't think is desirable

This is actually expected

inknos: >Also, we don't check if locklist_fn is set in the _write_locklist (running dnf versionlock add libdnf prints that it adds libdnf, but it actually doesn't have any effect), so if you could add it, too, that would be great.

I will add the check inside the `_write_locklist` as it is in `_read_locklist`

pkratoch:
There was another change to versionlock plugin recently: https://github.com/rpm-software-management/dnf-plugins-core/pull/376
Surprisingly, there are no conflicts, but it changes the behaviour because for "add" and "exclude" it reads the locklist before writing into it, so with missing locklist, it raises an error instead of creating new file.
Can you please rebase to newest master and fix this problem?

pkratoch:
@rh-atomic-bot r+

PR #381, opened 2020-01-15
config-manager calls parser error when without options (RhBug:1782822)

inknos:
The arguments are checked directly inside configure.
Parser error called explicitly
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1782822

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -61,16 +61,15 @@ class ConfigManagerCommand(dnf.cli.Command):
         demands.available_repos = True
 
         # if no argument was passed then error
-        if (not (self.opts.crepo != [] or
-                 self.opts.add_repo != [] or
+        if (not (self.opts.add_repo != [] or
                  self.opts.save or
                  self.opts.dump or
                  self.opts.dump_variables or
                  self.opts.set_disabled or
                  self.opts.set_enabled) ):
-            self.cli.optparser.error(_("one of the arguments {} is required")
+            self.cli.optparser.error(_("one of the following arguments is required: {}")
                                      .format(' '.join([
-                                         "repo", "--save", "--add-repo",
+                                         "--save", "--add-repo",
                                          "--dump", "--dump-variables",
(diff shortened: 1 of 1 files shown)


inknos:
Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/776

plugins/config_manager.py
```diff
@@ -60,6 +60,20 @@ def configure(self):
...
+                 self.opts.save or
+                 self.opts.dump or
+                 self.opts.dump_variables or
+                 self.opts.set_disabled or
+                 self.opts.set_enabled) ):
+            self.cli.optparser.error(_("one of the arguments {} is required")
+                                     .format(' '.join([
+                                         "repo", "--save", "--add-repo",
+                                         "--dump", "--dump-variables",
+                                         "--enable", "--disable"])))
```
pkratoch: I think that ``dnf config-manager fedora`` should also fail, because there is no action to do. So, can you remove the "repo" option(s) from the list?

plugins/config_manager.py
```diff
@@ -60,6 +60,20 @@ def configure(self):
...
 
+        # if no argument was passed then error
+        if (not (self.opts.crepo != [] or
+                 self.opts.add_repo != [] or
+                 self.opts.save or
+                 self.opts.dump or
+                 self.opts.dump_variables or
+                 self.opts.set_disabled or
+                 self.opts.set_enabled) ):
+            self.cli.optparser.error(_("one of the arguments {} is required")
```
pkratoch: This format is not very readable. I'd rather put the list of arguments at the very end, such as: "one of the following arguments is required: {}".

pkratoch:
@rh-atomic-bot r+

pkratoch:
The failing test scenario is expected and adjusted in the PR https://github.com/rpm-software-management/ci-dnf-stack/pull/776 , so I am merging both.

PR #383, opened 2020-01-20
Drop support of tito release command

pkratoch:


xsuchy:
What you are using for pushing to dist-git nowadays?

pkratoch:
> What you are using for pushing to dist-git nowadays?

Currently, we don't have any good tool, but we are not using tito for quite some time now. This is just a cleanup, same as in dnf: https://github.com/rpm-software-management/dnf/pull/1566/commits/e8e82503ce3fe998f55e1118866c58cd9282f2b8
Does this change affect you in any way?

PR #384, opened 2020-01-29
Update reposync.py

lvg01:
Added option --omit-repoid for backwards compatibility of reposync behaviour.
THis options does not append the repoid to the downloadpath.

### DIFF
--- a/plugins/reposync.py
+++ b/plugins/reposync.py
@@ -71,6 +71,8 @@ class RepoSyncCommand(dnf.cli.Command):
                             help=_('download only newest packages per-repo'))
         parser.add_argument('-p', '--download-path', default='./',
                             help=_('where to store downloaded repositories'))
+        parser.add_argument('--norepopath', default=False, action='store_true',
+                            help=_("Don't add the reponame to the download path."))
         parser.add_argument('--metadata-path',
                             help=_('where to store downloaded repository metadata. '
                                    'Defaults to the value of --download-path.'))
@@ -101,6 +103,10 @@ class RepoSyncCommand(dnf.cli.Command):
 
         if self.opts.source:
             repos.enable_source_repos()
+
+        if len(list(repos.iter_enabled())) > 1 and self.opts.norepopath:
+            raise dnf.cli.CliError(
+                _("Can't use --norepopath with multiple repositories"))
 
         for repo in repos.iter_enabled():
             repo._repo.expire()
@@ -148,7 +154,8 @@ class RepoSyncCommand(dnf.cli.Command):
                 self.delete_old_local_packages(repo, pkglist)
 
     def repo_target(self, repo):
-        return _pkgdir(self.opts.destdir or self.opts.download_path, repo.id)
+        return _pkgdir(self.opts.destdir or self.opts.download_path,
+                       repo.id if not self.opts.norepopath else '')
 
     def metadata_target(self, repo):
         if self.opts.metadata_path:
--- a/doc/reposync.rst
+++ b/doc/reposync.rst
@@ -41,6 +41,9 @@ All general DNF options are accepted. Namely, the ``--repoid`` option can be use
 
 ``-p <download-path>, --download-path=<download-path>``
     Root path under which the downloaded repositories are stored, relative to the current working directory. Defaults to the current working directory. Every downloaded repository has a subdirectory named after its ID under this path.
+    
+``--norepopath``
+    Don't add the reponame to the download path. Can only be used when syncing a single repository (default is to add the reponame).
 
 ``--download-metadata``
     Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c on it.


plugins/reposync.py
```diff
@@ -148,7 +150,7 @@ def run(self):
                 self.delete_old_local_packages(repo, pkglist)
 
     def repo_target(self, repo):
-        return _pkgdir(self.opts.destdir or self.opts.download_path, repo.id)
+        return _pkgdir(self.opts.destdir or self.opts.download_path, repo.id if not self.opts.omit_repoid else '')
```
m-blaha: please split this line into two to keep line length under 100 charactere.

plugins/reposync.py
```diff
@@ -71,6 +71,8 @@ def set_argparser(parser):
                             help=_('download only newest packages per-repo'))
         parser.add_argument('-p', '--download-path', default='./',
                             help=_('where to store downloaded repositories'))
+        parser.add_argument('--omit-repoid', default=False, action='store_true',
+                            help=_('don\'t add repoid to download-path'))
```
m-blaha: The yum implementation of reposync used --norepopath switch for this functionality. Is there any reason not to use it also here?
Also yum allowed --norepopath only when just one repository was about to be synced. I think this restriction makes sense - otherwise repositories content would be mixed together. What do you think?

See https://github.com/rpm-software-management/yum-utils/blob/12dc3da81c8b200fbac15e8fd70d777ff4bb57dc/reposync.py#L108

plugins/reposync.py
lvg01: I've splitted the line

plugins/reposync.py
lvg01: Indeed, you're correct. The original --norepopath does the same. It's better to use that as optionname. Sorry for that, I will change this.

m-blaha:
And what do you think about limiting usage of the --norepopath switch only to situations where just one repository is about to be downloaded? Otherwise the downloaded contents of multiple repositories will be mixed together. Especially when also --download-metadata switch is used and the "repodata" directory will contain only metadata of the latest downloaded repository.

plugins/reposync.py
```diff
@@ -71,6 +71,8 @@ def set_argparser(parser):
                             help=_('download only newest packages per-repo'))
         parser.add_argument('-p', '--download-path', default='./',
                             help=_('where to store downloaded repositories'))
+        parser.add_option('--norepopath', default=False, action="store_true",
+                            help=_('Don't add the reponame to the download path.')
```
m-blaha: You have syntax error here - either escape: `'Don\'t...` or use double quotes around help string: `"Don't add..."`

lvg01: In fact I had this one escaped but when I copied from the yum version of norepopath it disappeared. Iǘe now used the double quote syntax therefore.

lvg01:
The one-repo limitation to norepopath is a good onbe. It was not my use-case but I will look into that tomorrow.

m-blaha:
sorry, i forgot my review comments in "pending" state...

plugins/reposync.py
```diff
@@ -102,6 +104,10 @@ def configure(self):
         if self.opts.source:
             repos.enable_source_repos()
 
+        if len(repos) > 1 and self.opts.norepopath:
+            raise dnf.exceptions.Error(
+                _("Error: Can't use --norepopath with multiple repositories"))
```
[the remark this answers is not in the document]
lvg01: I ysed the reused the code from the pkg_download_path function. But the suggested method is beter I suppose. I've changed the code accordingly.

plugins/reposync.py
```diff
@@ -71,6 +71,8 @@ def set_argparser(parser):
                             help=_('download only newest packages per-repo'))
         parser.add_argument('-p', '--download-path', default='./',
                             help=_('where to store downloaded repositories'))
+        parser.add_option('--norepopath', default=False, action="store_true",
+                            help=_("Don't add the reponame to the download path.")
```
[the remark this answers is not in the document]
lvg01: OK

m-blaha:
So the code looks good now. But before merging I would like to have all the commits squashed into one (well, the documentation can be left in a separate commit) with a commit message that would summarize the change.

lvg01:
I understand what you mean but I'm not so experienced in GIT. I've been rebasing and the 45a9d30 commit seems the one which does this. I don know how to get rid of all other commits.... Maybe you can help me with that?

m-blaha:
The easiest way to do it is interactive rebase to the master branch:
$ git rebase -i master
Then you will be given the list of the commits and a help on commands:

```
pick 6ef0aa2 Update reposync.py with --norepopath option
pick 45a9d30 Update reposync.py with --norepopath option
pick 72aad1f Update reposync.py
pick a499c85 Splitted codeline > 100 characters
pick 89cf668 Added documentation for --omit-repoid option
pick 52d998c Change option to yum existing option --norepopath
pick 36c5e7c Updated --norepopath option documentation
pick 5a4c9b1 Fix quote syntax in string
pick 892a4f2 Raise Error on use of --norepopath and multiple repos selected
pick 8ffe4a9 Updated with latest suggestions from review
pick 7d3db1e Resolve missing parenthasis
pick b184730 Use parser.add_argument instead of .option
```


Given that the very first commit in the list already contains all required changes, you can remove all lines except the first one (what you usually do here is reorder the commits, mark some of them for squashing, reword some messages. But in this specific case you want only the last commit to remain). The file will after editing look like this:

```
pick 6ef0aa2 Update reposync.py with --norepopath option
```

Save, and you should end up with only one commit in the history.
Then you can push to the github (you will need to use --force switch) and you are done.

lvg01:
OK, that was the trick. I was a bit scary for using --force... Thank you for your help with the complete review!

m-blaha:
@rh-atomic-bot r+

PR #386, opened 2020-02-14
Repoclosure for modules

j-mracek:
Requires: https://github.com/rpm-software-management/libdnf/pull/1005

### DIFF
--- a/plugins/repoclosure.py
+++ b/plugins/repoclosure.py
@@ -58,7 +58,7 @@ class RepoClosureCommand(dnf.cli.Command):
 
     def run(self):
         arch = self.opts.arches if self.opts.arches else None
-        if self.opts.modules:
+        if self.opts.modules or self.opts.modules_with_defaults:
             module_dict, base_query, non_modular = self._prepare_module_data()
             module_string_list = sorted(module_dict.keys())
             combinations, broken_module_dict = self._get_module_combinations(module_string_list)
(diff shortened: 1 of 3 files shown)


plugins/repoclosure.py
```diff
@@ -53,20 +54,239 @@ def configure(self):
...
+            problem = self._analyze_modular_combinations(
+                arch, combinations, base_query, module_dict, non_modular)
+            if problem or broken_module_dict:
+                self._raise_unresolved_dependencies()
         else:
-            unresolved = self._get_unresolved()
+            unresolved = self._get_unresolved(arch, self.base.sack.query().available(),
+                                              self.base.sack.query().available())
+            self._report_results_to_terminal(unresolved)
+            if len(unresolved) > 0:
```
sandrobonazzola: Why not just "if unresolved:" ?

sandrobonazzola:
@j-mracek can you please comment on why this got closed?

j-mracek:
I am sorry, I forget to add a description. The reason is that DNF is going to be replaced by DNF5 in Fedora 39. Therefore we will create another implementation for DNF5 anyway. Do you think it make sense to get the feature to current code base?

sandrobonazzola:
No strong feelings, it would still be useful if backported to RHEL8 or RHEL9 but on Fedora I don't see a need.

j-mracek:
@sandrobonazzola Ok, for RHEL we need a bug report on bugzilla.

sandrobonazzola:
I see someone already did, is https://bugzilla.redhat.com/show_bug.cgi?id=2183091 enough?

j-mracek:
> I see someone already did, is https://bugzilla.redhat.com/show_bug.cgi?id=2183091 enough?

This is something else. It is about presence of fail safe metadata on the system

PR #388, opened 2020-03-02
copr: add warning for enabling runtime dependencies

dturecek:


plugins/copr.py
```diff
@@ -463,21 +464,50 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+            if section in self.repo_filename:
+                continue
+            runtime_deps += section + "\n"
+
+        msg = _("""
+The Copr project {0} uses the following repositories as runtime dependencies:
+{1}
+Do you want to keep them enabled?""".format(project_name, runtime_deps))
+
+        if self.base._promptWanted():
```
FrostyX: I see that we already use this in `_ask_user`, but we should be careful about that since it is a private variable.

plugins/copr.py
```diff
@@ -239,7 +239,7 @@ def run(self):
             copr_projectname = project[2]
             project_name = copr_username + "/" + copr_projectname
 
-        repo_filename = "{0}/_copr:{1}:{2}:{3}.repo".format(
+        self.repo_filename = "{0}/_copr:{1}:{2}:{3}.repo".format(
```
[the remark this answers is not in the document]
dturecek: Well, we change `repo_filename` in `_download_repo` based on the downloaded repo's ID, so I could either do it the way I do it in the commit or make `_download_repo` return the changed value or call `_runtime_deps_warning` from inside `_download_repo`. The way I did it seemed nicer to me at the time.

plugins/copr.py
```diff
@@ -463,21 +464,50 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        os.chmod(self.repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+
+    def _runtime_deps_warning(self, project_name, chroot=None):
+        repo_file = ConfigParser()
+        repo_file.read(self.repo_filename)
+
+        if len(repo_file.sections()) == 1:
+            return
+
+        runtime_deps = ""
```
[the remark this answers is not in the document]
dturecek: Thanks, I agree it'd be better with a list.

plugins/copr.py
dturecek: So what do you propose to do with this?

dturecek:
I've added a list of runtime dependencies.

plugins/copr.py
```diff
@@ -463,21 +464,49 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        if len(repo_file.sections()) == 1:
+            return
+
+        runtime_deps = []
+        for section in repo_file.sections():
+            if section in self.repo_filename:
+                continue
+            runtime_deps.append(section)
+
+        print("The Copr project {0} uses the following repositories as runtime dependencies:"\
```
FrostyX: I think you should use `_` so the sentence can be translated into other languages. IIRC you had it before and it disappeared during the last change.

IMHO you don't need the trailing `\`

plugins/copr.py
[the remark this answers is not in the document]
FrostyX: I see, thank you for the explanation.

plugins/copr.py
FrostyX: I would keep it as is. The function is pretty straightforward and likely will not change

    def _promptWanted(self):
        # shortcut for the always-off/always-on options
        if self.conf.assumeyes and not self.conf.assumeno:
            return False
        return True

I just brought it up in case you wanted to talk to the DNF team and tell them that it is a useful function, that we use, and if they would like to add it to their API. But I guess I can just ... @dmach ^^

plugins/copr.py
dturecek: Fixed.

plugins/copr.py
```diff
@@ -463,21 +464,49 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
 
-        shutil.copy2(f.name, repo_filename)
-        os.chmod(repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+        shutil.copy2(f.name, self.repo_filename)
+        os.chmod(self.repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+
+    def _runtime_deps_warning(self, project_name, chroot=None):
+        repo_file = ConfigParser()
+        repo_file.read(self.repo_filename)
+
```
praiskup: As we discussed [here](https://github.com/rpm-software-management/mock/pull/454/files#diff-930178008d34e91de7c2ae4a81516511R309), I'd suggest you to use strict=False.

plugins/copr.py
```diff
@@ -463,21 +464,49 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        if len(repo_file.sections()) == 1:
+            return
+
+        runtime_deps = []
+        for section in repo_file.sections():
+            if section in self.repo_filename:
```
praiskup: This feels a bit unclear.  Deserves proper comment ....   btw. we probably
need pattern like `copr:*` something ... because e.g. multilib projects at
least have two repos by default ...
I don't really like that we bind the filename to section name.

plugins/copr.py
```diff
@@ -463,21 +464,49 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+                continue
+            runtime_deps.append(section)
+
+        print(_("The Copr project {0} uses the following repositories as runtime dependencies:"
+              .format(project_name)))
```
praiskup: Don't you want to print it in format
```
[repoid]
baseurl=...
```

praiskup: Ah, it's printed as complete section... OK.

praiskup: It is not printed as whole repo, only this is printed:
```
The Copr project praiskup/ping uses the following repositories as runtime dependencies:
copr:copr.fedorainfracloud.org:praiskup:ping:ml
```
It would be nice to see more than just the ID, so users can properly review before
they ACK it.  So please consider format:

```
The Copr project praiskup/ping uses the following repositories as runtime dependencies:
1. [copr:copr.fedorainfracloud.org:praiskup:ping:ml]
   baseurl=something
2. [something_else]
   baseurl=something_else
...
```

plugins/copr.py
```diff
@@ -463,21 +464,49 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        for dep in runtime_deps:
+            print(dep)
+
+        msg = _("Do you want to keep them enabled?")
+        if self.base._promptWanted():
+            if self.base.conf.assumeno or not self.base.output.userconfirm(
+                    msg='{} [y/N]: '.format(msg), defaultyes_msg='{} [Y/n]: '.format(msg)):
```
praiskup: Does this mean that if `dnf -y` isn't used, default is no?  (that would be correct)

praiskup:
`dnf copr disable` doesn't disable the dependencies

plugins/copr.py
dturecek: That's correct.

plugins/copr.py
dturecek: I've changed it to check the repo name that should always start with `Runtime dependency ...`

plugins/copr.py
dturecek: Changed the warning to this format.

dturecek:
I've addressed all your comments (except the one about repo id and repo filename mapping). Please take another look.

plugins/copr.py
```diff
@@ -463,21 +464,55 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+
+    def _runtime_deps_warning(self, project_name, chroot=None):
+        repo_file = ConfigParser(strict=False)
+        repo_file.read(self.repo_filename)
+
+        runtime_deps = []
+        for repo in repo_file.sections():
+            # repository is the project itself
+            if repo in self.repo_filename:
+                continue
```
praiskup: This can not work for repositories that contain `:ml`, right?  Also, old
copr versions had different variant of repoid/filenames.

Cam you invent something like repoid starting with  `coprdep:....` on frontend,
and do the if based on that?

dturecek: You're right, repositories with `:ml` wouldn't work. Changed the repo ID as you suggested.

plugins/copr.py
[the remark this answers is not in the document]
praiskup: Do we still need this change?

plugins/copr.py
```diff
@@ -463,21 +464,52 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        os.chmod(self.repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+
+    def _runtime_deps_warning(self, project_name, chroot=None):
+        repo_file = ConfigParser(strict=False)
+        repo_file.read(self.repo_filename)
+
+        runtime_deps = []
+        for repo in repo_file.sections():
+            if not repo.startswith("coprdep:"):
+                continue
```
praiskup, line 490: Please use `if repo.startswith("copr:"): continue`.  It is more secure.

plugins/copr.py
```diff
@@ -529,6 +561,29 @@ def _disable_repo(self, copr_username, copr_projectname):
             self.base.conf.write_raw_configfile(repo.repofile, multilib_repo_id,
                                                 self.base.conf.substitutions, {"enabled": "0"})
 
+        # disable runtime dependencies as well
+        repo_filename = "{0}/_copr:{1}:{2}:{3}.repo".format(self.base.conf.get_reposdir,
+                                                            self.copr_hostname.rsplit(":", 1)[0],
+                                                            self._sanitize_username(copr_username),
+                                                            copr_projectname)
```
praiskup, line 568: Why not to use `repo.repofile`?

plugins/copr.py
```diff
@@ -529,6 +561,22 @@ def _disable_repo(self, copr_username, copr_projectname):
...
+
+        for section in repo_file.sections():
+            if "Runtime dependency" not in repo_file.get(section, "name"):
+                continue
+            repo_file.set(section, "enabled", "0")
+            repo_changed = True
+
+        if repo_changed:
+            with open(repo_filename, "w") as repo:
+                repo_file.write(repo)
```
praiskup, line 577: This is still really, really ugly.  What if we simply added - on copr fe side - something
like `copr_dep_of=copr:safdsfdasd:asdfasdf`, and - instead of another ConfigParser
hack went through all the `if multilib_repo_id in self.base.repos` which have
`copr_dep_of=<COPR_ID>`?

praiskup: I mean, we could disable the repo the same way as we do with the normal repo.

plugins/copr.py
```diff
@@ -463,21 +464,52 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+
+        msg = _("Do you want to keep them enabled?")
+        if self.base._promptWanted():
+            if self.base.conf.assumeno or not self.base.output.userconfirm(
+                    msg='{} [y/N]: '.format(msg), defaultyes_msg='{} [Y/n]: '.format(msg)):
+                for dep in runtime_deps:
+                    repo_file.set(dep, "enabled", "0")
+
+                with open(self.repo_filename, "w") as f:
+                    repo_file.write(f)
```
praiskup, line 512: What if we used the same mechanism as we use for `disable`?  (Instead of writing the file from scratch)

plugins/copr.py
praiskup: I was told by @j-mracek that it will throw some warnings into `dnf -v` output,
but I don't view this as huge problem.  Same as we can have several warnings
because of duplicate repo ID enabled as a result of dep hell.

We could have another config file with dependency tree (I don't know whether
@j-mracek thought this in 1:1 irc chat).  The question is whether it's worth it
in this case (my guess is no).

praiskup: Simply speaking, id much rather iterate over all repositories - and disable those
which are special - then reading the config file again by ConfigParser, and
parse it again.

plugins/copr.py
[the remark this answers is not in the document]
praiskup: Still valid question.  Much nicer would be to just call `_runtime_deps_warning` from `_download_repo` directly, or at least it could simply accept the same arguments as the `_disable_repo`.

plugins/copr.py
```diff
@@ -463,21 +464,52 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
             if os.path.exists(old_repo_filename):
                 os.remove(old_repo_filename)
 
-        shutil.copy2(f.name, repo_filename)
-        os.chmod(repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+        shutil.copy2(f.name, self.repo_filename)
+        os.chmod(self.repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
+
+    def _runtime_deps_warning(self, project_name, chroot=None):
+        repo_file = ConfigParser(strict=False)
```
praiskup, line 484: Please also consider using dnf API instead:
```
self.base.reset(repos=True)                                                                                                                                                                                                            
self.base.read_all_repos()
```
From quick try, this should do what we need it from.  And then we could simply
do `_disable_repo(copr_username, copr_projectname, deps_only=True)`.
what we do in `_disable_copr()` except for

dturecek: Yeah, this seems to be working. Thanks for the suggestion.

dturecek:
I've fixed the PR. I was even able to simplify it a little bit (since I now use `_get_copr_repo`, I don't need the `self.repo_filename` anymore).

plugins/copr.py
j-mracek: Using a repo file for plugin configuration or external data is not good idea for long or final solution, but it is fine as a prove of concept.

.repo files have some disadvantages. They are difficult to update, because they are editable by an end user. After modification they cannot be easily updated by rpm. In case of origin from plugin (copr, substriction-manager) they are updated but with lose of information provided by user or not updated at all.

.repo files have also some advantages. They are loaded before metadata therefore the set of required repositories can be adjusted.

Also implementation of the functionality only to DNF is not suitable as a long term solution. Think like that you want to create a new industry standard and not only to enhance DNF plugin.

plugins/copr.py
```diff
@@ -529,6 +562,13 @@ def _disable_repo(self, copr_username, copr_projectname):
             self.base.conf.write_raw_configfile(repo.repofile, multilib_repo_id,
                                                 self.base.conf.substitutions, {"enabled": "0"})
 
+        # disable runtime dependencies as well
+        for repo_id in repo.cfg.sections():
+            if repo_id.startswith("copr:"):
+                continue
```
praiskup, line 568: It is much nicer now, but now I see potential to remove the special-cased
`copr:...` and `copr:...:ml` repos.  Could you consolidate the code and disable
everything by this loop?

plugins/copr.py
```diff
@@ -479,6 +480,38 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+            runtime_deps.append(repo_id)
+
+        if not runtime_deps:
+            return
+
+        print(_("The Copr project {0} uses the following repositories as runtime dependencies:"
+              .format(copr_projectname)))
+        dep_idx = 1
+        for repo_id in runtime_deps:
+            print("{0}. [{1}]".format(dep_idx, repo_id))
```
praiskup, line 502: Can you do this?  It makes the output more airy.  If you disagree, no problem (it's matter of taste I guess).
```
--- /usr/lib/python3.8/site-packages/dnf-plugins/copr.py        2020-04-21 07:39:33.965821445 +0200
+++ /tmp/backup.py      2020-04-21 07:38:04.733792653 +0200
@@ -495,15 +495,15 @@
         if not runtime_deps:
             return
 
-        print(_("\nThe Copr project {0} uses the following repositories as runtime dependencies:"
+        print(_("The Copr project {0} uses the following repositories as runtime dependencies:"
               .format(copr_projectname)))
         dep_idx = 1
         for repo_id in runtime_deps:
-            print("\n{0}. [{1}]".format(dep_idx, repo_id))
+            print("{0}. [{1}]".format(dep_idx, repo_id))
             print("   baseurl={0}".format(repo.cfg.getValue(repo_id, "baseurl")))
             dep_idx += 1
 
-        msg = _("\nDo you want to keep them enabled?")
+        msg = _("Do you want to keep them enabled?")
         if self.base._promptWanted():
             if self.base.conf.assumeno or not self.base.output.userconfirm(
                     msg='{} [y/N]: '.format(msg), defaultyes_msg='{} [Y/n]: '.format(msg)):
```

plugins/copr.py
```diff
@@ -479,6 +480,38 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
         shutil.copy2(f.name, repo_filename)
         os.chmod(repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
 
+    def _runtime_deps_warning(self, copr_username, copr_projectname, chroot=None):
```
praiskup, line 483: Per pylint, `chroot` arg is unused.

plugins/copr.py
```diff
@@ -479,6 +480,38 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        for repo_id in repo.cfg.sections():
+            if repo_id.startswith("copr:"):
+                continue
+            runtime_deps.append(repo_id)
+
+        if not runtime_deps:
+            return
+
+        print(_("The Copr project {0} uses the following repositories as runtime dependencies:"
+              .format(copr_projectname)))
```
praiskup, line 499: the `.format` should be moved two spaces to the right

praiskup:
@j-mracek wrote:
> Using a repo file for plugin configuration or external data is not good idea
> for long or final solution, but it is fine as a prove of concept.

Agreed.

> .repo files have some disadvantages. They are difficult to update, because
> they are editable by an end user. After modification they cannot be easily
> updated by rpm. In case of origin from plugin (copr, substriction-manager)
> they are updated but with lose of information provided by user or not updated
> at all.

Agreed, but this is about our repo file handling design, not about the
format.  We could have /usr/share/dnf/repos/*.repo files maintained by
RPM, and have manual overrides under /etc.  There are probably other ways
around.

> .repo files have also some advantages. They are loaded before metadata
> therefore the set of required repositories can be adjusted.

Yes.  One of the main reasons we picked this variant.  Baking deps into
metadata format is be possible (even though it would eat quite some
manpower to implement in copr), but I personally did not want to allow
installing another set of repositories without explicit end-user ack.

So once (if ever) there's some more standard way implemented in dnf stack,
for copr it will very much depend on how dnf will allow copr to tweak
(enable) the set of transitive (external) repo dependencies.

> Also implementation of the functionality only to DNF is not suitable as a long
> term solution. Think like that you want to create a new industry standard and
> not only to enhance DNF plugin.

In this particular case (on dnf level we need something much more robust),
my idea was to minimize the pain by moving the main copr.py logic at least
to python-copr package, and let dnf-plugins-core (and anyone else
interested..) to just consume our python api.

dturecek:
I''ve addressed all your points.

praiskup:
Thank you.  I tested now:
```
$ sudo dnf copr enable local/praiskup/tar epel-6-x86_64
Updating Subscription Management repositories.

You are about to enable a Copr repository. Please note that this
repository is not part of the main distribution, and quality may vary.

The Fedora Project does not exercise any power over the contents of
this repository beyond the rules outlined in the Copr FAQ at
<https://docs.pagure.org/copr.copr/user_documentation.html#what-i-can-build-in-copr>,
and packages are not held to any quality or security level.

Do you really want to enable localhost:55555/praiskup/tar? [y/N]: y
Repository successfully enabled.

The Copr project tar uses the following repositories as runtime dependencies:

1. [coprdep:localhost:praiskup:rpm-config]
   baseurl=http://copr-be.cloud.fedoraproject.org/results/praiskup/rpm-config/epel-6-$basearch/

2. [https_example_com_blah]
   baseurl=https://example.com/blah

Do you want to keep them enabled? [y/N]: n
```
.. and it just works.  Only the normal + multilib repo is enabled.

`sudo dnf -y copr enable local/praiskup/tar epel-6-x86_64` command enables
them without asking, but prints that they are enabled.  Good.

LGTM.

plugins/copr.py
```diff
@@ -479,6 +480,38 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+
+        print(_("\nThe Copr project {0} uses the following repositories as runtime dependencies:"
+                .format(copr_projectname)))
+        dep_idx = 1
+        for repo_id in runtime_deps:
+            print("\n{0}. [{1}]".format(dep_idx, repo_id))
+            print("   baseurl={0}".format(repo.cfg.getValue(repo_id, "baseurl")))
+            dep_idx += 1
+
+        msg = _("\nDo you want to keep them enabled?")
```
xsuchy, line 506: nitpick comment: should this be "Do you want to enable it?"

praiskup: It's one I requested, to make the output with separated blocks.
The newline can though be placed on a different place in code of course..

dturecek: The output looks likes this:

```
Do you really want to enable localhost:5000/dturecek/test-dependencies? [y/N]: y
Repository successfully enabled.

The Copr project test-dependencies uses the following repositories as runtime dependencies:

1. [coprdep:localhost:dturecek:foo]
   baseurl=http://backend_httpd:5002/results/dturecek/foo/fedora-$releasever-$basearch/
...
Do you want to keep them enabled? [y/N]:
```

so it seemed weird to me to first write that the repo was enabled and then ask if you want to enable the dependencies. However, I don't mind to change the message if you find it confusing.

xsuchy: I am enabling "dturecek/test-dependencies" and I do not have "dturecek:foo" enabled (yet). So it does not have the sense to ask to **keep** them enabled. It does not have a sense from the language point of view.

dturecek: I get your point. However, when you are enabling `test-dependencies`, we ask if you really want to enable it. Then, it is enabled (and thus `foo` is enabled as well) and then we ask if you want to keep it enabled. So I'd argue it makes sense.

Anyway, I see that it is confusing and I don't really mind to change the message.

dturecek:
I've changed the wording of the message.

plugins/copr.py
praiskup: I agree that we need better text.  Proposal? E.g. this?
```
Maintainer of the enabled Copr repository decided to make it dependant on
other repositories.  Such repositories are usually necessary for
successful installation of RPMs from the main Copr repository (runtime
dependencies provided).

Be aware that the note about quality and bug-reporting above applies here
too, Fedora Project doesn't control the content.  Please review the list:

... the list ...

Do you want to enable the repositories, too? [y/N]:
```

Sure, ... if something goes wrong (CTRL+C, e.g.) the repositories will
stay enabled.  But that's not a big issue.

xsuchy:
```
Do you really want to enable localhost:5000/dturecek/test-dependencies? [y/N]: y
Repository successfully enabled.
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ I read this that the dturecek/test-dependencies ONLY is enabled.

The Copr project test-dependencies uses the following repositories as runtime dependencies:
^^^^^^^^^^^^^^ this did not informed me that by enabling dturecek/test-dependencies the deps were enabled on MY machine too.

1. [coprdep:localhost:dturecek:foo]
   baseurl=http://backend_httpd:5002/results/dturecek/foo/fedora-$releasever-$basearch/
...
Do you want to keep them enabled? [y/N]:
```

Something like:
```
The Copr project X uses the following repositories as runtime dependencies:

1. XXXXX

These repositories have been automatically enabled.  <------- This is important

Do you want to keep them enabled? [Y/n]: <------- also note the flip in default, most people will want it enabled.
```

I will vote even for something like:
```
The Copr project X uses the following repositories as runtime dependencies:

1. XXXXX

These repositories have been automatically enabled.
```
I.e. no asking about enabling or disabling. I think most people will keep it enabled. The admin just wants to be informed of what is happening.

praiskup: > also note the flip in default

I think the default is correct (y) but we should swap the help string
`s/yN/Yn/`.  Now I understand why I was confused about the default before
....

> These repositories have been automatically enabled.
> I.e. no asking about enabling or disabling

Please no, that's what we have `dnf -y` for.  This feature brings another
level of trust.

You could as well claim that we should drop the first question because
all the people doing 'dnf copr enable ...' really want to have the project
enabled, but we still print the question because we need to have it
printed ...

Which makes me think that we should provide a warning that even the
dependencies are not maintained by Fedora Project.  This is IMO a must.

plugins/copr.py
lukash: Might be better to @m-blaha for this :slightly_smiling_face:

plugins/copr.py
praiskup: I mean, I like this variant, thanks!
```
These repositories have been automatically enabled.

Do you want to keep them enabled? [Y/n]:
```
But we still need to tell the people why they need them, and what are
the cons.

plugins/copr.py
```diff
@@ -520,13 +553,9 @@ def _disable_repo(self, copr_username, copr_projectname):
                 _("Failed to disable copr repo {}/{}"
                   .format(copr_username, copr_projectname)))
 
-        self.base.conf.write_raw_configfile(repo.repofile, repo.id,
-                                            self.base.conf.substitutions, {"enabled": "0"})
-
-        # disable also multilib counterpart, if found
-        multilib_repo_id = repo.id + ':ml'
-        if multilib_repo_id in self.base.repos:
-            self.base.conf.write_raw_configfile(repo.repofile, multilib_repo_id,
```
lukash, line 529: The multilib part here was removed... I guess this is intentional but seems unrelated to the commit, it probably should go into a separate commit with its own justification for removal?

lukash:
Can you _please_ add a description of what this actually does to the commit?

Seems to me this considers any sections in the same config file of a copr repository to be its dependencies? I might be wrong, I'd need to dig deep to find this out and probably test it. This information should come with the PR.

Also, it would be great if we could have a test for it. Someone from the team (yeah, I guess me :slightly_smiling_face: ) could help with that.

plugins/copr.py
lukash: (Looking at it some more it seems this is a hack that is being removed by a generic solution - worth putting into the commit message as well)

praiskup:
> Can you please add a description of what this actually does to the commit?

+1, but switch s/what/why/, I agree (answer to what is in the patch).

lukash:
> +1, but switch s/what/why/, I agree (answer to "what" is in the patch).

Both should be there. It can be hard to figure out what a patch is doing if you're not intimately familiar with the code and don't actually have the data it runs on (e.g. the `.repo` configs) in front of you.

I find it good manners to provide as much (succinct) information as you can when asking others for a review. There are multiple reviewers here and not all will go down to all the details, some might just want to know a TLDR; of what it does.

praiskup:
>  It can be hard to figure out what a patch is doing

In such case we probably miss some in-line comments.  What concretely
isn't clear?

lukash:
> In such case we probably miss some in-line comments.

Not necessarily... I'll just leave this here: https://chris.beams.io/posts/git-commit/ :slightly_smiling_face:

Let's not derail this further by discussing the whats and whys, I think that post describes it well enough and we can agree at least in general? I think this is important but don't want to spend my time bikeshedding.

praiskup:
> I think that post describes it well enough and we can agree at least in general?

Heh, yeah, that one is which I try to follow and wanted to cite it, too.
I now realized that it uses "what" keyword in "what/why/how" paragraph
(which underlines the "why" anyway).  I'd adjust the paragraph a bit, if
I could :-).  Anyway, we are talking about the same thing.

plugins/copr.py
```diff
@@ -479,6 +480,45 @@ def _download_repo(self, project_name, repo_filename, chroot=None):
...
+        self.base.reset(repos=True)
+        self.base.read_all_repos()
+
+        repo = self._get_copr_repo(self._sanitize_username(copr_username), copr_projectname)
+
+        runtime_deps = []
+        for repo_id in repo.cfg.sections():
+            if repo_id.startswith("copr:"):
+                continue
+            runtime_deps.append(repo_id)
```
lukash, line 493: @dturecek thanks for extending the commit description. Do I understand correctly the main repo you are talking about is the one that starts with "copr:" and then there may be other repository configurations in the config file and those are considered the dependencies?

Might be a good idea to sum this up in a comment here (e.g. for this line, or in a docstring for this function), to help clarify how this works.

dturecek: That is correct. I've added a docstring with a short description.

xsuchy:
I tested this and it works as expected.

The only issue I found is that when two enabled projects use msuchy/msuchy1 then you get a warning from DNF:
```
Repository copr:copr-fe-dev.cloud.fedoraproject.org:msuchy:test is listed more than once in the configuration
```
But that is just warning and we can work on better UX experience later. I do not want to block this PR on this.

xsuchy:
@rh-atomic-bot r+

lukash:
As mentioned, here are some tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/833

Two things regarding the new message (sorry for being late):
> ... from the main Copr repository (runtime dependencies provided)

Sounds weird, I'd change it to e.g.:
> ... from the main Copr repository (they provide runtime dependencies)

And then:
> Be aware that the note about quality and bug-reporting above applies here too, Fedora Project doesn't control the content.

I don't know what note above this is referring to, there is no such note in the output.

dturecek:
Thank you for the tests.

 > ... from the main Copr repository (runtime dependencies provided)

I agree that this could have been worded better.

> > Be aware that the note about quality and bug-reporting above applies here too, Fedora Project doesn't control the content.
> 
> I don't know what note above this is referring to, there is no such note in the output.

The note refers to [this warning](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/copr.py#L248).

lukash:
> The note refers to [this warning](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/copr.py#L248).

I see, so it only gets printed in an interactive terminal. Ideally these messages would be printed both (in an interactive terminal) or neither would be printed (non-interactive), as the other refers to the first :slightly_smiling_face:

praiskup:
We should print it both interactively and non-interactively, thanks for spotting that.
Should we use stdout (print), logging.info, or sys.stderr.write?

lukash:
> We should print it both interactively and non-interactively, thanks for spotting that.
> Should we use stdout (print), logging.info, or sys.stderr.write?

Good question. Certainly not `logging`, I'm not entirely sure whether these informative strings that are really not part of the regular output should go to `stdout` or `stderr`. @praiskup do you have an opinion on this? I think in this case the informative string is paired with the prompt and therefore should probably go to `stdout`, along with the `[y/N]` prompt (which is what `self.base.output.userconfirm` is doing). And therefore the newly added message should also go there? But then if you print these in non-interactive terminals, any script parsing the output will be getting all this. Should we print to `stdout` in interactive mode and to `stderr` in non-interactive? Or is that actually making it worse?

praiskup:
Dunno.  I need to start somewhere, so I picked `sys.stderr.write` (always),
according to your ideas above.  Let's move this discussion to #398.

PR #391, opened 2020-04-03
[needs-restarting] add opt using config file (RhBug:1810123)

inknos:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1810123


lukash:
@inknos could you add a short summary of what the new code does to the commit description? The Commit title itself also isn't very clear... Thanks!

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
 NEED_REBOOT = ['kernel', 'glibc', 'linux-firmware', 'systemd', 'dbus',
                'dbus-broker', 'dbus-daemon']
 
+class DirectoryOption(object):
+    """ This class handles a directory containing file options
+        This class can sanitize dir path
+    """
+    def __init__(self, path=None):
+        self._path = self._parse(path)
```
lukash, line 60: Since you're only calling `_parse` from here and it's also the only thing in `__init__()` the indirection seems unnecessary, you could put the parsing code here directly. Though see the whole review first, the whole class doesn't seem to be really needed.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+
+    def _parse(self, path):
+        if not isinstance(path, str):
+            return set(path)
+        strval = path.replace('\n', ' ')
+        strval = path.replace(',', ' ')
+        strval = { x for x in strval.split(' ') if x }
+        strval = { x for x in strval if self._is_dir(x) }
```
lukash, line 68: Overwriting this generically-named variable is not very nice. It's better to not store different things into the same variable. It can lead to errors such as overwriting the variable without actually using it's value in the first place :sunglasses: like in this case, on line 54 you don't call `strval.replace()`, so you lose the replacement of `\n` you've done on line 53.

How about this here:
```python
replaced = path.replace('\n', ' ').replace(',', ' ')
return { x for x in replaced.split() if self._is_dir(x) }
```
Note that `split()` without an argument splits on any whitespace and won't include empty strings in the result.

Again, see the whole review though.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+        return strval
+
+    def _is_dir(self, path=None):
+        return os.path.isdir(path)
+
+    #def _is_empty(self):
+    #    if len(os.listdir(self._path) ) == 0:
+    #        return False
+    #    return True
```
lukash, line 77: Remove this if not needed?

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+
+    @property
+    def path(self):
+        return self._path
```
lukash, line 81: Naming the variable `path` is actually quite confusing, since it's a set of paths.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+
+class FileOption(object):
+    """This class takes a list of files and sanitizes them
+    """
+    def __init__(self, path):
+        self._path = path
```
lukash, line 91: Again, the `path` here is a set of paths.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+    @property
+    def files(self):
+        return self._files
+
+class Option(object):
+    """ This class handles options inside .conf file
+    """
+    def __init__(self, path):
+        self._d_opt = DirectoryOption(path)
+        self._path = self._d_opt.path
```
lukash, line 133: The business code of the `DirectoryOption` class is roughly 3 LOCs :slightly_smiling_face: The rest is just boilerplate (glue). Please minimize the amount of boilerplate unless the infrastructure you're putting up is actually needed :slightly_smiling_face:

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+
+        self._f_opt = FileOption(self._path)
+        self._files = self._f_opt.files
```
lukash, line 136: While the `FileOption` class actually does more things and you've structured it quite nicely, ultimately it appears to be pretty linear and just list the files in a directory and do some filtering on them. I'd prefer to cut the boilerplate again and simplify this as much as possible...

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+
+        self._values = self._read_values()
+
+    def _read_values(self):
+        values = set()
+        for filepath in self._files:
+            with open(filepath) as fp:
+                line = fp.readline()
+                while line:
```
lukash, line 145: You can just do:
```python
for line in fp:
    # ...
```

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+                    if self._is_single_word(line):
```
lukash, line 146: This check seems a bit arbitrary (checking whether it contains a space). I'm not entirely sure if we should validate the package name and how, or whether to not do anything perhaps...

One way to validate the name somewhat using the NEVRA "parser" we have, just with the name, could be:
```python
subj = hawkey.Subject(line)
names = subj.get_nevra_possibilities(forms=[hawkey.FORM_NAME])
```
And `names` should be an array of one item if it was parsed correctly...

plugins/needs_restarting.py
```diff
@@ -40,6 +40,113 @@
...
+                        values.add(line)
+                    line = fp.readline()
+        return self._sanitize_values(values)
+
+    def _is_single_word(self, line):
+        return not (' ' in line)
+
+    def _sanitize_values(self, values):
+        values = { x.replace('\n', '') for x in values }
+        return { x for x in values if x }
```
lukash, line 156: I don't think you can have `'\n'` in `values` anymore, since you've iterated over the lines of the file, so it was split on any linebreaks. Also these set comprehensions are more complicated than necessary :slightly_smiling_face: IMO you can just remove this function.

inknos:
I did a lot of rework and now a single function is called.
Now the function accepts only one path or a list of paths. The other calls are almost the same but the code should be cleaner and more straightforward.
I also added a warning check to be sure that the package in added to needs-restarting list is installed.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,50 @@
...
+    # find options
+    for file in sanitized:
+        with open(file) as fp:
+            for line in fp:
+                options.add((line.rstrip(), file))
+
+    # sanitize packages
+    for pkg in base.sack.query().installed().filter(name={x[0] for x in options}):
+        packages.add(pkg.name)
+    if options != packages:
```
lukash, line 81: I think this check is not necessary, the `for` loop right below will just not produce anything if they are equal?

plugins/needs_restarting.py
```diff
@@ -40,6 +40,50 @@
...
+    # find files
+    for dir in replaced:
+        for file in os.listdir(dir):
+            configfiles.add(os.path.join(dir,file))
+
+    # sanitize files
+    sanitized = set()
+    for file in configfiles:
+        if file.endswith('.conf'): sanitized.add(file)
+        if not os.path.isdir(file): sanitized.add(file)
```
lukash, line 70: Unless I'm missing something, this is wrong, it adds the file if either of the above is true, not if both of them are?

I think this single-line if syntax is at least discouraged, if not against PEP8? Maybe this is why? :slightly_smiling_face:

plugins/needs_restarting.py
```diff
@@ -40,6 +40,50 @@
 NEED_REBOOT = ['kernel', 'glibc', 'linux-firmware', 'systemd', 'dbus',
                'dbus-broker', 'dbus-daemon']
 
+def get_options_from_dir(filepath):
+    """
+    Provide filepath as string if single dir or list of strings
+    Return set of package names contained in files under filepath
+    """
+    base = dnf.Base()
```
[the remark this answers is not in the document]
inknos: did not notice, thanks

plugins/needs_restarting.py
```diff
@@ -40,6 +40,50 @@
...
+    base.fill_sack()
+    packages = set()
+    replaced = set()
+    configfiles = set()
+    options = set()
+
+    # find dirs
+    if isinstance(filepath, str):
+        replaced = { filepath } if os.path.isdir(filepath) else set()
```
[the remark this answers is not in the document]
inknos: you could call it by `get_options_from_dir(["path1", "path2"])`

I can remove it if you want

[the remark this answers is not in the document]
lukash: Yes, but it is not used atm. and it's questionable whether it ever will be needed (I'd say likely it won't). I'd keep it simple for now, it can be extended very easily if it ever is needed.

plugins/needs_restarting.py
```diff
@@ -40,6 +40,30 @@
...
+            for line in fp:
+                options.add((line.rstrip(), file))
+
+    packages = set()
+    for pkg in base.sack.query().installed().filter(name={x[0] for x in options}):
+        packages.add(pkg.name)
+    for name, file in {x for x in options if x[0] not in packages }:
+        logger.warning(
+            _('No installed package found for package name "{pkg}"'
+            'specified in needs-restarting file "{file}".'.format(pkg=name, file=file)))
```
lukash, line 64: Please indent this line by 4 spaces, this certainly doesn't adhere to PEP8.

lukash:
@rh-atomic-bot r+

PR #395, opened 2020-05-05
Add option -s (RhBug:1772939)

inknos:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1772939

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -29,6 +29,7 @@ from dnfpluginscore import logger, _
 
 import dnf
 import dnf.cli
+import dbus
 import functools
 import os
 import re
@@ -126,6 +127,30 @@ def print_cmd(pid):
     print('%d : %s' % (pid, command))
 
 
+def get_service_dbus(pid):
+    bus = dbus.SystemBus()
+    systemd_manager_object = bus.get_object(
+        'org.freedesktop.systemd1',
+        '/org/freedesktop/systemd1'
+    )
+    systemd_manager_interface = dbus.Interface(
+        systemd_manager_object,
+        'org.freedesktop.systemd1.Manager'
+    )
+    service_proxy = bus.get_object(
+        'org.freedesktop.systemd1',
+        systemd_manager_interface.GetUnitByPID(pid)
+    )
+    service_properties = dbus.Interface(
+        service_proxy, dbus_interface="org.freedesktop.DBus.Properties")
+    name = service_properties.Get(
+        "org.freedesktop.systemd1.Unit",
+        'Id'
+    )
+    if name.endswith(".service"):
+        return name
+    return
+
 def smap2opened_file(pid, line):
     slash = line.find('/')
     if slash < 0:
@@ -205,6 +230,8 @@ class NeedsRestartingCommand(dnf.cli.Command):
         parser.add_argument('-r', '--reboothint', action='store_true',
                             help=_("only report whether a reboot is required "
                                    "(exit code 1) or not (exit code 0)"))
+        parser.add_argument('-s', '--services', action='store_true',
+                            help=_("only report affected systemd services"))
 
     def configure(self):
         demands = self.cli.demands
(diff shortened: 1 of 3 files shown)


plugins/needs_restarting.py
```diff
@@ -100,6 +101,18 @@ def print_cmd(pid):
...
+def get_service(pid):
+    cmdline = 'ps -o unit= %d' % pid
+    process = subprocess.Popen(['ps', '-o', 'unit=', str(pid)],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+    stdout, stderr = process.communicate()
+    servicename = str(stdout, 'utf-8').rstrip()
+    if servicename.endswith(".service"):
+        return servicename
+    return
```
lukash, line 113: I think some error handling should be in place here? At least check the exit status for the `ps` call and possibly also if the service doesn't match? Not sure what can possibly happen here and how it should be handled but I think it should be handled somehow :slightly_smiling_face:

kontura:
What is the status here? If I recall there was some separate test for this?

Also I might have missed this, but have you considered using the same approach as the [yum-utils needs-restarting](https://github.com/rpm-software-management/yum-utils/commit/b7271d67c1e425fa5494e1642273d6d54a8372b6)? It doesn't depend on ps with systemd support or does that no longer work?

lukash:
> What is the status here? If I recall there was some separate test for this?

I think @lucab85 is writing a test for it? I'm not sure if I'm supposed to be reviewing this though :slightly_smiling_face: 

> Also I might have missed this, but have you considered using the same approach as the [yum-utils needs-restarting](https://github.com/rpm-software-management/yum-utils/commit/b7271d67c1e425fa5494e1642273d6d54a8372b6)? It doesn't depend on ps with systemd support or does that no longer work?

That actually looks quite a bit better than calling the `ps` binary. In theory it would seem it depends on `cgroup` support in the kernel but I'm sure we can depend on that for all RPM systems anyway.

kontura:
I have tried testing this on my system and it doesn't work. :cry: 
Since fedora 31 cgroups v2 is used by default and it uses a different /proc/*/cgroup format. In this [doc](https://www.kernel.org/doc/Documentation/cgroup-v2.txt) in the section `Processes` it specifies: 
> The entry for cgroup v2 is always in the format "0::$PATH"::

The $PATH for services seem to end with `.service` but I don't know if we can rely on that, we could try asking someone from systemd if they know or if they know about some better way to do this. Otherwise the only other option I see is going back to `ps` I guess.

Also @lucab85 I have found we already have a test for this behavior, its a yum-utils test: `needs-restarting-show-service-names`.

plugins/needs_restarting.py
```diff
@@ -126,6 +127,28 @@ def print_cmd(pid):
...
+        'org.freedesktop.systemd1',
+        systemd_manager_interface.GetUnitByPID(pid)
+    )
+    service_properties = dbus.Interface(
+        service_proxy, dbus_interface="org.freedesktop.DBus.Properties")
+    name = service_properties.Get(
+        "org.freedesktop.systemd1.Unit",
+        'Id'
+    )
+    return name
```
kontura, line 150: I believe this will now return all unit types not just services for given pid (in my local test it also printed `session-2.scope`).

We probably want to check if the suffix of the `name` is ".service", I think this should be safe because the man page for `systemd.unit` specifies: 
> The type suffix must be one of ".service", ".socket", ".device", ".mount", ".automount", ".swap", ".target", ".path", ".timer", ".slice", or ".scope".

kontura:
Can you use the = changelog = format with this new commit? It makes it easier for releasing later.

Otherwise LGTM.

kontura:
@rh-atomic-bot r+

kontura:
@rh-atomic-bot retry

lukash:
It's missing the dbus python module dependency which was added here:
https://copr-be.cloud.fedoraproject.org/results/rpmsoftwaremanagement/rpm-gitoverlay-1600415817.434388/fedora-31-x86_64/01672833-dnf-plugins-core/build.log.gz

kontura:
@rh-atomic-bot r+

PR #396, opened 2020-05-06
Move args --set-enabled/--set-disabled, add messages (RhBug:1727882)

jrohel:
The arguments `--set-enabled`, `--set-disabled` are valid for `config-manager` plugin. 
The fix moves they from general DNF arguments into the plugin arguments.
Note:
There are another "config-manager" arguments `--enable` and `--disable` in the DNF general section. They cannot be moved into config-manager plugin because of Python argparser abbreviation.

Added warning when `--enablerepo`/`--disablerepo` arguments were passed. These global DNF arguments have no meaning with confim-manager plugin. Users are confused and use them instead of --set-enabled/--set-disabled. But it does not work.

Depends on: https://github.com/rpm-software-management/dnf/pull/1621

### DIFF
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -81,6 +81,11 @@ class ConfigManagerCommand(dnf.cli.Command):
                                          "--set-enabled", "--enable",
                                          "--set-disabled", "--disable"])))
 
+        # warn with hint if --enablerepo or --disablerepo argument was passed
+        if self.opts.repos_ed != []:
+            logger.warning(_("Warning: --enablerepo/--disablerepo arguments have no meaning"
+                             "with config manager. Use --set-enabled/--set-disabled instead."))
+
         if (self.opts.save or self.opts.set_enabled or
                 self.opts.set_disabled or self.opts.add_repo):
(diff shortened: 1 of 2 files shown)


akostadinov:
Looks nice, thank you!

plugins/config_manager.py
```diff
@@ -71,7 +78,13 @@ def configure(self):
                                      .format(' '.join([
                                          "--save", "--add-repo",
                                          "--dump", "--dump-variables",
-                                         "--enable", "--disable"])))
+                                         "--set-enabled", "--enable",
+                                         "--set-disabled", "--disable"])))
```
m-blaha, line 82: You need to modify the tests in plugins-core/config-manager.feature to cope with this modified stderr message:
```Command line error: one of the following arguments is required: --save --add-repo --dump --dump-variables --set-enabled --enable --set-disabled --disable```
instead of original
```Command line error: one of the following arguments is required: --save --add-repo --dump --dump-variables --enable --disable```

m-blaha:
Tested locally.

PR #397, opened 2020-05-11
[config-manager] enable/disable comma separated pkgs (RhBug:1830530)

inknos:
test: https://github.com/rpm-software-management/ci-dnf-stack/pull/828
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1830530

### DIFF
--- a/doc/config_manager.rst
+++ b/doc/config_manager.rst
@@ -86,6 +86,12 @@ Examples
 ``dnf config-manager --set-enabled <repoid>``
     Enable repository identified by <repoid> and make the change permanent.
 
+``dnf config-manager --set-disabled <repoid1> <repoid2>``
+    Disable repositories identified by <repoid1> and <repoid2>
+
+``dnf config-manager --set-disabled <repoid1>,<repoid2>``
+    Disable repositories identified by <repoid1> and <repoid2>
+
 ``dnf config-manager --save --setopt=*.proxy=http://proxy.example.com:3128/ <repo1> <repo2>``
     Update proxy setting in repositories with repoid <repo1> and <repo2> and make the change
     permanent.
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -90,6 +90,11 @@ class ConfigManagerCommand(dnf.cli.Command):
                 self.opts.set_disabled or self.opts.add_repo):
             demands.root_user = True
 
+        # sanitize commas https://bugzilla.redhat.com/show_bug.cgi?id=1830530
+        temp_list = [x.split(',') for x in self.opts.crepo if x != ',']
+        # flatten sublists
+        self.opts.crepo = [item for sublist in temp_list
+            for item in sublist if item != '']
 
     def run(self):
         """Execute the util action here."""


plugins/config_manager.py
```diff
@@ -77,6 +77,8 @@ def configure(self):
                 self.opts.set_disabled or self.opts.add_repo):
             demands.root_user = True
 
+        # sanitize commas https://bugzilla.redhat.com/show_bug.cgi?id=1830530
+        self.opts.crepo = [x.replace(',', '') for x in self.opts.crepo if x is not ',']
```
pkratoch, line 81: This actually does `repo1,repo2 -> repo1repo2` which doesn't exist. So, it works for `repo1, repo2`, but not for `repo1,repo2`.

pkratoch:
I am not sure we should even do this, because it's just not the way that python arguments work. Couldn't we instead improve documentation, so that it's clear that `dnf config-manager --enable repo1 repo2` can be used? And even better if we put the `--enable` at the end, so that it's clear that the repos are not arguments of `--enable` but are positional arguments. What do you think?

pkratoch:
Btw, I just found out that jrohel already moved the `--set-enabled/--set-disabled` arguments from dnf options to the plugin options in https://github.com/rpm-software-management/dnf-plugins-core/commit/aa68255138246c2e33f6fe8406d113988e0a5a39 . So that should help with clarity, too, because now users see this option in help right away, before they find `--disablerepo/--enablerepo`.

plugins/config_manager.py
inknos: I changed the way the commas are parsed. Now I hope it should be more clear
Wildcards are also supported as we discussed

plugins/config_manager.py
```diff
@@ -90,6 +90,11 @@ def configure(self):
                 self.opts.set_disabled or self.opts.add_repo):
             demands.root_user = True
 
+        # sanitize commas https://bugzilla.redhat.com/show_bug.cgi?id=1830530
+        temp_list = [x.split(',') for x in self.opts.crepo if x is not ',']
+        # flatten sublists
+        self.opts.crepo = [item for sublist in temp_list
+            for item in sublist if item is not '']
```
pkratoch, line 97: Use rather "!=" instead of "is not", because you want to check if the strings are equal, not if the objects are the same. The result is the same here, but not in general.

doc/config_manager.rst
```diff
@@ -86,6 +86,12 @@ Examples
 ``dnf config-manager --set-enabled <repoid>``
     Enable repository identified by <repoid> and make the change permanent.
 
+``dnf config-manager --disable <repoid1> <repoid2>``
+    Disable repositories identified by <repoid1> and <repoid2>
+
+``dnf config-manager --disable <repoid1>,<repoid2>``
+    Disable repositories identified by <repoid1> and <repoid2>
```
pkratoch, line 93: This is maybe just a matter of preference, but can you rather use ``--set-disabled``? We added the ``--enable`` and ``--disable`` options only for compatibility reasons and these options are not listed in the config-manager help section (only in the general section). Also, one of the other examples above uses the ``--set-*`` variant and it would be better to stay consistent and promote just one of the variants.

pkratoch:
@rh-atomic-bot r+

PR #398, opened 2020-05-18
Always print no liability text

praiskup:


plugins/copr.py
```diff
@@ -502,22 +506,29 @@ def _runtime_deps_warning(self, copr_username, copr_projectname):
...
-            print("   baseurl={0}".format(repo.cfg.getValue(repo_id, "baseurl")))
-            dep_idx += 1
-
-        msg = _("\nThese repositories have been enabled automatically."
-                "\nDo you want to keep them enabled?")
+        msg = _(
+            "\nMaintainer of the enabled Copr repository decided to make\n"
+            "it dependent on other repositories. Such repositories are\n"
+            "usually necessary for successful installation of RPMs from\n"
+            "the main Copr repository (runtime dependencies provided).\n\n"
```
lukash, line 512: Can you fix the text here while at it? To something like:
> ... (they provide runtime dependencies).

plugins/copr.py
```diff
@@ -502,22 +506,29 @@ def _runtime_deps_warning(self, copr_username, copr_projectname):
...
+        )
+
+        counter = itertools.count(1)
+        sys.stderr.write(msg.format("\n\n".join([
+            "{num:2}. [{repoid}]\n    baseurl={baseurl}".format(
+                num=next(counter),
+                repoid=repoid,
+                baseurl=repo.cfg.getValue(repoid, "baseurl"))
+            for repoid in runtime_deps
+        ])))
```
lukash, line 528: I'm not sure I find this more readable myself but I guess it's subjective :sunglasses:

plugins/copr.py
```diff
@@ -502,22 +506,29 @@ def _runtime_deps_warning(self, copr_username, copr_projectname):
...
+
+        msg = _("Do you want to keep them enabled?")
```
lukash, line 530: The issue I find here is that the information above is on `stderr` and the prompt here is on `stdout`, so they can end up being separated and the prompt doesn't make sense without the information.

In case I wasn't clear in my previous comment (https://github.com/rpm-software-management/dnf-plugins-core/pull/388#issuecomment-630022872), I meant to either print everything to `stdout` (interactive mode) or everything to `stderr` (non-interactive). But I'm not too happy about that either, that makes it inconsistent in a different way.

praiskup: > The issue I find here is that the information above is on stderr and the
> prompt here is on stdout, so they can end up being separated and the
> prompt doesn't make sense without the information.

How likely is that you redirect stderr, and still want to work
interactively?  I don't think we want to spent too much time on such
details...  which are kind of precedent I think.  I think we should
make a compromise fix here, and wait till there's proper tooling
in dnf for such warnings.

As you pointed out, some software in the wild might already read the
stdout...  and since we used logger.info before, I think using just stdout
in the #388 was mistake.  I think there are two options, either use logger
(my slight preference, since nobody complaint so far and we were just
fine) or switch to sys.stderr.write.  Please tell me your preference.

plugins/copr.py
praiskup: accepted

plugins/copr.py
praiskup: Meh, `logger.info` goes to stdout, interesting..  Which means that IMO we
should use sys.stderr ... eventually.

praiskup: Still ... if you have preference, I don't see a problem neither way.

lukash: > ... and wait till there's proper tooling in dnf for such warnings.

Not wanting to derail much here, but what would be the proper tooling? I think we are talking conceptually here, you print to one or the other, but it always seems to be a compromise. At least I'm not sure what a clean solution is.

I've come to the conclusion (at least for the moment, not saying I can't be convinced) that using logger for console output is a bad idea. Things you want to go to the logger and to console often differ, what goes to console you don't want to go to the logger and vice versa. (that said outputting logger lines to stderr with e.g.`--debug` is a fine)

Yes, logger.info going to `stdout` is bad.

I'm also thinking whether we should print the information text to `stderr` and actually list the repositories to `stdout`?

However, if you just want to be done with this, we can keep the current version, i.e. printing it to `stderr` (except for the "[y/N]" prompt).

praiskup: > what would be the proper tooling? 

I mean some dnf api like `self.print_user_message(...)` which will do exactly
what DNF people expect it to do (the point is to abstract this stdout/stderr
problem out so plugin contributors don't have to care).

> I'm also thinking whether we should print the information text to stderr and actually list the repositories to stdout?

I don't know.  Feel free to submit PR if you want to make it even nicer.

> However, if you just want to be done with this, we can keep the current version, i.e. printing it to stderr (except for the "[y/N]" prompt).

That's what I prefer, yes.  I don't think anything else is better :-/

plugins/copr.py
```diff
@@ -256,8 +260,8 @@ def run(self):
 Please do not file bug reports about these packages in Fedora
 Bugzilla. In case of problems, contact the owner of this repository.
 
-Do you really want to enable {0}?""".format('/'.join([self.copr_hostname,
-                                                      copr_username, copr_projectname])))
+""").format('/'.join([self.copr_hostname, copr_username, copr_projectname])))
```
lukash, line 262: You don't need to format this string anymore.

Also the string has two line breaks at the end, and the potential dependencies followup starts with a line break. This results in two empty lines between the two, I think one empty line would be better. I'd also rather not have leading/trailing line breaks in the strings, as it could confuse the translators and they could omit some of them by accident? What about printing the line breaks separately?

plugins/copr.py
```diff
@@ -256,8 +260,8 @@ def run(self):
...
+            msg = "Do you really want to enable {0}?".format(self.copr_hostname)
```
lukash, line 263: This new prompt only prints the `copr_hostname`, the previous one had the whole `hostname/username/projectname` identifier. Note this prompt is not tested by the tests...

plugins/copr.py
lukash: > I mean some dnf api like `self.print_user_message(...)` which will do exactly
> what DNF people expect it to do (the point is to abstract this stdout/stderr
> problem out so plugin contributors don't have to care).

Good point.

> > However, if you just want to be done with this, we can keep the current version, i.e. printing it to stderr (except for the "[y/N]" prompt).
> 
> That's what I prefer, yes. I don't think anything else so far discussed is better :-/

Okay.

plugins/copr.py
praiskup: > You don't need to format this string anymore.

Good catch.

> Also the string has two line breaks at the end, and the potential dependencies
> followup starts with a line break. This results in two empty lines between the
> two, I think one empty line would be better. I'd also rather not have
> leading/trailing line breaks in the strings, as it could confuse the
> translators and they could omit some of them by accident? What about printing
> the line breaks separately?

Yeah, good idea ... done in separate wrapping function now, by calling
msg.strip() first.

plugins/copr.py
praiskup: Fixed.

praiskup:
Thank you for the review @lukash, PTAL.

plugins/copr.py
```diff
@@ -502,29 +515,34 @@ def _runtime_deps_warning(self, copr_username, copr_projectname):
...
-        for repo_id in runtime_deps:
-            print("\n{0}. [{1}]".format(dep_idx, repo_id))
-            print("   baseurl={0}".format(repo.cfg.getValue(repo_id, "baseurl")))
-            dep_idx += 1
-
-        msg = _("\nThese repositories have been enabled automatically."
-                "\nDo you want to keep them enabled?")
-        if self.base._promptWanted():
-            if self.base.conf.assumeno or not self.base.output.userconfirm(
-                    msg='{} [Y/n]: '.format(msg), defaultyes_msg='{} [Y/n]: '.format(msg)):
```
lukash, line 523: This code here has a capital Y: `"[Y/n]"` in both `msg` and `defaultyes_msg` and your new code unifies it to `"[y/N]"` for `msg`, but that is you actually fixing a mistake, unless I'm missing something, right?

I just don't understand why the `userconfirm()` method doesn't abstract away the whole `"[y/N]"` prompt part...

plugins/copr.py
```diff
@@ -188,6 +192,9 @@ def _read_config_item(self, config, hub, section, default):
         except (NoOptionError, NoSectionError):
             return default
 
+    def _user_warning_before_prompt(self, text):
+        sys.stderr.write("\n{0}\n\n".format(text.strip()))
```
lukash, line 195: So the leading `"/n"` makes it print an empty line at the beginning of the first block of text and also results in there being two line-breaks between the first block of text and the potential second one (the dependency warning). I would remove it, making it not print an empty line at the beginning and resulting in a single empty line between the two blocks of text, which IMO looks better?

plugins/copr.py
```diff
@@ -255,9 +262,10 @@ def run(self):
 
 Please do not file bug reports about these packages in Fedora
 Bugzilla. In case of problems, contact the owner of this repository.
-
-Do you really want to enable {0}?""".format('/'.join([self.copr_hostname,
-                                                      copr_username, copr_projectname])))
+"""))
+            project = '/'.join([self.copr_hostname, copr_username,
+                                copr_projectname])
```
lukash, line 266: Nitpick: this could go on a single line.

praiskup: Not within 80 characters.

plugins/copr.py
praiskup: Correct.  I initially wanted to have n/Y by default in the second warning,
but then I was over-voted in previous PR... we changed the semantics,
but not the message.

praiskup: Well, no ... we don't have powers to decide the default in plugin, right?  It depends on the system configuration, and on my box it is e.g. always `y/N`.  Is that correct?

plugins/copr.py
lukash: Our limit is 100.

plugins/copr.py
lukash: > Well, no ... we don't have powers to decide the default in plugin, right? It depends on the system configuration, and on my box it is e.g. always `y/N`. Is that correct?

I understood it the same way.

plugins/copr.py
praiskup: Ok, I'm a bit distracted by `Updating Subscription Management repositories.` _stdout_ text :-( which made me think the initial space is needed.  Good catch.

plugins/copr.py
praiskup: Is it wrong to wrap it if I consider limit 100 to be a wrong limit?

plugins/copr.py
lukash: Oh, I see. I'm testing on Fedora :slightly_smiling_face:

plugins/copr.py
lukash: Hard to believe, you really like 80? You can keep it in the patch, I don't mind too much. Though it would be better if you adhered to the project's coding style I think :sunglasses:

plugins/copr.py
praiskup: Me too.

lukash:
Thanks, merging.

PR #402, opened 2020-06-17
Fix debug-restore command

m-blaha:
- correctly work with install-only packages (BZ#1844533)
- do not remove current versions of packages that are supposed to be
  replaced (downgraded / upgraded)


kontura:
@rh-atomic-bot r+

PR #406, opened 2020-07-15
return error when dnf download failed

zhangrui182:
fix RhBug:1844925: https://bugzilla.redhat.com/show_bug.cgi?id=1844925


pkratoch:
@rh-atomic-bot r+

PR #407, opened 2020-07-23
[reposync] Add latest NEVRAs per stream to download (RhBug: 1833074)

m-blaha:
This covers situation when package with the newest NEVRA is part of
an older version of a stream.

https://bugzilla.redhat.com/show_bug.cgi?id=1833074

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/875

### DIFF
--- a/plugins/reposync.py
+++ b/plugins/reposync.py
@@ -207,27 +207,60 @@ class RepoSyncCommand(dnf.cli.Command):
 
     def _get_latest(self, query):
         """
-        return query with latest nonmodular package and all packages from latest version per stream
+        return union of these queries:
+        - the latest NEVRAs from non-modular packages
+        - all packages from stream version with the latest package NEVRA
+          (this should not be needed but the latest package NEVRAs might be
+          part of an older module version)
+        - all packages from the latest stream version
         """
         if not dnf.base.WITH_MODULES:
             return query.latest()
+
         query.apply()
         module_packages = self.base._moduleContainer.getModulePackages()
         all_artifacts = set()
         module_dict = {}  # {NameStream: {Version: [modules]}}
+        artifact_version = {} # {artifact: {NameStream: [Version]}}
         for module_package in module_packages:
-            all_artifacts.update(module_package.getArtifacts())
+            artifacts = module_package.getArtifacts()
+            all_artifacts.update(artifacts)
             module_dict.setdefault(module_package.getNameStream(), {}).setdefault(
                 module_package.getVersionNum(), []).append(module_package)
-        non_modular_latest = query.filter(
+            for artifact in artifacts:
+                artifact_version.setdefault(artifact, {}).setdefault(
+                    module_package.getNameStream(), []).append(module_package.getVersionNum())
+
+        # the latest NEVRAs from non-modular packages
+        latest_query = query.filter(
             pkg__neq=query.filter(nevra_strict=all_artifacts)).latest()
-        latest_artifacts = set()
-        for version_dict in module_dict.values():
-            keys = sorted(version_dict.keys(), reverse=True)
-            for module in version_dict[keys[0]]:
-                latest_artifacts.update(module.getArtifacts())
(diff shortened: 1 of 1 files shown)


plugins/reposync.py
```diff
@@ -207,27 +207,60 @@ def download_metadata(self, repo):
...
+            # add the newest stream version
+            versions.add(sorted(version_dict.keys(), reverse=True)[0])
+            # collect all artifacts in all stream versions
+            stream_artifacts = set()
+            for modules in version_dict.values():
+                for module in modules:
+                    stream_artifacts.update(module.getArtifacts())
+            # find versions to which the packages with the highest NEVRAs belong
+            for lp in query.filter(nevra_strict=stream_artifacts).latest():
+                #XXX contains modules.yaml allways full NEVRA (including epoch?)
```
j-mracek, line 253: I am sorry but I do not understand xxx note.

plugins/reposync.py
```diff
@@ -207,27 +207,60 @@ def download_metadata(self, repo):
...
+                nevra = "{0.name}-{0.epoch}:{0.version}-{0.release}.{0.arch}".format(lp)
+                versions.update(artifact_version[nevra][namestream])
```
j-mracek, line 255: I am not 100 % sure, but what happen in case when the latest artefact is present in all or multiple versions? Somehow I feel that in this case all versions will be downloaded. I suggest that in this case only the latest context should be used.

plugins/reposync.py
m-blaha: OK, the thing is that I need to pair package objects to artefacts as they are written in modules.yaml (artifacts are keys of the `artifact_version` dictionary). Currently the code relies on having artifacts in form of full NEVRA. I was just wondering if there is possibility that artifacts in modules.yaml are in a different format (e.g. zero epoch being left out).

plugins/reposync.py
m-blaha: Yes, currently all versions with the latest artefact versions are going to be downloaded. I'll try to limit it to the latest version in this case.

plugins/reposync.py
```diff
@@ -207,27 +207,61 @@ def download_metadata(self, repo):
...
+            versions = set()
+            # add the newest stream version
+            versions.add(sorted(version_dict.keys(), reverse=True)[0])
+            # collect all artifacts in all stream versions
+            stream_artifacts = set()
+            for modules in version_dict.values():
+                for module in modules:
+                    stream_artifacts.update(module.getArtifacts())
+            # find versions to which the packages with the highest NEVRAs belong
+            for lp in query.filter(nevra_strict=stream_artifacts).latest():
```
j-mracek, line 252: What about to rename lp to `pkg` or `latest_pkg`?

plugins/reposync.py
```diff
@@ -207,27 +207,61 @@ def download_metadata(self, repo):
...
+                #XXX contains modules.yaml allways full NEVRA (including epoch?)
+                nevra = "{0.name}-{0.epoch}:{0.version}-{0.release}.{0.arch}".format(lp)
+                # download only highest version containing the latest artifact
+                versions.add(max(artifact_version[nevra][namestream]))
+            # add all artifacts from selected versions for synchronization
+            for version, modules in version_dict.items():
+                if version in versions:
```
j-mracek, line 259: This not a problem, but anyway, the code could use a different logic.
```
for version in versions:
    for module in version_dict[version]:
```

j-mracek:
The logic and code LGTM!

m-blaha:
Thanks Jardo, I squashed all the commits into one (they were just unnecessary development noise) and fixed issues you found in  it.

j-mracek:
@rh-atomic-bot try

j-mracek:
@rh-atomic-bot r+

m-blaha:
I suspect that `@rh-atomic-bot r+` does not work correctly after `@rh-atomic-bot try`. Merging manually.

PR #410, opened 2020-08-23
copr: don't try to list runtime dependencies

praiskup:
These are not necessarily other copr repositories, but rather any
(external) repositories.  We can not disable them "separately" by
'dnf copr disable', etc.  Printing them in 'dnf copr list' isn't
semantically correct.

Fixes: rhbz#1871373


praiskup:
@xsuchy, please take a look

xsuchy (approved):

PR #411, opened 2020-10-23
[reposync] Check GPG signatures of downloaded packages (RhBug:1856818) 

m-blaha:
YUMv3 reposync used to have --gpgcheck option to remove packages that fail GPG signature checking after downloading.
This patch implements the option for DNF.

https://bugzilla.redhat.com/show_bug.cgi?id=1856818

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/900

### DIFF
--- a/doc/reposync.rst
+++ b/doc/reposync.rst
@@ -39,35 +39,35 @@ Options
 
 All general DNF options are accepted. Namely, the ``--repoid`` option can be used to specify the repositories to synchronize. See `Options` in :manpage:`dnf(8)` for details.
 
-``-p <download-path>, --download-path=<download-path>``
-    Root path under which the downloaded repositories are stored, relative to the current working directory. Defaults to the current working directory. Every downloaded repository has a subdirectory named after its ID under this path.
-    
-``--norepopath``
-    Don't add the reponame to the download path. Can only be used when syncing a single repository (default is to add the reponame).
-
-``--download-metadata``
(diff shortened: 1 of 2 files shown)


doc/reposync.rst
```diff
@@ -48,6 +48,9 @@ All general DNF options are accepted. Namely, the ``--repoid`` option can be use
 ``--download-metadata``
     Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c on it.
 
+``-g, --gpgcheck``
+    Remove packages that fail GPG signature checking after downloading. Exit code is ``1`` if at least one package was removed.
```
kontura, line 52: I suggest we should mention here that if a specific repo has `gpgcheck=0` its packages are not checked even with this option.

Or perhaps this option should override the `gpgcheck` config value and check even packages from repos with `gpgcheck=0`?

m-blaha: May be the cleanest solution would be to do the gpgcheck by default and skip it only when `--nogpgcheck` option was used. In that case `--gpgcheck` option would not be needed at all. But I do not want to change the reposync behaviour during RHEL 8 cycle.
So I will go with clarifying this in documentation.

kontura:
@rh-atomic-bot r+

kontura:
The tests pass locally, LGTM.

PR #412, opened 2020-11-05
groups-manager plugin (RhBug:1826016)

m-blaha:
Re-introduces functionality of yum-groups-manager

https://bugzilla.redhat.com/show_bug.cgi?id=1826016

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/906


plugins/groups_manager.py
```diff
@@ -0,0 +1,95 @@
...
+            try:
+                # xml_f returns a list of errors / log entries
+                errors = self.comps.xml_f(file_name)
+            except libcomps.XMLGenError as err:
+                errors = [err]
+            if errors:
+                logger.warning(''.join(errors).strip())
+
+    def run(self):
+        self.load_input_files()
```
lukash, line 283: The return value from `load_input_files()` is not used, so I wonder what happens on a failure...

Also, shouldn't the command fail on any error in any file? This way it doesn't seem robust at all.

plugins/groups_manager.py
```diff
@@ -21,14 +21,54 @@
...
+
+def text_to_id(text):
+    '''generate group id based on its name'''
+    group_id = text.lower()
+    group_id = re.sub('[^{}]'.format(RE_GROUP_ID_VALID), '', group_id)
+    if not group_id:
+        raise dnf.cli.CliError(
+            _("Can't generate group id from '{}'. Please specify group id using --id.").format(
+                text))
+    return group_id
```
lukash: I know I'm complicating things (so feel free to push back - e.g. regexps will likely be a problem...), but I think this kind of validation belongs to libcomps to be possibly shared among projects.

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+Description
+-----------
+groups-manager plugin is used to create or edit a group metadata file for a repository. This is often much easier than writing/editing the XML by hand. The groups-manager can load an entire file of groups metadata and either create a new group or edit an existing group and then write all of the groups metadata back out.
+
+---------
+Arguments
+---------
+
+``<package-name-spec>``
+    Packages to add/remove to group.
```
lukash, line 40: > Packages to add to or remove from a group.

?

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+
+-------
+Options
+-------
+
+All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for details.
+
+``--load=<path_to_comps.xml>``
+    Load the groups metadata information from the specified file, before performing any operations. Metadata from all files are merged together if the option is specified multiple times.
```
lukash, line 49: I'd drop the comma:
> ... from the specified file before performing ...

Nitpick! :astonished:

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+    The id to lookup/use for the group. If you don't specify an ``<id>``, but do specify a name that doesn't refer to an existing group, then an id for the group is generated based on the name.
+
+``-n <name>, --name=<name>``
+    The name to lookup/use for the group. If you specify an existing group id, then the group with that id will have it's name changed to this value.
+
+``--description=<description>``
+    The description to use for the group.
+
+``--display-order=<display_order>``
+    Change the integer which controls the order groups are presented in, for example in dnf grouplist.
```
lukash, line 70: `dnf grouplist` is a command right? Seems worth to put it in backticks...

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+
+``--translated-name=<lang:text>``
+    This is another name for the group within the locale given. The syntax is lang:text. Eg. en:my-group-name-in-english
```
lukash, line 73: Would this be a clearer formulation of the first sentence?
> A translation of the group name in the given language.

Note I think saying "language" is more accurate than saying "locale", those are different, right? Locale would be like "en_US" and language is just "en"?

`lang:text` as well as the example could also go in backticks.

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+
+``--translated-description=<lang:text>``
+    This is another description for the group within the locale given. The syntax is lang:text. Eg. en:my-group-description-in-english.
```
lukash, line 76: Same comment as above (for `--translated-name`) applies here.

doc/groups-manager.rst
```diff
@@ -0,0 +1,94 @@
...
+
+``--user-visible``
+    Make the group visible in dnf grouplist (this is the default).
+
+``--not-user-visible``
+    Make the group not visible in dnf grouplist.
```
lukash, line 82: Not sure you wanna make this `dnf grouplist` as the command (in backticks)?

dnf-plugins-core.spec
```diff
@@ -579,6 +588,7 @@ PYTHONPATH=./plugins nosetests-%{python3_version} -s tests/
 %{_bindir}/yum-config-manager
 %{_bindir}/yum-debug-dump
 %{_bindir}/yum-debug-restore
+%{_bindir}/yum-groups-manager
```
lukash, line 591: Do we really want to add this yum compatibility "bloat" at this stage? I'd only do it if explicitly requested, and from the bug it appears it is not (_"They requested to have yum-groups-manager (**or a similar capability**) ported over to dnf."_).

lukash (commented):
I'd actually prefer to have the commits here squashed, as they capture the development process more than logical separation. I don't want to review commit by commit, as the successive commits amend and fix the previous ones. So I review the final file as a whole locally, but then it's hard to find the commits where to actually put my comments to.

Hence I haven't really reviewed much of the code yet.

plugins/groups_manager.py
m-blaha: This is how original yum-groups-manager behaves. But thinking about it, you are right and we might be more strict - error out on any xml reading failure.
The return value of `load_input_files()` is left-over from previous version of the code, not used and I will remove it.

plugins/groups_manager.py
m-blaha: This check is meant for this situation:
`dnf groups-manager --name=ěščř`
In case the id of newly added group is not passed using `--id` command line argument the plugin generates one based on the name. The generator is quite simple (`text_to_id()` function) and it just converts name to lower case and removes characters that are not allowed in the id. And this might fail in case there are no characters left after the removal.

doc/groups-manager.rst
m-blaha: Well, the plugin can operate in two "modes"
1. by default it adds packages passed as positional arguments to selected group
2. with `--remove` option it removes them

dnf-plugins-core.spec
m-blaha: It's part of separate yum compatibility package (yum-utils on RHEL, dnf-utils on Fedoras). We do this for all other yum tools like reposync, repoquery... and I think it will be more confusing for the user to have different tools handled in different way.

m-blaha:
> I'd actually prefer to have the commits here squashed, as they capture the development process more than logical separation. I don't want to review commit by commit, as the successive commits amend and fix the previous ones. So I review the final file as a whole locally, but then it's hard to find the commits where to actually put my comments to.

I will fix issues you have found so far and squash all the development commits into one.

doc/groups-manager.rst
lukash: Yeah, sorry for not being specific, my comment was just about the language - "remove to" reads weird, it should be "remove from". Also note the indefinite article: "a group"

dnf-plugins-core.spec
lukash: Ok, I don't know, my point is we want users to stop using these, they're backwards compatibility only. Ideally new users shouldn't be confused as they shouldn't be using them. Right? I don't think they're marked as deprecated, except for that first line of text in `doc/dnf-utils.rst` saying they're for backwards compatibility? (meaning the users don't necessarily get the right message)

I get your point about consistency, but i don't see that much value in adding in essence deprecated compatibility layer to maintain the consistency...

That said I'm not going to insist if you insist :slightly_smiling_face: I'd just rather be removing these instead of adding them.

plugins/groups_manager.py
```diff
@@ -0,0 +1,307 @@
...
+        parser.add_argument('--print', action='store_true', default=False,
+                            help=_('print the result metadata to stdout'))
+        # group options
+        parser.add_argument('--id', type=group_id_type,
+                            help=_('group id'))
+        parser.add_argument('-n', '--name', help=_('group name'))
+        parser.add_argument('--description',
+                            help=_('group description'))
+        parser.add_argument('--display-order', type=int,
+                            help=_('sort order override'))
```
lukash, line 105: I feel the help text doesn't really describe the option - at first I thought it's influencing the output in some way, but it is actually setting a group's attribute. Saying simply "group display order" seems better.

plugins/groups_manager.py
```diff
@@ -0,0 +1,307 @@
...
+                             help=_('make the group user invisible'))
+
+        # package list options
+        section = parser.add_mutually_exclusive_group()
+        section.add_argument('--mandatory', action='store_true',
+                             help=_('add packages to the mandatory section'))
+        section.add_argument('--optional', action='store_true',
+                             help=_('add packages to the optional section'))
+        section.add_argument('--remove', action='store_true', default=False,
+                             help=_('remove packages from group instead of adding them'))
```
lukash, line 127: > remove packages from **the** group

plugins/groups_manager.py
```diff
@@ -0,0 +1,307 @@
...
+                if file_name.endswith('.gz'):
+                    # libcomps does not support gzipped files - decompress to temporary
+                    # location
+                    with gzip.open(file_name) as gz_file:
+                        with tempfile.NamedTemporaryFile() as temp_file:
+                            shutil.copyfileobj(gz_file, temp_file)
+                            # The xml written to temp_file  might remain in buffers.
+                            # Force writing it to disk.
+                            temp_file.file.flush()
+                            os.fsync(temp_file.file.fileno())
```
lukash, line 175: Shouldn't you close the file instead? I.e. do the reading on the line below out of the `with` block?

Also, shouldn't the `temp_file` be cleaned up? I would guess you can probably do it straight after loading it? (Regardless of whether the loading was successful)

plugins/groups_manager.py
```diff
@@ -0,0 +1,307 @@
...
+                            file_comps.fromxml_f(temp_file.name)
+                else:
+                    file_comps.fromxml_f(file_name)
+            except (IOError, OSError, libcomps.ParserError) as err:
+                # gzip module raises OSError on reading from malformed gz file
+                logger.error(_("Can't load file \"{}\": {}").format(file_name, err))
```
lukash, line 181: How about putting this line into the description of `dnf.exceptions.Error()` instead of logging it here? The error is without a description and here it is logged, I know the order will be different but it seems cleaner to me.

plugins/groups_manager.py
```diff
@@ -0,0 +1,307 @@
...
+
+    def save_output_files(self):
+        for file_name in self.opts.save:
+            try:
+                # xml_f returns a list of errors / log entries
+                errors = self.comps.xml_f(file_name, xml_options=COMPS_XML_OPTIONS)
+            except libcomps.XMLGenError as err:
+                errors = [err]
+            if errors:
+                logger.error(''.join(errors).strip())
```
lukash, line 201: Does this not result in an error (along with a non-zero exit code)? Seems like there should be an exception raised. Similar to the load method, I would put the error description into the exception if there is just one. If there are multiple I would log them and then raise an exception like "Error saving file FILENAME".

lukash (commented):
Looking mostly good, just a few comments. But, the plugin doesn't work with environments at all, I assume that's on purpose?

plugins/groups_manager.py
m-blaha: I used `with` block so that I do not need to care about handling error states, closing files and removing the temporary one - it is all provided by the context managers. That is also why I cannot do the reading out of the `with` block.

lukash: Ok, I see. It's a bit awkward, the fact you need to do two calls and the second one being kind-of non-obvious doesn't make me too confident about this..

m-blaha: I've changed flushing the file to closing it. Does this looks better?

plugins/groups_manager.py
m-blaha: I've changed it so that the exit code 1 is returned (along with details of the error logged) when saving the output file fails.

plugins/groups_manager.py
m-blaha: Yes, thanks - I've changed it according to your suggestion.

plugins/groups_manager.py
lukash: Actually, why not just use `fromxml_str()`? I guess that does mean loading the whole contents into memory... Is that an issue here?

What I don't like abou the new solutoin is using the context manager and then just circumventing it for the closing/unlinking... so if you wanna go this way, maybe just not use the context manager at all? Sorry about the back and forth...

plugins/groups_manager.py
```diff
@@ -0,0 +1,312 @@
...
+                # xml_f returns a list of errors / log entries
+                errors = self.comps.xml_f(file_name, xml_options=COMPS_XML_OPTIONS)
+            except libcomps.XMLGenError as err:
+                errors = [err]
+            if errors:
+                if len(errors) > 1:
+                    errors.append(_("Error saving file {}").format(file_name))
+                for err in errors[:-1]:
+                    logger.error(err.strip())
+                raise dnf.exceptions.Error(errors[-1].strip())
```
lukash, line 206: Feels like I'm missing something once again, but why append the "Error saving file" to the list of errors, only to then skip it when logging and pick it from the end of the list when throwing the exception?

If there's only one error, you'll raise it directly, but then you don't include the "Error saving file PATH" part... Seems like the error handling should be the same as when loading, which seems to make more sense?

plugins/groups_manager.py
m-blaha: Exactly. I wanted to avoid loading the whole file into memory. I'll remove the inner context manager since is really doesn't do much.

plugins/groups_manager.py
m-blaha: The problem here is that there is not one particular error that occurred during saving the file - there is whole list of them. So in case the particular error is not clear (i.e. number of errors returned is greater than one) I add a generic "Error saving file" message without any details (the details were logged before).

But there of course are other possibilities - for example log all errors except the last one and raise the last one as `dnf.exceptions.Error` hoping that it contains something reasonable. That actually might make sense given that I was not able to make libcomps emit more than one error during xml saving.

As far as error message wording - it sounds a bit clumsy (either because dnf.Error handler automatically adds 'Error: ' prefix or due to repeating the libcomps error). That is why I did not add the "Error saving file" part (but I'm not sure I can rely on libcomps adding the filename to its message):
```
Error: Error saving file "nnn/out.xml": Can't write file nnn/out.xml
Error: Can't save file "nnn/out.xml": Can't write file nnn/out.xml
Error: Can't write file nnn/out.xml
```

There is one more problem - libcomps writes some errors directly to stdout (sometimes even more than once) and do not return them in error array (but this is out of scope of this PR). So currently the full output in case of trying to save the xml to non-existend directory looks like this:
```
I/O error : No such file or directory
I/O error : No such file or directory
Error: Can't write file nonexistent/path/out.xml
```
The I/O error lines are written directly by libcomps and I cannot handle them.

lukash: Okay, seems like you'd have to go fixing libcomps, which is probably not a priority now...

I think my point really is why not do it the same way as in `load_input_files()`, i.e. log all errors in the `errors` list and then raise an exception using the message from the exception you've caught. Now I'm not sure you've answered this question in your writeup :slightly_smiling_face:

plugins/groups_manager.py
lukash: Thanks, I think this is better than the original :+1:

plugins/groups_manager.py
m-blaha: The thing is, that usually there is no exception caught. Just list of returned errors.

lukash: Ok, I don't know, let's keep it this way then... can you add a short comment describing the situation so that future readers know what's up?

PR #414, opened 2020-11-20
[repomanage] Don't use cached metadata

kontura:
Always expire the repo, otherwise repomanage could use cached metadata
and give identical results for multiple runs (depending on metadata_expire
config value) even if the actual repo changed in the meantime.

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/913

### DIFF
--- a/plugins/repomanage.py
+++ b/plugins/repomanage.py
@@ -71,8 +71,11 @@ class RepoManageCommand(dnf.cli.Command):
             raise dnf.exceptions.Error(_("No files to process"))
 
         try:
-            this_repo = self.base.repos.add_new_repo("repomanage_repo", self.base.conf, baseurl=[self.opts.path])
-            self.base._add_repo_to_sack(this_repo)
+            repo_conf = self.base.repos.add_new_repo("repomanage_repo", self.base.conf, baseurl=[self.opts.path])
+            # Always expire the repo, otherwise repomanage could use cached metadata and give identical results
+            # for multiple runs even if the actual repo changed in the meantime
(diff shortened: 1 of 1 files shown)


pkratoch:
LGTM. Btw, doesn't it fix bug https://bugzilla.redhat.com/show_bug.cgi?id=1899852 ?

kontura:
> LGTM. Btw, doesn't it fix bug https://bugzilla.redhat.com/show_bug.cgi?id=1899852 ?

I think so.

kontura:
I have asked the reporter to see if its the same problem.

kontura:
I have added a commit that fixes ordering of lists for module versions. This was causing newest packages to be shown as older and the oldest packages as newer.

pkratoch:
> I have asked the reporter to see if its the same problem.

It looks like the same. Can you please add a reference to the bug into the commit message?

kontura:
Thanks! I have added the Bugzilla reference and also changelogs for both commits, since they both fix important bugs I think.

pkratoch:
> Thanks! I have added the Bugzilla reference and also changelogs for both commits, since they both fix important bugs I think.

Great, thanks! Just a last nitpick: the changelog message for the first commit mentions multiple runs, but repomanage doesn't use cached metadata even for one run. The multiple runs were only an easy way to test the bug, but are not necessary, are they?

kontura:
You are correct. :slightly_smiling_face:

pkratoch:
The tests passed locally.

PR #415, opened 2020-11-25
copr: adding calling of the _guess_chroot method

schlupov:
when the user runs the command sudo dnf copr enable user/project sudo dnf install project -y then sudo is taken as the chroot variable and the _guess_chroot method is never called, enabling the copr project then ends with an error in the _download_repo method, because the chroot does not take the form distribution-version-architecture

Relates: bg#1801017

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -223,8 +223,15 @@ class CoprCommand(dnf.cli.Command):
                   'copr command are required'))
         try:
             chroot = self.opts.arg[1]
+            if len(self.opts.arg) > 2:
+                raise dnf.exceptions.Error(_('Too many arguments.'))
+            self.chroot_parts = chroot.split("-")
+            if len(self.chroot_parts) < 3:
+                raise dnf.exceptions.Error(_('Bad format of optional chroot. The format is '
+                                             'distribution-version-architecture.'))
         except IndexError:
             chroot = self._guess_chroot()
+            self.chroot_parts = chroot.split("-")
 
         # commands without defined copr_username/copr_projectname
         if subcommand == "search":
@@ -267,7 +274,7 @@ Bugzilla. In case of problems, contact the owner of this repository.
                                 copr_projectname])
             msg = "Do you really want to enable {0}?".format(project)
             self._ask_user(info, msg)
-            self._download_repo(project_name, repo_filename, chroot)
+            self._download_repo(project_name, repo_filename)
             logger.info(_("Repository successfully enabled."))
             self._runtime_deps_warning(copr_username, copr_projectname)
         elif subcommand == "disable":
@@ -453,11 +460,9 @@ Bugzilla. In case of problems, contact the owner of this repository.
             chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
         return chroot
 
-    def _download_repo(self, project_name, repo_filename, chroot=None):
-        if chroot is None:
-            chroot = self._guess_chroot()
-        short_chroot = '-'.join(chroot.split('-')[:2])
-        arch = chroot.split('-')[2]
+    def _download_repo(self, project_name, repo_filename):
+        short_chroot = '-'.join(self.chroot_parts[:-1])
+        arch = self.chroot_parts[-1]
         api_path = "/coprs/{0}/repo/{1}/dnf.repo?arch={2}".format(project_name, short_chroot, arch)
 
         try:
@@ -660,7 +665,7 @@ class PlaygroundCommand(CoprCommand):
                 f.close()
                 if (output2 and ("output" in output2)
                         and (output2["output"] == "ok")):
-                    self._download_repo(project_name, repo_filename, chroot)
+                    self._download_repo(project_name, repo_filename)
             except dnf.exceptions.Error:
                 # likely 404 and that repo does not exist
                 pass


plugins/copr.py
```diff
@@ -223,6 +223,9 @@ def run(self):
                   'copr command are required'))
         try:
             chroot = self.opts.arg[1]
+            if len(self.opts.arg) > 2:
+                self.cli.optparser.print_help(self)
+                raise dnf.cli.CliError(_('too many arguments, a maximum of two arguments are allowed'))
```
lukash, line 228: Looking at the referenced bug: https://bugzilla.redhat.com/show_bug.cgi?id=1801017

This doesn't solve the original traceback in the report:
```
# dnf copr enable kwizart/fedy foo
...
Do you really want to enable copr.fedorainfracloud.org/kwizart/fedy? [y/N]: y
  File "/home/lu/dev/dnf/bin/dnf-3", line 58, in <module>
  File "/home/lu/dev/dnf/dnf/cli/main.py", line 201, in user_main
    errcode = main(args)
  File "/home/lu/dev/dnf/dnf/cli/main.py", line 67, in main
    return _main(base, args, cli_class, option_parser_class)
  File "/home/lu/dev/dnf/dnf/cli/main.py", line 106, in _main
    return cli_run(cli, base)
  File "/home/lu/dev/dnf/dnf/cli/main.py", line 122, in cli_run
    cli.run()
  File "/home/lu/dev/dnf/dnf/cli/cli.py", line 1167, in run
    return self.command.run()
  File "/home/lu/dev/dnf-plugins-core/plugins/copr.py", line 273, in run
    self._download_repo(project_name, repo_filename, chroot)
  File "/home/lu/dev/dnf-plugins-core/plugins/copr.py", line 463, in _download_repo
    arch = chroot.split('-')[2]
IndexError: list index out of range
```

The problem is two-fold, one thing is checking for the correct number of arguments, the other is validation of the chroot argument being a valid chroot.

lukash (commented):
Hello @schlupov, thanks for the PR. A few comments:
- Please wrap the lines on the commit description to 72 characters line length
- The `sudo dnf copr enable user/project sudo dnf install project -y` example is quite confusing, because it is obviously an invalid concatenation of two commands, I understand it after looking at the referenced bug, but on it's own it's hard to understand. To be honest I'd just leave it out and provide a short description of the commit fixing incorrect number of arguments and validating the chroot.
- We have a common format for referencing related bugs, check the `README.rst` for details, tldr; you can run `git config commit.template ./.git-commit-template` to use our commit template. Please use that instead of _"Relates: bg#1801017"_.

lukash:
I'll also note we have tests for the copr plugin here: https://github.com/rpm-software-management/ci-dnf-stack/blob/master/dnf-behave-tests/features/plugins-core/copr.feature

This could be potentially added there, but the CI stack is hard to run so I'm not asking you to deal with that, this is just another reminder for us to work on lowering the barrier for CI tests contributions.

plugins/copr.py
schlupov: I know, unfortunately, the format of the chroot is not defined anywhere, so in the end, I just add a check on the number of arguments.

lukash: Okay, but you've got this error here which is causing a traceback, and the problem is the code expects the input to be in a certain format, which it isn't. So instead of a traceback you should print a message for this case, it's up to you how specific you want to be in the message - e.g. you can just say "invalid chroot format", or preferably state some details about the necessary format, it appears to be `distro-version-arch`.

So, IMO a good solution here is to split the chroot by `'-'` and verify the resulting list has exactly 3 items, because that's what the code expects. If it isn't, print an error message.

praiskup: > So, IMO a good solution here is to split the chroot by '-' and verify the resulting list has exactly 3 items, because that's what the code expects. If it isn't, print an error message.

+1, long-term we should fix the copr plugin and copr base code to allow artificial strings there (at least mock supports it, see e.g. mock-rpmfusion-free.rpm chroots).  But at this moment copr.py plugin seems to indeed strictly expect three parts in the chroot name so we should prevent the tracebacks.

schlupov: I added a control to the chroot format. Now we expect the format distribution-version-architecture, with anything else an error message will be displayed.

plugins/copr.py
```diff
@@ -456,8 +462,12 @@ def _guess_chroot(self):
...
         if chroot is None:
             chroot = self._guess_chroot()
-        short_chroot = '-'.join(chroot.split('-')[:2])
-        arch = chroot.split('-')[2]
+        if len(chroot.split("-")) == 4:
+            short_chroot = '-'.join(chroot.split('-')[:3])
+            arch = chroot.split('-')[3]
+        else:
+            short_chroot = '-'.join(chroot.split('-')[:2])
+            arch = chroot.split('-')[2]
```
lukash, line 470: What if the length of the splitted list is 5 or greater? Will this do what you expect? Maybe the initial check on line 229 should be for length equal to 3 or 4?

You are also calling `chroot.split()` multiple times, while in practice it probably has no impact, it's not nice, you should split once, store the splitted list and use that.

plugins/copr.py
```diff
@@ -223,6 +223,12 @@ def run(self):
                   'copr command are required'))
         try:
             chroot = self.opts.arg[1]
+            if len(self.opts.arg) > 2:
+                raise dnf.exceptions.Error(_('Too many arguments.'))
+            chroot_parts = chroot.split("-")
+            if len(chroot_parts) < 3:
+                raise dnf.exceptions.Error(_('Bad format of optional chroot. Format is '
+                                             'distribution-version-architecture.'))
```
lukash, line 231: Please add a definite particle for the format in the message:
> Bad format of optional chroot. **The** format is ...

Also, why not parse and store the `short_chroot` and `arch` in `self` here, so that you don't need to split the `chroot` for the second time later on? It's a bit cleaner to do all the validation and processing of input in one place and then just use those values.

plugins/copr.py
schlupov: if the list is longer, it still works. Currently 2 is taken as the index on line 465, but this obviously doesn't matter, because for example api_path /coprs/schlupov/test_chroot/repo/opensuse-leap-15.2/dnf.repo?arch=x86_64 for some reason works the same as /coprs/schlupov/test_chroot/repo/opensuse-leap/dnf.repo?arch=15.2.

lukash: Well, I don't find anything too obvious about this :/

The fact the two URLs both work is already quite confusing and just coincidental, I guess it's because the last part of the path is joined with the `arch` argument using a `"-"`, but I wonder what the architecture is for the latter URL. Is there a default of `"x86_64"` in case the architecture is missing?

So for e.g. a chroot like this: `"a-b-c-d-arch"` your current code will process it so that `short_chroot = "a-b"`, `arch = "c"` and `"d-arch"` will just be ignored, right? I haven't tried to run it to see what the error message would be ultimately, but I would guess it can't be right. Since I assume there is no valid chroot that has five elements like this, you may think it's not an issue, but if a user enters such chroot just by mistake, the error message they'll get will just be confusing.

You are imposing a structure by "parsing" the chroot and you should handle all corner cases.

You can just check the chroot splitted length is either 3 or 4 in the check on line 229 and actually also include length of 4 variant in the error message (right now it says the format is `"distribution-version-architecture"` but for length of 4 that's different.

praiskup: > just coincidental

That's not a coincidence, it's the intentional design of that route (you can name the '*.repo' link as you wish).

> You can just check the chroot splitted length is either 3 or 4

There's no reason to limit that to 3/4.  And actually, this is a can of worms, because e.g. mock-rpmfusion-free
package doesn't hold the `arch` on last position and it has more parts.  So strictly speaking, the plugin needs
some major re-design and this PR isn't the right time/place for it.

We should concentrate on the traceback.

plugins/copr.py
```diff
@@ -456,8 +462,9 @@ def _guess_chroot(self):
     def _download_repo(self, project_name, repo_filename, chroot=None):
         if chroot is None:
             chroot = self._guess_chroot()
-        short_chroot = '-'.join(chroot.split('-')[:2])
-        arch = chroot.split('-')[2]
+        index = 3 if len(self.chroot_parts) == 4 else 2
+        short_chroot = '-'.join(self.chroot_parts[:index])
+        arch = self.chroot_parts[index]
```
praiskup, line 467: I'd rather like to see something like:
```
# somewhat artificially we expect arch is on the last position
short_chroot = '-'.join(self.chroot_parts[:-1])
arch = self.chroot_parts[-1]
```

praiskup: But note that if you do just `dnf copr enable @copr/copr`, you get `self.opts.arg == ["@copr/copr"]` (len 1)
and the IndexError exception causes that `chroot_parts` is undefined here.

lukash: Okay then. It can work with using `-1` as the index, the only question is what to do when len is 1.

praiskup:
> when len is 1

Not very nicely we return "Error: exactly two additional parameters to copr command are required" now :-(.  We should at least say `specify <project> [<chroot>]` or something more friendly ... but that's yet another different issue.

plugins/copr.py
praiskup: >  when len is 1

Not very nicely we return "Error: exactly two additional parameters to copr command are required" now :-(. We should at least say specify <project> [<chroot>] or something more friendly ... but that's yet another different issue.

schlupov: I made a change as suggested by praiskup. I took care of the situation when `dnf copr enable @copr/copr`

plugins/copr.py
```diff
@@ -454,10 +461,8 @@ def _guess_chroot(self):
         return chroot
 
     def _download_repo(self, project_name, repo_filename, chroot=None):
-        if chroot is None:
-            chroot = self._guess_chroot()
```
praiskup, line 458: Looks like you removed this unintentionally

schlupov: No, that was intentional, at this point, it can't happen that the chroot is None.

lukash: Nitpicking, but then you could also remove the `=None` from the function arguments :slightly_smiling_face:. Otherwise LGTM now!

schlupov: Good point, I removed it.

lukash:
@schlupov thanks for the contribution and for bearing with me :slightly_smiling_face:

@rh-atomic-bot r+

xsuchy:
GH say that test failed, but Jenkins report SUCESS. @lukash can you resolve this dilema? :)

lukash:
@xsuchy I assume you're looking at the dnf-dnf-plugins-core-update-gh-status job in Jenkins, which finished with success, reporting an error :slightly_smiling_face: the problem is unrelated, librepo fails to build right now, I'll fix it and then retry here.

lukash:
@rh-atomic-bot retry

lukash:
@schlupov oh, you've actually deleted the whole `chroot=None` argument, instead of just `=None`. Which seems to be correct, the argument is not used in the function anymore. But it is still being called with the extra argument (from two places) and it tracebacks on it now. I apparently haven't been paying enough attention, luckily the tests caught it :slightly_smiling_face: (sorry it's not easy to link to the error in the tests)

schlupov:
Oh, I see :-) My mistake, I fixed it.

lukash:
@rh-atomic-bot r+

PR #416, opened 2020-12-03
[reposync] Fix downloading uncompressed comps.xml (RhBug:1895059)

m-blaha:
"reposync -m" command did not copy comps.xml to destination location in
case the group metadata in repository cache was a plain xml file (not
compressed).

https://bugzilla.redhat.com/show_bug.cgi?id=1895059

Test: https://github.com/rpm-software-management/ci-dnf-stack/pull/921

### DIFF
--- a/plugins/reposync.py
+++ b/plugins/reposync.py
@@ -218,7 +218,10 @@ class RepoSyncCommand(dnf.cli.Command):
             dest_path = self.metadata_target(repo)
             dnf.util.ensure_dir(dest_path)
             dest = os.path.join(dest_path, 'comps.xml')
-            dnf.yum.misc.decompress(comps_fn, dest=dest)
+            if dnf.yum.misc.decompress(comps_fn, dest=dest) == comps_fn:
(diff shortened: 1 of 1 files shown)


kontura:
I am thinking it might be better to change the `decompress` function to actually copy an uncompressed file if it has `dest` specified.

I have looked where else it is used and I think its causing another bug in `read_comps`. Specifically the `decompress` function is used in `repo_gen_decompress` but if it gets an uncompressed file its not copied to the `dirname/gen/generated_name` location. If later dnf is run with `--cacheonly` the `read_comps` is looking for the comps file in `dirname/gen/generated_name` but its not there and the command fails even though the comps.xml is cached.

I made a failing test that illustrates it more clearly: https://github.com/kontura/ci-dnf-stack/commit/a165d420aee3686bc50bc3f15d6d37dedb448a6a

m-blaha:
@kontura thanks for pointing out the related issue. I'll take another look.

m-blaha:
Closed in favour of fixing it in dnf: https://github.com/rpm-software-management/dnf/pull/1702

PR #418, opened 2020-12-10
Modify behaviour of generate_completion_cache

j-mracek:
It will generate only cache when transaction is available. It avoid an
issue with detection of metadata regeneration. The problem resulted in
regeneration of cache for every run. On my system it represents about
0.3 s performance improvement.

It will also store only names of rpms, therefore database will be
smaller.


Conan-Kudo:
bors try

Conan-Kudo (approved):

lukash:
@Conan-Kudo the CI is now failing on an unrelated error, @m-blaha is looking into it... We're also hopefully close to having a CI running on Github Actions, look forward to that :slightly_smiling_face:

PR #419, opened 2020-12-10
Set ASAN_OPTIONS for tests

lukash:


kontura:
LGTM and  it builds successfully when libdnf is compiled with sanitizers.

Plus is enables `make test` which is nice.
Thanks!

PR #423, opened 2021-01-29
Add missing command line option to documentation

m-blaha:
--alldeps option of download command was not documented.

### DIFF
--- a/doc/download.rst
+++ b/doc/download.rst
@@ -68,6 +68,9 @@ All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for det
 ``--resolve``
     Resolves dependencies of specified packages and downloads missing dependencies in the system.
 
+``--alldeps``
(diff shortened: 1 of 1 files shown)


doc/download.rst
```diff
@@ -68,6 +68,9 @@ All general DNF options are accepted, see `Options` in :manpage:`dnf(8)` for det
 ``--resolve``
     Resolves dependencies of specified packages and downloads missing dependencies in the system.
 
+``--alldeps``
+    When used with ``--resolve`` download all dependencies (do not skip already installed ones)
```
lukash, line 72: > When used with ``--resolve``**,** download all dependencies (do not skip already installed ones)**.**

Not sure it's visible, I've added a comma after `--resolve` and a full stop at the end.

m-blaha: Thanks, fixed.

PR #424, opened 2021-02-05
Set blacklist subcommand as deprecated

j-mracek:
There is an initiative for removal of politically problematic words from
IT projects (blacklist, whitelist, slave).

### DIFF
--- a/plugins/versionlock.py
+++ b/plugins/versionlock.py
@@ -28,6 +28,7 @@ import hawkey
 import os
 import tempfile
 import time
+import warnings
 
 NOT_READABLE = _('Unable to read version lock configuration: %s')
 NO_LOCKLIST = _('Locklist not set')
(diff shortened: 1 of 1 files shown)


j-mracek:
bors try

plugins/versionlock.py
```diff
@@ -152,6 +154,11 @@ def run(self):
                 self.opts.package.insert(0, self.opts.subcommand)
             elif self.opts.subcommand in EXC_CMDS:
                 cmd = 'exclude'
+            elif self.opts.subcommand in DEP_EXC_CMDS:
+                msg = "Subcommand '{}' is deprecated. Use 'exclude' subcommand instead.".format(
+                    self.opts.subcommand)
+                warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
m-blaha, line 160: We have dnf specific subclass of `DeprecationWarning` in `dnf.exceptions`. Maybe we should use it?

j-mracek: I am really not sure. I tried following but it did not provide any message to the terminal.
```
dnf.exceptions.DeprecationWarning(msg)
#warnings.warn(msg, DeprecationWarning, stacklevel=2)
```
What I really don't want is to raise an exception because then it will change the behaviour. @m-blaha What do you think?

j-mracek: Finally I get how to change it. I am sorry for delay.

m-blaha: Yes, this is what I meant by my comment :)

m-blaha:
bors try

m-blaha (approved):

PR #426, opened 2021-03-10
[reposync] Improve documentation of --source option

pkratoch:
Previous description may have implied that user must use the --source
option to download also the source packages and that they will be
excluded otherwise.

The new description is more in line with other options (e.g. --arch or
--newest-only) and implies that the option only limits the reuslt.

= changelog =
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1914827


m-blaha:
Thanks for clarifying this.

PR #427, opened 2021-03-18
Removed dependency on dnf.yum.misc.Checksum class

malmond77:
https://github.com/rpm-software-management/dnf/pull/1743 removes code this package uses, so we replaced it. We need to ensure that dnf does not permit version of this package before 4.0.20 to be installed else they will not be able to resolve the class. See https://github.com/rpm-software-management/dnf/pull/1743#pullrequestreview-615177905 for context.

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -33,7 +33,7 @@
 %endif
 
 Name:           dnf-plugins-core
-Version:        4.0.19
+Version:        4.0.20
 Release:        1%{?dist}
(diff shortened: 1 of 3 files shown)


m-blaha:
The build has failed - see https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/CI-dnf-plugins-core-pr427/fedora-33-x86_64/02081443-dnf-plugins-core/builder-live.log.gz

m-blaha:
Thanks, now the tests are passing.

PR #431, opened 2021-06-08
Repomanage (docs) enhacement (RhBug:1898293) 

kontura:


doc/repomanage.rst
```diff
@@ -31,9 +31,10 @@ Synopsis
...
+`repomanage` prints newest or older packages in a repository specified by <path> for easy piping to xargs or similar programs. In case <path> doesn't contain a valid repodata it is searched for rpm packages which are then used instead.
+If the repodata are present repomage uses them as the source of truth, it doesn't verify that they match with the present rpm packages. In fact repomanage can run with just the repodata, no rpm packages are needed.
 
-In order to work correctly with modular packages <path> has to contain repodata with modular metadata. If modular content is present `repomanage` prints packages from newest or oldest versions of each stream in addition to newest or oldest non-modular packages.
+In order to work correctly with modular packages <path> has to contain repodata with modular metadata. If modular content is present `repomanage` prints packages from newest or older stream versions in addition to newest or older non-modular packages.
```
pkratoch, line 37: Looks good, both the code and the explanation. I'd just add a few commas, so that it reads cleaner:

In case `<path>` doesn't contain a valid repodata, it is searched...
If the repodata are present, repomage uses them...
In fact, repomanage can run...
In order to work correctly with modular packages, `<path>` has to contain...
If modular content is present, `repomanage` prints...

General rule is that there should be a comma in a sentence "If condition, something.", but not in a sentence "Something if condition." (because there the "if" separates the two parts clearly enough). Similarly for "in order to" and "in case". And I am not absolutely sure with the "in fact", but I'd put a comma there, because it's fronted (it doesn't belong to the "repomanage", but rather to the "can run" if it makes sense).

Oh, and also, I think you can drop the "with" in "they match with the present rpm packages".

kontura: Thank you for the feedback and explanations!

Updated the PR.

PR #432, opened 2021-07-08
[needs-restarting] Fix wrong boot time (RhBug:1960437)

inknos:


m-blaha:
I did a brief investigation on detecting the boot time and found this:

1. `grep btime /proc/stat`
- gives time when the kernel booted. This works well except for containers - they do not have their own kernel so the boot time of the host kernel is returned instead.

2. `stat /proc/1/cmdline`
- unfortunately doesn't work when hibernation is used. It returns time when the machine woke up from hibernation instead of the boot time. I'm not sure whether this is the reason of incorrect boot time in bz#1960437 (add needinfo from the reporter comment?)

3. `stat /proc/1`
- using `/proc/1` directory instead of `/proc/1/cmdline` file seems to work reliably on bare metal (even after hibernation), virtual machines, podman containers, and systemd-nspawn containers.

inknos:
I updated the code using `/proc/1` and tested on ci, machinectl and podman. looks like it works

PR #433, opened 2021-07-12
Turn off dnf_local by default

panchenbo:
DNF plugin local not provides /var/ lib/DNF/plugins/local/, so exec command "yum list" after installation will prompt the following errors:
image

### DIFF
--- a/etc/dnf/plugins/local.conf
+++ b/etc/dnf/plugins/local.conf
@@ -1,5 +1,5 @@
 [main]
-enabled = true
+enabled = false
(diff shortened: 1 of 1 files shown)


lukash:
The plugin is in a separate package, if you don't want to use it, don't install it. If you decide to install it, it is supposed you will provide the directory with your local repository.

panchenbo:
> The plugin is in a separate package, if you don't want to use it, don't install it. If you decide to install it, it is supposed you will provide the directory with your local repository.
ok

PR #435, opened 2021-07-16
Fix 'dnf copr enable' on Fedora 35

praiskup:
The output from linux_distribution() changed so it returns:
>>> distro.linux_distribution()
('Fedora Linux', '35', '')

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -431,7 +431,7 @@ Bugzilla. In case of problems, contact the owner of this repository.
             dist = linux_distribution()
         # Get distribution architecture
         distarch = self.base.conf.substitutions['basearch']
-        if "Fedora" in dist:
+        if any([name in dist for name in ["Fedora", "Fedora Linux"]]):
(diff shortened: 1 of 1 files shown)


xsuchy (approved):

xsuchy:
@rh-atomic-bot r+

poncovka:
Could you please release a new version with this fix soon? I think it could fix our tests run with `packit` on rawhide (http://artifacts.dev.testing-farm.io/5edbfc8d-7f1e-4846-9cf4-06e24342874a/).

ericcurtin:
Hi @praiskup @xsuchy and @poncovka why was this needed?

The old code should have worked for "Fedora" and "Fedora Linux" as "Fedora" is a substring of both?

In fact I want to revert it back because the old code works for the case it's "Asahi Fedora Linux Remix" also

praiskup:
Good question, after two years I don't remember :-) the `dist` seems to contain a triplet after a quick look.  But a better commit message would be, ... better.

PR #436, opened 2021-08-02
Better error message for dnf copr enable

schlupov:
= changelog =
msg: show better error message if the command dnf copr enable fails
type: enhancement


plugins/copr.py
```diff
@@ -507,6 +509,38 @@ def _download_repo(self, project_name, repo_filename):
...
         os.chmod(repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
 
+    def _resolve_error_message(self, project_name):
+        list_chroots = requests.get(self.copr_url + "/api_3/mock-chroots/list")
+        project_name_items = project_name.split("/")
+        user = project_name_items[0]
+        project = project_name_items[1]
+        params = {"ownername": user, "projectname": project}
+        project = requests.get(self.copr_url + "/api_3/project", params=params)
+        list_builds = requests.get(self.copr_url + "/api_3/build/list", params=params)
```
praiskup, line 519: This means several API queries for each repo file miss ... could the original error 404 page be modified to provide all the necessary info, and we could just present that here?

I know that the repo file is not a part of our API, though in future it could be?

Note that it would be nice to see the info on the page itself, e.g.:
https://copr.fedorainfracloud.org/coprs/praiskup/ping/repo/epel-6/praiskup-ping-epel-6.repo?arch=ppc64le

xsuchy: It is only during a failure. So very rarely. I do not think we need to optimize here. I would rather prefer clean code over speed.

schlupov: I agree with @praiskup, those requests take time and the information could be displayed directly on the 404 page. I could add a repo file to the API.

praiskup: If you found an elegant way to read the "error message" from the current web page, I wouldn't be against keeping it as is (without an additional API route).  Perhaps we could even provide some additional HTTP header to avoid the HTML parsing?

schlupov: Do you mean that a header would contain the error message or at least part of the error message?

praiskup:
I think it could contain the whole error message?  Encoded if needed?

plugins/copr.py
```diff
@@ -465,35 +467,32 @@ def _download_repo(self, project_name, repo_filename):
...
+                raise dnf.exceptions.Error(_(error_msg))
+            else:
+                raise dnf.exceptions.Error(_("Such repository does not exist."))
+
+        if os.path.exists(repo_filename):
+            os.remove(repo_filename)
+        if re.match(r"\[copr:", response.text):
+            start = response.text.find("[")
+            end = response.text.find("]")
+            repo_filename = os.path.join(self.base.conf.get_reposdir, "_" + response.text[start+1:end] + ".repo")
```
praiskup, line 495: This is an unnecessarily invasive change, perhaps we need to split this to two or more commits to better explain why it is needed?

plugins/copr.py
praiskup: Yes, e.g. part of the message.  Or even e.g. some json data, like `{'available_chroots': ['a', 'b', 'c']}` (we could easily construct useful message from such data here).

plugins/copr.py
schlupov: It doesn't make sense to me to split it into multiple commits. The use of the re library is now applied to the entire text from the response.text, originally the f file obtained using self.base.urlopen was read line by line, now requests.get is used because I could not read the headers in the original version. Changes in the use of urlopen and requests are related.

plugins/copr.py
schlupov: See lines
```
msg = base64.b64decode(response.headers["Error-Message"]).decode('utf-8')
error_msg = ''.join(["It wasn't possible to enable this project.\n", msg])
```
The `Error-Message` header contains the whole error message which is intended just for cli, for HTML another message is displayed to users.

plugins/copr.py
praiskup: Well, thanks for the explanations but this is .... complicated.  This probably needs in-depth comment in commit message :-)  And at least I don't think `requests` is an acceptable dependency for the dnf-plugins-core package (you would have to put it into Requires: in the spec file).

So if headers are not readable using the dnf-internal downloader, I'd propose using the `urllib.urlopen()` (we already used that before), and using the url.readline() method (to keep the old line-by-line logic and to not change the code we don't have to change).

plugins/copr.py
```diff
@@ -465,35 +467,32 @@ def _download_repo(self, project_name, repo_filename):
...
-                                             "_" + line[1:-2] + ".repo")
-            break
+                except urllib.error.HTTPError as e:
+                    status_code = e.getcode()
+            else:
+                import urllib
+                res = urllib.urlopen(self.copr_url + "/coprs/" + project_name)
+                status_code = res.getcode()
+            if str(status_code) != '404':
+                msg = base64.b64decode(response.headers["Error-Message"]).decode('utf-8')
```
praiskup, line 484: We should use `response.headers.get()` instead, just in case the Error-Message isn't provided.

plugins/copr.py
```diff
@@ -465,35 +467,32 @@ def _download_repo(self, project_name, repo_filename):
...
-                repo_filename = os.path.join(self.base.conf.get_reposdir,
-                                             "_" + line[1:-2] + ".repo")
-            break
+                except urllib.error.HTTPError as e:
+                    status_code = e.getcode()
+            else:
+                import urllib
+                res = urllib.urlopen(self.copr_url + "/coprs/" + project_name)
+                status_code = res.getcode()
+            if str(status_code) != '404':
```
praiskup, line 483: Why don't we compare integers here?  Btw., it's weird, did you want to write `status_code == 404`?

plugins/copr.py
```diff
@@ -504,9 +503,42 @@ def _download_repo(self, project_name, repo_filename):
             if os.path.exists(old_repo_filename):
                 os.remove(old_repo_filename)
 
-        shutil.copy2(f.name, repo_filename)
+        with open(repo_filename, 'w+') as f:
+            f.write(response.text)
         os.chmod(repo_filename, stat.S_IRUSR | stat.S_IWUSR | stat.S_IRGRP | stat.S_IROTH)
 
+    def _resolve_error_message(self, project_name):
```
praiskup, line 510: I suppose this can go away entirely, as it is unused function.

schlupov:
PTAL

plugins/copr.py
```diff
@@ -482,9 +483,19 @@ def _download_repo(self, project_name, repo_filename):
...
-                if str(status_code) != '404':
-                    raise dnf.exceptions.Error(_("This repository does not have"
-                                                 " any builds yet so you cannot enable it now."))
+                if status_code != 404:
+                    error_msg = "It wasn't possible to enable this project.\n"
+                    try:
+                        if PY3:
+                            response = urllib.request.urlopen(self.copr_url + api_path)
+                        else:
+                            response = urllib.urlopen(self.copr_url + api_path)
```
praiskup, line 492: I think we already have the response in 'res' variable - so it shouldn't be needed to urlopen the file again.

plugins/copr.py
```diff
@@ -482,9 +483,19 @@ def _download_repo(self, project_name, repo_filename):
                     import urllib
                     res = urllib.urlopen(self.copr_url + "/coprs/" + project_name)
                     status_code = res.getcode()
-                if str(status_code) != '404':
-                    raise dnf.exceptions.Error(_("This repository does not have"
-                                                 " any builds yet so you cannot enable it now."))
+                if status_code != 404:
+                    error_msg = "It wasn't possible to enable this project.\n"
```
praiskup, line 487: Can we print some more info we know, like what is the owner, project and chroot?

plugins/copr.py
```diff
@@ -482,9 +483,19 @@ def _download_repo(self, project_name, repo_filename):
...
+                        if PY3:
+                            response = urllib.request.urlopen(self.copr_url + api_path)
+                        else:
+                            response = urllib.urlopen(self.copr_url + api_path)
+                    except urllib.error.HTTPError as e:
+                        msg = e.headers.get("Error-Message")
+                        if msg:
+                            msg = base64.b64decode(msg).decode('utf-8')
+                            error_msg = ''.join([error_msg, msg])
+                    raise dnf.exceptions.Error(_(error_msg))
```
praiskup, line 498: I'm not sure translation via `_()` is doable here, as the message is server-side generated thingy.  Therefore, eventually, I'd prefer if we could  send the list of available chroots, and fully construct the message here in copr.py plugin ...

praiskup:
I like the shape of this change much more now, thank you.

plugins/copr.py
praiskup: Aha, I see...  can we drop the `urllib.urlopen(self.copr_url + "/coprs/" + project_name)` calls then?  I don't think they are useful anymore, as long as we only provide message like `Such repository does not exist.` (and that could be enhanced a bit, like "Copr project `aaa/bbb` doesn't exist".

schlupov: If we drop those calls we will not know if the `dnf copr enable` fails because the project doesn't exist or just the chroot doesn't exist.

praiskup: Perhaps we could blindly claim that project doesn't exist if there's no error message?  Or does that sound too non-deterministic/optimistic :-P

praiskup:
It's a pitty to do everything twice, so small wish wish :-) can you add a new patch (applied before the one you have now) which would use:

```python
if PY3:
    from urllib.request import urlopen, HTTPError
else:
   from urllib2 import urlopen, HTTPError

res = urlopen(self.copr_url + "/coprs/" + project_name)
status_code = res.getcode()
```

Instead of this:
```
            if '404' in str(e):
                if PY3:
                    import urllib.request
                    try:
                        res = urllib.request.urlopen(self.copr_url + "/coprs/" + project_name)
                        status_code = res.getcode()
                    except urllib.error.HTTPError as e:
                        status_code = e.getcode()
                else:
                    import urllib
                    res = urllib.urlopen(self.copr_url + "/coprs/" + project_name)
                    status_code = res.getcode()
```

Perhaps we could move the import urllib part to the beginning of the file?

schlupov:
> It's a pitty to do everything twice, so small wish wish :-) can you add a new patch (applied before the one you have now) which would use:
> 
> ```python
> if PY3:
>     from urllib.request import urlopen, HTTPError
> else:
>    from urllib2 import urlopen, HTTPError
> 
> res = urlopen(self.copr_url + "/coprs/" + project_name)
> status_code = res.getcode()
> ```
> 
> Instead of this:
> 
> ```
>             if '404' in str(e):
>                 if PY3:
>                     import urllib.request
>                     try:
>                         res = urllib.request.urlopen(self.copr_url + "/coprs/" + project_name)
>                         status_code = res.getcode()
>                     except urllib.error.HTTPError as e:
>                         status_code = e.getcode()
>                 else:
>                     import urllib
>                     res = urllib.urlopen(self.copr_url + "/coprs/" + project_name)
>                     status_code = res.getcode()
> ```
> 
> Perhaps we could move the import urllib part to the beginning of the file?

I've changed these lines however, the try-except is still needed here. I've added a separate commit because of this.

plugins/copr.py
schlupov: I would rather not make any estimates  :slightly_smiling_face:

plugins/copr.py
```diff
@@ -473,20 +477,30 @@ def _download_repo(self, project_name, repo_filename):
...
+                if status_code != 404:
+                    error_msg = "It wasn't possible to enable this project.\n"
+                    try:
+                        response = urlopen(self.copr_url + api_path)
+                    except HTTPError as e:
+                        error_msg = ''.join(
+                            [error_msg,
+                             "Chroot {0} does not exist in {1}.\n".format('-'.join(self.chroot_parts), project_name)])
+                        chroots = e.headers.get("Copr-Error-Data")
+                        if chroots:
```
praiskup, line 494: Please change the variable name

plugins/copr.py
```diff
@@ -473,20 +477,30 @@ def _download_repo(self, project_name, repo_filename):
...
+                            msg = base64.b64decode(chroots).decode('utf-8')
+                            msg = json.loads(msg)
```
praiskup, line 496: Here as well.

plugins/copr.py
praiskup: To make the message translatable, you probably need to do something like `_("Available chroots: {}").format()`.  The string for translation needs to be a constant.

plugins/copr.py
```diff
@@ -472,28 +474,30 @@ def _download_repo(self, project_name, repo_filename):
...
+            error_data = e.headers.get("Copr-Error-Data")
+            if error_data:
+                error_data_decoded = base64.b64decode(error_data).decode('utf-8')
+                error_data_decoded = json.loads(error_data_decoded)
+                if len(error_data_decoded["available chroots"]) != 0:
+                    error_msg = _(''.join([error_msg, "Available chroots: {0}".format(
+                        error_data_decoded["available chroots"])]))
+                raise dnf.exceptions.Error(error_msg)
+            else:
+                raise dnf.exceptions.Error(_("Such repository does not exist."))
```
praiskup, line 494: +1 to this simplified logic

plugins/copr.py
```diff
@@ -472,28 +474,30 @@ def _download_repo(self, project_name, repo_filename):
...
+            error_msg = _("It wasn't possible to enable this project.\n")
+            error_msg = _(''.join(
+                [error_msg,
+                 "Chroot {0} does not exist in {1}.\n".format('-'.join(self.chroot_parts), project_name)]))
+            error_data = e.headers.get("Copr-Error-Data")
+            if error_data:
+                error_data_decoded = base64.b64decode(error_data).decode('utf-8')
+                error_data_decoded = json.loads(error_data_decoded)
+                if len(error_data_decoded["available chroots"]) != 0:
+                    error_msg = _(''.join([error_msg, "Available chroots: {0}".format(
```
praiskup, line 490: the `_()` method needs to have a constant, so please move it around the `"Available chroots: {0}"` string directly.

plugins/copr.py
```diff
@@ -472,28 +474,30 @@ def _download_repo(self, project_name, repo_filename):
...
-        for line in f:
-            if re.match(r"\[copr:", line):
-                repo_filename = os.path.join(self.base.conf.get_reposdir,
-                                             "_" + line[1:-2] + ".repo")
-            break
+        except HTTPError as e:
+            error_msg = _("It wasn't possible to enable this project.\n")
+            error_msg = _(''.join(
+                [error_msg,
+                 "Chroot {0} does not exist in {1}.\n".format('-'.join(self.chroot_parts), project_name)]))
```
praiskup, line 484: Movement of `_()` is needed here, too.

plugins/copr.py
```diff
@@ -472,28 +474,30 @@ def _download_repo(self, project_name, repo_filename):
...
-                    raise dnf.exceptions.Error(_("Such repository does not exist."))
-            raise
-
-        for line in f:
-            if re.match(r"\[copr:", line):
-                repo_filename = os.path.join(self.base.conf.get_reposdir,
-                                             "_" + line[1:-2] + ".repo")
-            break
+        except HTTPError as e:
+            error_msg = _("It wasn't possible to enable this project.\n")
```
praiskup, line 481: This isn't used.

plugins/copr.py
praiskup: Did you want to concatenate "Such repository doesn't exist" with the previous `error_msg` content?

plugins/copr.py
```diff
@@ -470,34 +474,28 @@ def _download_repo(self, project_name, repo_filename):
...
+            error_data = e.headers.get("Copr-Error-Data")
+            if error_data:
+                error_msg += _("Chroot {0} does not exist in {1}.\n").format('-'.join(self.chroot_parts), project_name)
```
praiskup, line 484: Please escape use `Chroot '{0}' ...` instead of `Chroot {0} ...`, so it is consistently escaped in the output with the following message.

plugins/copr.py
```diff
@@ -470,34 +474,28 @@ def _download_repo(self, project_name, repo_filename):
...
+                error_data_decoded = base64.b64decode(error_data).decode('utf-8')
+                error_data_decoded = json.loads(error_data_decoded)
+                if len(error_data_decoded["available chroots"]) != 0:
+                    error_msg += _("Available chroots: {0}").format(error_data_decoded["available chroots"])
```
praiskup, line 488: `', '.join("'{}'".format(x) for x in available_chroots)` would remove the wrapping `[]` brackets

praiskup:
Current output looks like:
```
$ sudo dnf copr enable praiskup/test-copr-enable
...
Do you really want to enable copr.fedorainfracloud.org/praiskup/test-copr-enable? [y/N]: y
Error: It wasn't possible to enable this project.
Chroot fedora-35-x86_64 does not exist in praiskup/test-copr-enable.
Available chroots: ['fedora-33-ppc64le']
```

I'd like to make it a bit more obvious by changing it to (note the Chroot vs. Repository change):

```
$ sudo dnf copr enable praiskup/test-copr-enable
...

Do you really want to enable copr.fedorainfracloud.org/praiskup/test-copr-enable? [y/N]: y

Error: It wasn't possible to enable this project.
Repository 'fedora-35-x86_64' does not exist in project 'praiskup/test-copr-enable'.
Available repositories: 'fedora-33-ppc64le'

If you want to enable a non-default repository, use the following command:
  'dnf copr enable copr.fedorainfracloud.org/praiskup/test-copr-enable <repository>'
But note that the installed repo file will likely need a manual modification.

```
@msuchy, @FrostyX, what do you think?  Would you add/remove/change something?

plugins/copr.py
```diff
@@ -470,34 +474,28 @@ def _download_repo(self, project_name, repo_filename):
...
-                    raise dnf.exceptions.Error(_("Such repository does not exist."))
-            raise
-
-        for line in f:
-            if re.match(r"\[copr:", line):
-                repo_filename = os.path.join(self.base.conf.get_reposdir,
-                                             "_" + line[1:-2] + ".repo")
-            break
+        except HTTPError as e:
+            error_msg = _("It wasn't possible to enable this project.\n")
```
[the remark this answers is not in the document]
praiskup: This is probably the last remark.  The command produces solid amount of text on stderr/out and it is hard to separate the error message from the boilerplate.

lukash:
Okay, I came with the intention to just fix the messages. Apologies for adding more comments as this PR already seems to have a long history, but in the tested use-cases the new messages seem to be for the most part a change for the worse. Please open the CI test run (click _Details_ next to the failed CI job), then search for "1 failed" in the search box in the top right corner of the dark area with the test log. You should find 9 occurrences of the string in 3 failed test cases, below are just the relevant diffs of the outputs:

1. `dnf/plugins-core/copr.feature:65`:
```
...
      Error: Such repository does not exist.                                                 |  Error: It wasn't possible to enable this project.
                                                                                             |  Such repository does not exist.
```
This one is not really a regression, but I'll note that the phrase _"Such repository does not exist."_ is really confusing to me. "Such repository" in itself is not a great formulation here, but what repository is it even talking about? I'd expect it to tell me which repository and actually also what is the repository's relation to the project?

2. `dnf/plugins-core/copr.feature:84`:
```
...
      Curl error (7): Couldn't connect to server for http://localhost:2/coprs/testuser/testproject/repo/fedora-30/dnf.repo?arch=x86_64 [Failed to connect to localhost port 2: Connection refused]  |  <urlopen error [Errno 111] Connection refused>
```
Thew error is strictly worse than the previous one, it doesn't contain the URL that couldn't be downloaded. In general here the user should get some sort of human-readable sentence telling them the plugin failed to connect to the Copr server.

3. `dnf/plugins-core/copr.feature:114`:
```
...
      Error: This repository does not have any builds yet so you cannot enable it now.       |  Error: It wasn't possible to enable this project.
                                                                                             |  Such repository does not exist.
```
Here the error message changed completely, so I assume it's just wrong? Or is this actually correct and intentional? If yes then again the new messages don't really seem that great...

I also wonder whether there are other cases of errors this PR addresses that would be really good to add as test cases?

FWIW our CI now has an updated README that should be good enough to get the tests running locally without much trouble, though the standard way involves (in this case) building the dnf-plugins-core package and installing it in a container, I can help with a more straightforward dev setup.

schlupov:
1. So output like:
`Repository 'XY' does not exist in project 'YZ'`
would be ok for you?
2. This error is very unlikely, we even use metrics to check that the frontend server is running and copr url is always available.
3. This change was intentional. After I change the output from 1., the change will be reflected here as well.

lukash:
>     1. So output like:
>        `Repository 'XY' does not exist in project 'YZ'`
>        would be ok for you?

Yes, that would be much better :slightly_smiling_face:

>     2. This error is very unlikely, we even use metrics to check that the frontend server is running and copr url is always available.

The likelihood is not really relevant here. I'm sure you have monitoring set up for your servers, but a server can always go down, and there can also be a number of issues on the client machine or the network that will cause slightly different errors which would also be missing information.

>     3. This change was intentional. After I change the output from 1., the change will be reflected here as well.

IIUC one use-case is enabling a non-existent project and the other one is a repository is missing for this particular distribution, right? While you can put these into the same "something not found" bag, it would be better to have two more specific error messages.

praiskup:
ad 3. at least from me:
> While you can put these into the same "something not found" bag, it would be better to have two more specific error messages.

The thing is that such situation doesn't happen, for quite some time ... so having the error message doesn't make sense. From end-user perspective, it is completely valid use-case to enable a project with no RPMs (no build happened, yet, or the build was removed).  A few years back, when user created a new project, Copr did not create the RPM metadata there (createrepo_c run) immediately, but it was waiting for the first build (therefore the old error message was added to `dnf copr enable`, and still it was pretty misleading if you think about it, does the user care about builds? what is a build?).

schlupov:
PTAL, I fixed the integration tests. Note to 2., I started catching the URLError exception when the server is unavailable.

schlupov:
@praiskup @lukash can you please take another look? :)

plugins/copr.py
```diff
@@ -473,34 +477,36 @@ def _download_repo(self, project_name, repo_filename):
...
+                if error_data_decoded.get("available chroots"):
+                    error_msg += _("\nAvailable repositories: ") + ', '.join(
+                        "'{}'".format(x) for x in error_data_decoded["available chroots"])
+                error_msg += _("\n\nIf you want to enable a non-default repository, use the following command:\n"
+                               "  'dnf copr enable {0} <repository>'\n"
+                               "But note that the installed repo file will likely need a manual modification.").format(
+                    project_name)
+                raise dnf.exceptions.Error(error_msg)
+            else:
+                raise dnf.exceptions.Error(error_msg)
```
praiskup, line 500: Ok, I observed a weird behavior now .... even though it is probably not a new thing:
```
$ sudo dnf copr enable non/existing  -y
...
Error: It wasn't possible to enable this project.
Repository 'fedora-35-x86_64' does not exist in project 'non/existing'.
```

The sentence `Repository 'fedora-35-x86_64' does not exist in project 'non/existing'.` could make the user think that the project `non/existing` actually exists, but doesn't have the fedora-35 chroot enabled.

Would you mind special case this, and (when the error_data is None) print something like
`Project 'non/existing' doesn't exist` instead?

lukash: Agreed. Ideally, instead of:
```
Error: It wasn't possible to enable this project.
Repository 'fedora-35-x86_64' does not exist in project 'non/existing'.
```
You would print:
```
Error: Project 'non/existing' doesn't exist.
```

plugins/copr.py
```diff
@@ -475,28 +477,36 @@ def _download_repo(self, project_name, repo_filename):
...
-                else:
-                    raise dnf.exceptions.Error(_("Such repository does not exist."))
-            raise
-
-        for line in f:
-            if re.match(r"\[copr:", line):
-                repo_filename = os.path.join(self.base.conf.get_reposdir,
-                                             "_" + line[1:-2] + ".repo")
-            break
+        except HTTPError as e:
```
lukash, line 483: If I'm not mistaken this exception represents any non-successful HTTP error code. This means you need to check the actual HTTP code, the error handling below I think is only meant for 404 not found. I'm not entirely sure but any other code is just an unexpected error which you should just print like:
```
Error: Request to "http://localhost:2/coprs/testuser/testproject/repo/fedora-30/dnf.repo?arch=x86_64" failed:  STATUS_CODE - SATUS_MESSAGE
```

schlupov:
I've opened this PR https://pagure.io/copr/copr/pull-request/1989 to make this work.
When a project doesn't exist, the ObjectNotFound is returned with a message: "copr dir X/Y does not exist." and I send this message in the header to dnf-plugins-core. We previously showed this message just in the web UI when the user used a nonexistent project.

praiskup:
Looks good to me now.  Thank you!

plugins/copr.py
```diff
@@ -475,28 +477,40 @@ def _download_repo(self, project_name, repo_filename):
...
+            if e.code != 404:
+                error_msg = _("Request to {0} failed: {1} - {2}".format(self.copr_url + api_path, e.code, str(e)))
+                raise dnf.exceptions.Error(error_msg)
+            error_msg = _("It wasn't possible to enable this project.\n")
```
lukash, line 487: I would just delete this line, I find it redundant. But if you decide to keep it, note that on line 503 you overwrite the variable instead of appending, so this sentence is only printed for the `if` branch starting on line 489, which is confusing, if that's desired, you should put it inside the `if`.

lukash (commented):
Two minor comments, otherwise looks good to me, but I'll leave the final call to @praiskup.

Additionally the PR for updating the test scenarios needs to be updated: https://github.com/rpm-software-management/ci-dnf-stack/pull/1051

This scenario is failing now as the message has changed again:
```
  Scenario: Test enabling a repo without any builds for the distribution                                                   # dnf/plugins-core/copr.feature:114
```

Now this scenario is facing a non-trivial issue, as the new code depends on the http server sending a specific response that the plugin code parses. I think it would be highly beneficial to have this tested and it shouldn't be too hard to mock this response in the tests' http server. The test suite currently lacks a step definition to mock a response for a particular URL, so that would need to be added. I can provide some pointers or maybe a rough mockup if needed. Let me know what you think.

schlupov:
@lukash Updated. As I understood it, it looks good to praiskup now.

> I can provide some pointers or maybe a rough mockup if needed.

I would appreciate some hints.

plugins/copr.py
```diff
@@ -473,34 +477,40 @@ def _download_repo(self, project_name, repo_filename):
...
+                    error_msg += _("\n\nIf you want to enable a non-default repository, use the following command:\n"
+                                   "  'dnf copr enable {0} <repository>'\n"
+                                   "But note that the installed repo file will likely need a manual "
+                                   "modification.").format(project_name)
+                raise dnf.exceptions.Error(error_msg)
+            else:
+                error_msg += _("Project {0} does not exist.").format(project_name)
+                raise dnf.exceptions.Error(error_msg)
+        except URLError as e:
+            error_msg = _("Failed to connect to {0}: {1}").format(self.copr_url + api_path, e.reason.strerror)
```
praiskup, line 506: Ad the `if` comment above, ... I think we can calmly use `error_msg +=` here, right?

praiskup:
Otherwise looks good.

plugins/copr.py
schlupov: I wouldn't say that. If we need to handle the URLErrort exception, then line 506 is the first line where the `error_msg` variable is defined.

praiskup: Indeed, I had to look at the whole file (not just diff) to see this.  Then LGTM.

praiskup:
Thank you!

PR #442, opened 2021-10-06
[copr] Silence a deprecation warning

oturpe:
In version 1.6.0, the 'distro' package deprecated linux_distribution().
This causes a deprecation warning to printed to stdout
every time the user calls the copr plugin.
    
In order to avoid the printout
and to protect against possible removal in the future,
the function is reimplemented
using still supported functions from 'distro'.


xsuchy:
Thank you.

PR #443, opened 2021-10-07
copr: allow specifying protocol as part of --hub

t184256:
This way it doesn't try to connect to `https://http//url` if `--hub` started with `http://`.


xsuchy (approved):
LGTM. Sorry, that it took me so long.

PR #444, opened 2021-10-08
Add new command modulesync

j-mracek:
It will download module metadata from all enabled repositories,
module artifacts and profiles of matching modules. Then it creates
a repository.

Requires: https://github.com/rpm-software-management/dnf/pull/1794
For RHEL8 it requires: https://github.com/rpm-software-management/libdnf/pull/1381

### DIFF
--- a/doc/modulesync.rst
+++ b/doc/modulesync.rst
@@ -1,0 +1,103 @@
+..
+  Copyright (C) 2015  Red Hat, Inc.
+
+  This copyrighted material is made available to anyone wishing to use,
+  modify, copy, or redistribute it subject to the terms and conditions of
+  the GNU General Public License v.2, or (at your option) any later version.
+  This program is distributed in the hope that it will be useful, but WITHOUT
+  ANY WARRANTY expressed or implied, including the implied warranties of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+  Public License for more details.  You should have received a copy of the
+  GNU General Public License along with this program; if not, write to the
+  Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+  02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+  source code or documentation are not subject to the GNU General Public
+  License and may only be used or replicated with the express permission of
+  Red Hat, Inc.
+
+====================
+DNF modulesync Plugin
+====================
+
+Download packages from modules and/or create a repository with modular data.
+
+--------
+Synopsis
+--------
+
+``dnf modulesync [options] [<module-spec>...]``
+
+-----------
+Description
+-----------
+
+`modulesync` downloads packages from modules according to provided arguments and creates a repository with modular data
+in working directory. In environment with modules it is recommend to use the command for redistribution of packages,
+because DNF does not allow installation of modular packages without modular metadata on the system (Fail-safe
+mechanism). The command without an argument creates a repository like `createrepo_c` but with modular metadata collected
+from all available repositories.
+
+See examples.
+
+---------
+Arguments
+---------
+
+``<module-spec>``
+    Module specification for the package to download. The argument is an optional.
+
+-------
+Options
+-------
+
+All general DNF options are accepted. Namely, the ``--destdir`` option can be used to specify directory where packages
+will be downloaded and the new repository created. See `Options` in :manpage:`dnf(8)` for details.
+
+
+``-n, --newest-only``
+    Download only packages from the newest modules.
+
+``--enable_source_repos``
+    Enable repositories with source packages
+
+``--enable_debug_repos``
+    Enable repositories with debug-info and debug-source packages
+
+``--resolve``
+    Resolve and download needed dependencies
+
+--------
+Examples
+--------
+
+``dnf modulesync nodejs``
+    Download packages from `nodejs` module and crete a repository with modular metadata in working directory
+
+``dnf download nodejs``
+
+``dnf modulesync``
+    The first `download` command downloads nodejs package into working directory. In environment with modules `nodejs`
+    package can be a modular package therefore when I create a repository I have to insert also modular metadata
+    from available repositories to ensure 100% functionality. Instead of `createrepo_c` use `dnf modulesync`
+    to create a repository in working directory with nodejs package and modular metadata.
+
+``dnf --destdir=/tmp/my-temp modulesync nodejs:14/minimal --resolve``
+    Download package required for installation of `minimal` profile from module `nodejs` and stream `14` into directory
+    `/tmp/my-temp` and all required dependencies. Then it will create a repository in `/tmp/my-temp` directory with
+    previously downloaded packages and modular metadata from all available repositories.
+
+``dnf module install nodejs:14/minimal --downloadonly --destdir=/tmp/my-temp``
+
+``dnf modulesync --destdir=/tmp/my-temp``
+    The first `dnf module install` command downloads package from required for installation of `minimal` profile from module
+    `nodejs` and stream `14` into directory `/tmp/my-temp`. The second command `dnf modulesync` will create
+    a repository in `/tmp/my-temp` directory with previously downloaded packages and modular metadata from all
+    available repositories. In comparison to `dnf --destdir=/tmp/my-temp modulesync nodejs:14/minimal --resolve` it will
+    only download packages required for installation on current system.
+
+
+--------
+See Also
+--------
+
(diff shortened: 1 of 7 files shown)


m-blaha:
I'm not sure how useful is tracking the tool development in separate commits. I think squashing all of them into one commit which introduces the new plugin is cleaner.
But feel free to ignore this rant, it isn't that important.

doc/modulesync.rst
```diff
@@ -0,0 +1,88 @@
...
+
+``dnf modulesync [options] <module-spec>...``
+
+-----------
+Description
+-----------
+
+`modulesync` downloads packages from modules according to provided arguments and creates a repository with modular data
+in working directory. Im modular environment it is recommend to use the command in combination with other dnf commands
+which downloads RPM packages. See examples
```
m-blaha, line 36: I'm a bit confused here. Could you elaborate in more details the use case when user needs to download rpm packages separately (using `dnf download` or such) and then run `modulesync` to generate modular metadata? I thought that `modulesync` will download rpm packages from given modules.

j-mracek:
> I'm not sure how useful is tracking the tool development in separate commits. I think squashing all of them into one commit which introduces the new plugin is cleaner. But feel free to ignore this rant, it isn't that important.

I can do it.

doc/modulesync.rst
j-mracek: In case that a package has modular label in rpm header it cannot be used in new repository without modular metadata. module sync will create a repository including modular metadata.

m-blaha: But modulesync also downloads the rpm package itself, doesn't it? What I do not understand is the need of `dnf download` before `dnf modulesync`. I really think this deserves more detailed documentation in the man pages.

j-mracek: Good point

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+
+``dnf modulesync nodejs``
+    Download packages from `nodejs` module and crete a repository with modular metadata in working directory
+
+``dnf download nodejs``
+``dnf modulesync``
+    The first `download` command downloads nodejs package into working directory. In environment with modules `nodejs`
+    package can be a modular package therefore when I create a repository I have to insert also modular metadata
+    from available repositories to ensure 100% functionality. Instead of `createrepo_c` use `dnf modulesync`
+    to crate a repository in working directory with nodejs package and modular metadata.
```
inknos, line 81: crate, create. typo :)

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+
+``dnf --destdir=/tmp/my-temp modulesync nodejs:14/common --resolve``
```
inknos, line 83: I believe the available profiles are default, development and minimal. Not common

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+    Download package required for installation of `common` profile from module `nodejs` and stream `14` into directory
+    `/tmp/my-temp` and all required dependencies. Then it will crate a repository in `/tmp/my-temp` directory with
+    packages previously downloaded packages and modular metadata from all available repositories.
+
+``dnf module install nodejs:14/common --downloadonly --destdir=/tmp/my-temp``
```
inknos, line 88: nodejs:14/common, like above

inknos (changes requested):

inknos:
Please, could you update the commit message with `=changelog=` and specify the `msg`, `type` and `bug` fileds?

doc/modulesync.rst
j-mracek: This is an example valid for RHEL8.

j-mracek: I can use `minimal` which is less controversial.

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
```
m-blaha, line 17: Copyright here is not correctly formated - resulting in inclusion of copyright into the man page itself. Correct format of the comment in rst file is:
```
..
  Copyright (C) 2021  Red Hat, Inc.

  This copyrighted...
```

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+--------
+
+``dnf modulesync [options] [<module-spec>...]``
+
+-----------
+Description
+-----------
+
+`modulesync` downloads packages from modules according to provided arguments and creates a repository with modular data
+in working directory. Im environment with modules it is recommend to use the command for redistribution of packages,
```
m-blaha, line 35: typo: **In** environment

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+because DNF prevents from getting modular packages without modular metadata on the system (Fail-safe mechanism).
```
m-blaha, line 36: ```because DNF does not allow installation of modular packages without modular metadata...```?
I'm not sure what `getting` word here is supposed to mean.

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+
+--------
+Examples
+--------
+
+``dnf modulesync nodejs``
+    Download packages from `nodejs` module and crete a repository with modular metadata in working directory
+
+``dnf download nodejs``
+``dnf modulesync``
```
m-blaha, line 77: Multiline examples are incorrectly formatted. In the man page they are rendered this way:
```
       dnf download nodejs dnf modulesync
          The first download command ...
```

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+    The first `download` command downloads nodejs package into working directory. In environment with modules `nodejs`
+    package can be a modular package therefore when I create a repository I have to insert also modular metadata
+    from available repositories to ensure 100% functionality. Instead of `createrepo_c` use `dnf modulesync`
+    to create a repository in working directory with nodejs package and modular metadata.
+
+``dnf --destdir=/tmp/my-temp modulesync nodejs:14/minimal --resolve``
+    Download package required for installation of `minimal` profile from module `nodejs` and stream `14` into directory
+    `/tmp/my-temp` and all required dependencies. Then it will create a repository in `/tmp/my-temp` directory with
+    packages previously downloaded packages and modular metadata from all available repositories.
```
m-blaha, line 86: duplicated `packages` word

doc/modulesync.rst
```diff
@@ -0,0 +1,101 @@
...
+
+``dnf module install nodejs:14/minimal --downloadonly --destdir=/tmp/my-temp``
+``dnf modulesync --destdir=/tmp/my-temp``
+    The first `dnf module install` command downloads package from required for installation of `minimal` profile from module
+    `nodejs` and stream `14` into directory `/tmp/my-temp`. The second command `dnf modulesync` will create
+    a repository in `/tmp/my-temp` directory with packages previously downloaded packages and modular metadata from all
```
m-blaha, line 92: Again duplicated `packages` word.

m-blaha (changes requested):
I found some remaining issues with man pages / documentation. Also the modulesync should be included in `index.rst` file.

plugins/modulesync.py
```diff
@@ -0,0 +1,206 @@
...
+                            msg = _("No match for package name '{0}' in profile {1} from module {2}")\
+                                .format(pkg_name, profile.getName(), module.getFullIdentifier())
+                            logger.warning(msg)
+        if not module_list:
+            msg = _("No mach for argument '{}'").format(module_spec)
+            raise dnf.exceptions.Error(msg)
+
+        return result_query
+
+    def _get_providers_of_requires(self, to_test, done=None, req_dict={}):
```
m-blaha, line 171: using mutable default parameter values (here `req_dict={}`) is strongly discouraged and has significant side effects (see https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). Unless you have a reason, please use `req_dict=None`.

j-mracek: Good point.

j-mracek: Fixed

PR #445, opened 2021-10-12
copr: migrate all calls to APIv3

FrostyX:
In the latest Copr release we dropped all APIv1 code from frontend.
https://docs.pagure.org/copr.copr/release-notes/2021-10-01.html

Unfortunatelly we frogot to migrate DNF copr plugin to APIv3 and
therefore the following commands started failing with 404.

    dnf copr search tests
    dnf copr list --available-by-user frostyx

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -356,9 +356,10 @@ Bugzilla. In case of problems, contact the owner of this repository.
                     "Re-enable the project to fix this."))
 
     def _list_user_projects(self, user_name):
-        # http://copr.fedorainfracloud.org/api/coprs/ignatenkobrain/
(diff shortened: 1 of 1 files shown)


praiskup:
Hehe, off-topic: I really hate seeing `General DNF options:` in the `--help` output.  Those are entirely misleading when working with copr.py.

praiskup:
Ok `dnf copr search <keyword>` works again, so +1.

FrostyX:
> dnf copr search --installed pidgin doesn't seem to work (but perhaps unrelated)

I think it's unrelated. There doesn't seem to be any `--installed` specific code, so `dnf copr search --installed pidgin` is IMHO equal to `dnf copr search pidgin`.

PR #447, opened 2021-10-21
[versionlock] fix multi pkg lock (RhBug:2001039)

inknos:
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2013324
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2013332
test https://github.com/rpm-software-management/ci-dnf-stack/pull/1045

### DIFF
--- a/doc/versionlock.rst
+++ b/doc/versionlock.rst
@@ -97,6 +97,10 @@ Subcommands
             Adding versionlock on: mutt-5:1.11.4-1.fc30.*
             Adding versionlock on: mutt-5:1.12.1-3.fc30.*
 
+    .. note:: Be careful when adding specific versions
+
+        If you add a package specifying a version with ``dnf versionlock mutt-5:1.11.4-1.fc30.x86_64`` then, if you run ``dnf versionlock add mutt``
+        versionlock will not add ``mutt-5:1.12.1-3.fc30.x86_64``.
 
 ``dnf versionlock exclude <package-name-spec>``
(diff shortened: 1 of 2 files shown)


inknos:
At this state, the behaviour is the following:
1. `bash` is added:
- `dnf versionlock bash` -> nothing (as before)
- `dnf versionlock bind` -> adds `bind` (as before)
- `dnf versionlock bash bind` -> adds `bind` (new behaviour)
- `dnf versionlock exclude bash` -> does not exclude `bash` (as before)
- `dnf versionlock exclude curl` -> excludes `curl` (as before)
- `dnf versionlock exclude bash curl` -> does not exclude curl because bash is added (as before)
2. `bash` is excluded:
- `dnf versionlock bash` -> does not add `bash`, already excluded (as before)
- `dnf versionlock bind` -> adds `bind` (as before)
- `dnf versionlock bash bind` -> does not add anything because bash is already excluded (as before)
- `dnf versionlock exclude bash` ->  nothing (as before)
- `dnf versionlock exclude bind` -> excludes `bind` (as before)
- `dnf versionlock exclude bind curl` -> excludes `bind` and `curl` (new behaviour)

So, there is still inconsistency for the old versionlock. @m-blaha, what do you think? Now the issue in the bug is addressed.
Do you prefer to make the plugin consistent between add and exclude, or to just make versionlock consistent for a single subcommand?
Since it was inconsistent in the past, my preference is to keep compatibility, so this state would be fine.

PR #448, opened 2021-11-08
groups-manager: More benevolent resolving of packages (RhBug:2013633)

m-blaha:
= changelog =
msg: groups-manager uses for matching packages full NEVRA and not only name.
type: enhancement
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2013633


kontura:
LGTM.

PR #449, opened 2022-01-07
Add counts of missing deps

jcpunk:
For reviewing repoclosure results, it is handy to have raw counts sometimes.  Incremental fixes are easier to identify when you can see the numbers going down.

### DIFF
--- a/tests/test_repoclosure.py
+++ b/tests/test_repoclosure.py
@@ -50,10 +50,10 @@ class TestRepoClosureFunctions(support.TestCase):
                 support.command_run(self.cmd, args)
 
             self.assertEqual(context.exception.value,
-                             "Repoclosure ended with unresolved dependencies.")
+                             "Repoclosure ended with unresolved dependencies (1) across 1 packages.")
 
             expected_out = ["package: foo-4-6.noarch from @commandline",
-                            "  unresolved deps:",
+                            "  unresolved deps (1):",
                             "    bar = 4-6"]
             self.assertEqual(stdout.getvalue()[:-1], "\n".join(expected_out))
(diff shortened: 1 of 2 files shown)


tests/test_repoclosure.py
```diff
@@ -50,7 +50,7 @@ def test_check_option(self):
                 support.command_run(self.cmd, args)
 
             self.assertEqual(context.exception.value,
-                             "Repoclosure ended with unresolved dependencies.")
+                             "Repoclosure ended with unresolved dependencies (1) across 1 packages.")
 
             expected_out = ["package: foo-4-6.noarch from @commandline",
                             "  unresolved deps:",
```
kontura, line 56: The tests are still failing, you have to also add the `(1)` here.
`"  unresolved deps (1):",`

tests/test_repoclosure.py
```diff
@@ -69,7 +69,7 @@ def test_pkg_option(self):
...
```
kontura, line 75: Here as well:
`" unresolved deps (1):",`

tests/test_repoclosure.py
```diff
@@ -89,7 +89,7 @@ def test_base(self):
...
```
kontura, line 95: Here as well:
`" unresolved deps (1):",`

kontura (changes requested):
Could you add a changelog message into the commit (as we have described in the README)?

Perhaps something like:
```
= changelog =
msg: [repoclosure] Print counts of missing dependencies
type: enhancement
```

jcpunk:
I think I'm closer....

kontura (approved):

PR #454, opened 2022-04-26
Skip all non rpm tsi for transaction_action plugins (rhbug:2023652)

j-mracek:
It prevent traceback in output when reason change is in transaction

CI test: https://github.com/rpm-software-management/ci-dnf-stack/pull/1093


inknos:
Thank you, it looks good

PR #455, opened 2022-05-11
repomanage: Use modules only from repo they are handling

kontura:
= changelog =
msg: [repomanage/reposync] Modules are used only when they belong to target repo
type: bugfix
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2072441

Tests: https://github.com/rpm-software-management/ci-dnf-stack/pull/1094

### DIFF
--- a/plugins/repomanage.py
+++ b/plugins/repomanage.py
@@ -72,7 +72,8 @@ class RepoManageCommand(dnf.cli.Command):
         keepnum = int(self.opts.keep) # the number of items to keep
 
         try:
-            repo_conf = self.base.repos.add_new_repo("repomanage_repo", self.base.conf, baseurl=[self.opts.path])
+            REPOMANAGE_REPOID = "repomanage_repo"
+            repo_conf = self.base.repos.add_new_repo(REPOMANAGE_REPOID, self.base.conf, baseurl=[self.opts.path])
             # Always expire the repo, otherwise repomanage could use cached metadata and give identical results
             # for multiple runs even if the actual repo changed in the meantime
             repo_conf._repo.expire()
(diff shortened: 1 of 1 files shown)


plugins/repomanage.py
```diff
@@ -84,9 +84,10 @@ def run(self):
                 module_packages = self.base._moduleContainer.getModulePackages()
 
                 for module_package in module_packages:
-                    all_modular_artifacts.update(module_package.getArtifacts())
-                    module_dict.setdefault(module_package.getNameStream(), {}).setdefault(
-                        module_package.getVersionNum(), []).append(module_package)
+                    if module_package.getRepoID() == "repomanage_repo":
```
j-mracek, line 87: Please can you add a comment that will clarify why you do it?
It is clear that only one available repository is loaded "repoman age_repo", but the workflow also loads Fail-safe data stored on the disk. Please can you point out it somehow?
Please can you also add an information that the issue was caused by loading of failsafe data?

plugins/reposync.py
```diff
@@ -244,13 +244,14 @@ def _get_latest(self, query):
...
         artifact_version = {} # {artifact: {NameStream: [Version]}}
         for module_package in module_packages:
-            artifacts = module_package.getArtifacts()
-            all_artifacts.update(artifacts)
-            module_dict.setdefault(module_package.getNameStream(), {}).setdefault(
-                module_package.getVersionNum(), []).append(module_package)
-            for artifact in artifacts:
-                artifact_version.setdefault(artifact, {}).setdefault(
-                    module_package.getNameStream(), []).append(module_package.getVersionNum())
+            if module_package.getRepoID() == repo_id:
```
j-mracek, line 247: I am not sure whether it is a correct approach for reposync.

```if module_package.getRepoID() != hawkey.MODULE_FAIL_SAFE_REPO_NAME:```

Also the condition requires a comment.

plugins/repomanage.py
kontura: Added. :+1:

plugins/reposync.py
kontura: As we discussed I have dropped the reposync changes because:
- The bz [2072441](https://bugzilla.redhat.com/show_bug.cgi?id=2072441) didn't report that as an issue
- It is not 100% clear what is the correct approach (lets wait if someone reports it)

plugins/repomanage.py
j-mracek: Please can you define `"repomanage_repo"` as local variable? Like `REPOMANAGE_REPONAME`?

j-mracek: Please can you also add to the commit header `
repomanage: Use modules only from repo they are handling ` `RhBug:` reference? Or is it not needed for generation of upstream changelog?

j-mracek: Ok, it is not needed anymore but it is nice to see it in the header.

kontura: Both issues resolved.

PR #458, opened 2022-06-01
copr: Detect Fedora remixes

korewaChino:
Co-authored-by: PizzaLovingNerd <pizzalovingnerd@teknik.io>
= changelog =
msg:           [copr] Add a check for ID_LIKE if it only contains fedora
type:          enhancement
resolves:      https://bugzilla.redhat.com/show_bug.cgi?id=2091662

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -40,7 +40,7 @@ import rpm
 # If that fails, attempt to import the deprecated implementation
 # from the platform module.
 try:
-    from distro import name, version, codename, os_release_attr
+    from distro import name, version, codename, os_release_attr, like
 
     # Re-implement distro.linux_distribution() to avoid a deprecation warning
     def linux_distribution():
@@ -469,6 +469,10 @@ Bugzilla. In case of problems, contact the owner of this repository.
                 chroot = ("opensuse-tumbleweed-{}".format(distarch))
             else:
                 chroot = ("opensuse-leap-{0}-{1}".format(dist[1], distarch))
+        elif "Red Hat Enterprise Linux" in dist:
+            chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
+        elif "fedora" in like() and not any(s in like() for s in ["rhel", "centos"]):
+            chroot = ("fedora-{0}-{1}".format(dist[1], distarch))
         else:
             chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
         return chroot


jadeastar:
Thank you for the fixes! This is incredibly helpful in @tau-OS!

plugins/copr.py
```diff
@@ -64,6 +64,20 @@ def linux_distribution():
...
+                os_release_data = {}
+                for line in os_release_file:
+                    try:
+                        os_release_key, os_release_value = line.rstrip().split('=')
+                        os_release_data[os_release_key] = os_release_value.strip('"')
+                    except ValueError:
+                        # Skip empty lines and everything that is not a simple
+                        # variable assignment
+                        pass
+                return os_release_data['ID_LIKE']
```
lukash, line 80: This is obviously wrong, the `except` execution branch handles the case of failing to import `linux_distribution`, which it then defines. If you wanna do this, best probably to import the `like` separately and then handle a failed import. When was the `like` added and which distros still don't have it?

Also the code above is somewhat complicated and possibly handles cases which are long since not relevant, so it could probably use some simplification (instead of complicating it more by adding to it).

korewaChino: AlmaLinux seems to not have `python3-distro` installed out of the box, so I tried to add a case for it in case of a failed import.

So should I simply ignore this import or add a seperate import for this?

xsuchy: > AlmaLinux seems to not have `python3-distro` installed out of the box, so I tried to add a case for it in case of a failed import.

It is in Requires https://github.com/rpm-software-management/dnf-plugins-core/blob/master/dnf-plugins-core.spec#L149 so installation of dnf-plugins-core should install it. If it is not there fix the requirements in Alma rather than doing workarounds here in a code.

lukash:
@korewaChino instead of reverting the commit, please remove it from the PR branch completely.

korewaChino:
> @korewaChino instead of reverting the commit, please remove it from the PR branch completely.

I have force pushed the branch without the commit. Please verify the changes.

lukash:
LGTM from technical side, I'll leave the verification of correct chroot selection to the Copr team, @xsuchy / @praiskup.

praiskup:
Thank you for the PR, and sorry for delay!

What is the ID_LIKE  content on remixes?  What is the "NAME" in /etc/os-release?  Are there some podman images available, so we can test?

I think I like the change, but the things above ^^^ need to be documented, ideally in-code (examples why we do the if-conditiions, etc.).

plugins/copr.py
```diff
@@ -471,6 +471,10 @@ def _guess_chroot(self):
                 chroot = ("opensuse-leap-{0}-{1}".format(dist[1], distarch))
         elif "CentOS Stream" in dist:
             chroot = ("centos-stream-{0}-{1}".format(dist[1], distarch))
+        elif "Red Hat Enterprise Linux" in dist:
+            chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
```
praiskup, line 475: I'm curuios why to handle RHEL explicitly here, wouldn't the default `else` branch work?

korewaChino:
> What is the ID_LIKE content on remixes? What is the "NAME" in /etc/os-release? Are there some podman images available, so we can test?
Remixes usually just have a `fedora` in ID_LIKE, similar to RHEL.
This is the case in Ultramarine, tauOS and risiOS.

If you would like a podman container, we have one here at `ghcr.io/ultramarine-linux/ultramarine:36`

plugins/copr.py
PizzaLovingNerd: > I'm curuios why to handle RHEL explicitly here, wouldn't the default `else` branch work?

RHEL has Fedora in it's ID_LIKE, so RHEL would get detected as a Fedora remix. However this isn't an issue for RHEL alternatives like Rocky/Alma/etc, because those all contain ID_LIKE=rhel or ID_LIKE=centos which the patch checks for.

Conan-Kudo:
Handling `ID_LIKE` is going to be very messy. For example, all currently developed members of the Mandrake family has `fedora` in its `ID_LIKE` field.

plugins/copr.py
Conan-Kudo: That means that those derivatives have messed up, because they are supposed to have `fedora` in their `ID_LIKE` fields.

Conan-Kudo:
Note that a Remix can also ship a config file to override the COPR plugin's OS detection. This capability was added by me several years ago specifically so that derivatives can declare what the plugin should use.

xsuchy:
> Note that a Remix can also ship a config file to override the COPR plugin's OS detection. This capability was added by me several years ago specifically so that derivatives can declare what the plugin should use.

Unfortunatelly no. Pavel removed that in b124dfefba61b2244ab47f9bfb3a2d54abf85b61

praiskup:
IOW, is it enough to configure
```
[main]                                                                                                                
releasever = 38                                                                                                       
distribution = Fedora   
```
in `/etc/dnf/plugins/copr.conf` in the remix package.

korewaChino:
any updates on this PR?

Conan-Kudo:
I think this will invite a ton of bugs, and I don't think we should merge this. I'd be open to a pull request to add extra configurability in the plugin config file if needed, though.

korewaChino:
> IOW, is it enough to configure
> 
> ```
> [main]                                                                                                                
> releasever = 38                                                                                                       
> distribution = Fedora   
> ```
> 
> in `/etc/dnf/plugins/copr.conf` in the remix package.

Seems like this actually works!

PR #459, opened 2022-06-28
copr: Guess EPEL chroots for CentOS Stream (RhBug:2058471)

carlwgeorge:
Packages built in epel-9 chroots are almost always compatible with CentOS Stream 9.  Not having the copr plugin guess this chroot is causing user friction.  Users are creating epel-9 chroots expecting them to work for both CentOS Stream 9 and RHEL 9.  When they get reports about `dnf copr enable` not working, they try to add a centos-stream-9 chroot, only to discover the dependencies they need from EPEL are not available.
    
Instead of making the majority of CentOS Stream users include an explicit chroot argument, let's reserve that workaround only for the people that don't want their CentOS Stream systems picking the EPEL chroot.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -469,8 +469,6 @@ Bugzilla. In case of problems, contact the owner of this repository.
                 chroot = ("opensuse-tumbleweed-{}".format(distarch))
             else:
                 chroot = ("opensuse-leap-{0}-{1}".format(dist[1], distarch))
-        elif "CentOS Stream" in dist:
-            chroot = ("centos-stream-{0}-{1}".format(dist[1], distarch))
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):
Users basically never want the non-EPEL chroots (which is why we don't use them for any Enterprise Linux distribution), so this makes sense to me.

Conan-Kudo:
Looks like we're having issues with COPR API...

Conan-Kudo:
I merged it as it works locally on my CentOS Stream 9 machines. For future note, if COPR gets support for epel-next chroots, then we should revisit this and make CentOS Stream use those.

PR #461, opened 2022-08-11
builddep: Warning when using macros with source rpms

jan-kolarik:
Defining RPM macros in `builddep` command is not supported for source rpms. Therefore adding note into man page and warning when using the `-D` or `--define` argument with source rpm.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2077820.


kontura:
LGTM
Thanks!

PR #462, opened 2022-08-30
Move system-upgrade plugin to core (RhBug:2054235)

jan-kolarik:
Move `system-upgrade` plugin from `dnf-plugins-extras` to `dnf-plugins-core` to enable offline upgrades on RHEL systems.

Related PRs from other components:

`dnf-plugins-extras`: https://github.com/rpm-software-management/dnf-plugins-extras/pull/210
`dnf`: https://github.com/rpm-software-management/dnf/pull/1846
`ci-dnf-stack`: https://github.com/rpm-software-management/ci-dnf-stack/pull/1150

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2054235.

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -148,14 +148,13 @@ Summary:    Core Plugins for DNF
 %{?python_provide:%python_provide python3-%{name}}
 BuildRequires:  python3-dbus
 BuildRequires:  python3-devel
+BuildRequires:  python3-distro
 BuildRequires:  python3-dnf >= %{dnf_lowest_compatible}
 BuildRequires:  python3-systemd
 BuildRequires:  pkgconfig(systemd)
 BuildRequires:  systemd
 %{?systemd_ordering}
-%if 0%{?fedora}
 Requires:       python3-distro
-%endif
 Requires:       python3-dbus
 Requires:       python3-dnf >= %{dnf_lowest_compatible}
 Requires:       python3-hawkey >= %{hawkey_version}
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -64,6 +64,9 @@ Provides:       dnf-command(repograph)
 Provides:       dnf-command(repomanage)
 Provides:       dnf-command(reposync)
 Provides:       dnf-command(repodiff)
+Provides:       dnf-command(system-upgrade)
+Provides:       dnf-command(offline-upgrade)
+Provides:       dnf-command(offline-distrosync)
 Provides:       dnf-plugins-extras-debug = %{version}-%{release}
 Provides:       dnf-plugins-extras-repoclosure = %{version}-%{release}
 Provides:       dnf-plugins-extras-repograph = %{version}-%{release}
@@ -80,6 +83,7 @@ Provides:       dnf-plugin-repodiff = %{version}-%{release}
 Provides:       dnf-plugin-repograph = %{version}-%{release}
 Provides:       dnf-plugin-repomanage = %{version}-%{release}
 Provides:       dnf-plugin-reposync = %{version}-%{release}
+Provides:       dnf-plugin-system-upgrade = %{version}-%{release}
 %if %{with yumcompatibility}
 Provides:       yum-plugin-copr = %{version}-%{release}
 Provides:       yum-plugin-changelog = %{version}-%{release}
@@ -133,8 +137,8 @@ Conflicts:      python-%{name} < %{version}-%{release}
 %description -n python2-%{name}
 Core Plugins for DNF, Python 2 interface. This package enhances DNF with builddep,
 config-manager, copr, degug, debuginfo-install, download, needs-restarting,
-groups-manager, repoclosure, repograph, repomanage, reposync, changelog
-and repodiff commands.
+groups-manager, repoclosure, repograph, repomanage, reposync, changelog,
+repodiff, system-upgrade, offline-upgrade and offline-distrosync commands.
 Additionally provides generate_completion_cache passive plugin.
 %endif
 
@@ -145,6 +149,10 @@ Summary:    Core Plugins for DNF
 BuildRequires:  python3-dbus
 BuildRequires:  python3-devel
 BuildRequires:  python3-dnf >= %{dnf_lowest_compatible}
+BuildRequires:  python3-systemd
+BuildRequires:  pkgconfig(systemd)
+BuildRequires:  systemd
+%{?systemd_ordering}
 %if 0%{?fedora}
 Requires:       python3-distro
 %endif
@@ -152,14 +160,17 @@ Requires:       python3-dbus
 Requires:       python3-dnf >= %{dnf_lowest_compatible}
 Requires:       python3-hawkey >= %{hawkey_version}
 Requires:       python3-dateutil
+Requires:       python3-systemd
 Provides:       python3-dnf-plugins-extras-debug = %{version}-%{release}
 Provides:       python3-dnf-plugins-extras-repoclosure = %{version}-%{release}
 Provides:       python3-dnf-plugins-extras-repograph = %{version}-%{release}
 Provides:       python3-dnf-plugins-extras-repomanage = %{version}-%{release}
+Provides:       python3-dnf-plugin-system-upgrade = %{version}-%{release}
 Obsoletes:      python3-dnf-plugins-extras-debug < %{dnf_plugins_extra}
 Obsoletes:      python3-dnf-plugins-extras-repoclosure < %{dnf_plugins_extra}
 Obsoletes:      python3-dnf-plugins-extras-repograph < %{dnf_plugins_extra}
 Obsoletes:      python3-dnf-plugins-extras-repomanage < %{dnf_plugins_extra}
+Obsoletes:      python3-dnf-plugin-system-upgrade < %{version}-%{release}
 
 Conflicts:      %{name} <= 0.1.5
 # let the both python plugin versions be updated simultaneously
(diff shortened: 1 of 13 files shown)


jan-kolarik:
Related PRs from other components:
- `dnf-plugins-extras`: https://github.com/rpm-software-management/dnf-plugins-extras/pull/210
- `dnf`: https://github.com/rpm-software-management/dnf/pull/1846
- `ci-dnf-stack`: https://github.com/rpm-software-management/ci-dnf-stack/pull/1150

jan-kolarik:
I tested `dnf-plugins-core` and `dnf-plugins-extras` PR changes locally.

RPMs were built using RH Copr within https://copr.devel.redhat.com/coprs/jkolarik/offline-updates/ project and deployed on RHEL9 VM. `offline-upgrade` and `offline-distsync` scenarios were tested as desribed in related bz ticket.

CI tests from `ci-dnf-stack` tested in local container with integration tests passing.

dnf-plugins-core.spec
```diff
@@ -415,6 +415,29 @@ Download module metadata from all enabled repositories, module artifacts and pro
...
 
+%if %{with python3}
+%package -n python3-dnf-plugin-system-upgrade
+Summary:        System Upgrade Plugin for DNF
+Requires:       python3-systemd
+Provides:       dnf-command(system-upgrade)
+Provides:       dnf-command(offline-upgrade)
+Provides:       dnf-command(offline-distrosync)
+Provides:       dnf-plugin-system-upgrade = %{version}-%{release}
+Provides:       python3-dnf-plugins-extras-system-upgrade = %{version}-%{release}
```
inknos, line 426: Shouldn't this be `python3-dnf-plugins-core-system-upgrade = %{version}-%{release}`?

inknos:
The PR and the related ones LGTM. I will not merge it as we discussed today and wait for the decision about RHEL system upgrade plugin. Thank you!

dnf-plugins-core.spec
```diff
@@ -415,6 +415,29 @@ Download module metadata from all enabled repositories, module artifacts and pro
...
+Conflicts:      python2-dnf-plugin-system-upgrade < %{version}-%{release}
+BuildRequires:  pkgconfig(systemd)
+BuildRequires:  systemd
+BuildRequires:  python3-systemd
+%{?systemd_requires}
+
+%description -n python3-dnf-plugin-system-upgrade
+System Upgrade Plugin for DNF. Enables offline system upgrades and distrosync
+using three commands: ``system-upgrade``, ``offline-upgrade``, and ``offline-distrosync``.
+%endif
```
Conan-Kudo, line 439: Is there a reason we should have this split out rather than part of the main dnf-plugins-core package like the rest of the plugins?

Conan-Kudo (changes requested):

dnf-plugins-core.spec
jan-kolarik: Hello, I was following the latest plugins added to core like `modulesync` that are also located in separate packages. Are there any advantages now having it in the main package? I thought the main package is there mainly because different approach was used back then and for backward compatibility of the original commands from core.

jan-kolarik: We will have a discussion regarding the moving into the core package across the teams and I'll let you know.

dnf-plugins-core.spec
```diff
@@ -145,21 +149,28 @@ Summary:    Core Plugins for DNF
 BuildRequires:  python3-dbus
 BuildRequires:  python3-devel
 BuildRequires:  python3-dnf >= %{dnf_lowest_compatible}
+BuildRequires:  python3-systemd
+BuildRequires:  pkgconfig(systemd)
+BuildRequires:  systemd
+%{?systemd_requires}
```
Conan-Kudo, line 155: This should be changed to `%{?systemd_ordering}` instead, so that we don't have a hard dep on systemd at runtime.

Conan-Kudo (changes requested):

dnf-plugins-core.spec
jan-kolarik: Thanks, updated.

Conan-Kudo (approved):

jan-kolarik:
@Conan-Kudo Thanks for the approval. I realized I still have to add a runtime warning when using `system-upgrade` command on non-Fedora systems. Therefore I am switching to draft until this is done.

Conan-Kudo:
> @Conan-Kudo Thanks for the approval. I realized I still have to add a runtime warning when using `system-upgrade` command on non-Fedora systems. Therefore I am switching to draft until this is done.

Wait, what? Why? It works fine on non-Fedora systems.

jan-kolarik:
This is officially supported only for Fedora systems as written in the manual. It doesn't work for RHEL and CentOS for which the warning will be implemented. Sorry for my inaccurate response.

Conan-Kudo:
Please only put the guard for RHEL for now if you must put one at all. There's a community goal to make DNF system upgrade work with CentOS Stream. Linux distributions outside of the Red Hat family using DNF also use the system-upgrade plugin just fine (Mageia and OpenMandriva both document using this for upgrades).

cc: @davide125, @michel-slm

j-mracek:
@Conan-Kudo Thank you very much for your feedback.

doc/system-upgrade.rst
```diff
@@ -0,0 +1,207 @@
...
+
+-----------
+Description
+-----------
+
+DNF system-upgrades plugin provides three commands: ``system-upgrade``, ``offline-upgrade``, and
+``offline-distrosync``. Only ``system-upgrade`` command requires increase of distribution major
+version (``--releasever``) compared to installed version.
+
+``dnf system-upgrade`` can be used to upgrade a Fedora system to a new major
```
Conan-Kudo, line 30: Drop the word `Fedora` here, this tool is used in more than Fedora.

doc/system-upgrade.rst
```diff
@@ -0,0 +1,207 @@
...
+
+Bugs should be filed here:
+
+  https://bugzilla.redhat.com/
+
+For more info on filing bugs, see the Fedora Project wiki:
+
+  https://fedoraproject.org/wiki/How_to_file_a_bug_report
+
+  https://fedoraproject.org/wiki/Bugs_and_feature_requests
```
Conan-Kudo, line 175: Don't we have DNF specific documentation to reference here instead?

Conan-Kudo (changes requested):

jan-kolarik:
> Please only put the guard for RHEL for now if you must put one at all. There's a community goal to make DNF system upgrade work with CentOS Stream. Linux distributions outside of the Red Hat family using DNF also use the system-upgrade plugin just fine (Mageia and OpenMandriva both document using this for upgrades).
> 
> cc: @davide125, @michel-slm

After discussion in the team I am adding a warning for any RHEL-based distribution, so we're safe from any potential customer issues. Anyway, it's just a displayed warning, if anybody really wants to proceed, they can :slightly_smiling_face:

plugins/system_upgrade.py
```diff
@@ -0,0 +1,713 @@
...
+        print(_("Screen blanking can't be disabled: %s") % e)
+
+
+def is_rhel_system_present():
+    ts = rpm.TransactionSet()
+    pkg = next(ts.dbMatch('provides', 'system-release'), None)
+    if not pkg:
+        logger.debug(_("No package providing 'system-release' was found."))
+        return False
+    return 'redhat-release' in pkg['provides']
```
Conan-Kudo, line 132: This check does not work as almost all Red Hat Linux derivatives would match on this. That includes Fedora, OpenMandriva, ROSA, and Mageia. You'll need to check via the `os-release` file instead. RHEL ships the `distro` Python module for this purpose.

Conan-Kudo (changes requested):

plugins/system_upgrade.py
jan-kolarik: Thanks for the quick response. Why do you think so? 

AFAIK there should be a single installed package providing the `system-release` and this one provides the `*-release` specific for given distribution. F.e. on my Fedora 36 I have installed the `fedora-release-workstation` providing the `system-release` and this package provides also the `fedora-release` + also there is none package providing `redhat-release`.

Conan-Kudo: I guess it was removed from most `fedora-release-*` variants, but it exists in `mageia-release-Default`, `distro-release-OpenMandriva`, `rosa-release`, and others.

j-mracek: Technically RHEL is a derivative of Fedora. Using redhat-release in any systems derived form Fedora is not technically correct and it is confusing. I also don't think that parsing any file on the system can help because each distribution can crate the same mess in file like with package provides. On the other hand it is only a warning.

doc/system-upgrade.rst
jan-kolarik: The DNF wiki was dropped and it's not going to be redeployed as our goal is to move to DNF5. Currently there is no living DNF-specific link related to bugs filing as I know, so I'll keep the existing doc state copied from `dnf-plugins-extras`.

plugins/system_upgrade.py
jan-kolarik: I realized that the `distro` module is already present in the `dnf-plugins-core`, therefore switching to this approach as there are no cons I can think of now.

plugins/system_upgrade.py
```diff
@@ -451,6 +457,9 @@ def pre_configure_clean(self):
 
     def configure_download(self):
         if 'system-upgrade' == self.opts.command or 'fedup' == self.opts.command:
+            if is_rhel_system_present():
```
Conan-Kudo, line 460: We can get rid of the function and just use `if distro.id() == 'rhel':` here.

Conan-Kudo (changes requested):

plugins/system_upgrade.py
jan-kolarik: Yep, sure. Hopefully there are no more `RHEL` conditions in the future :)

Conan-Kudo (approved):

PR #463, opened 2022-09-09
copr: Acknowledge downstream compatibility for non-EPEL distros

korewaChino:
= changelog =
msg: [copr] Also read chroot configs from other files in copr.conf.d
type: enhancement
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2091662

This PR is an alternative solution to #458, which allows other config files to override the main Copr config for chroot configs in the `[main]` section.

It adds a new optional config file called `copr.vendor.conf` that allows vendors to specify what distro their downstream is compatible with on COPR. This solves issues regarding COPR on Ultramarine Linux, Nobara, and risiOS.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -138,17 +138,15 @@ class CoprCommand(dnf.cli.Command):
         config_path = self.base.conf.pluginconfpath[0]
 
         default_config_file = os.path.join(config_path, PLUGIN_CONF + ".conf")
+        # would use appdirs, but that would mean a new dependency
+        vendor_config_file = os.path.join("/usr/share/dnf/plugins/", PLUGIN_CONF + ".vendor.conf")
+
         if os.path.isfile(default_config_file):
             config_files.append(default_config_file)
 
-            copr_plugin_config.read(default_config_file)
-            if copr_plugin_config.has_option('main', 'distribution') and\
-                    copr_plugin_config.has_option('main', 'releasever'):
-                distribution = copr_plugin_config.get('main', 'distribution')
-                releasever = copr_plugin_config.get('main', 'releasever')
-                self.chroot_config = [distribution, releasever]
-            else:
-                self.chroot_config = [False, False]
+        if os.path.isfile(vendor_config_file):
+            self._get_copr_chroot(vendor_config_file)
+        self._get_copr_chroot(default_config_file)
 
         for filename in os.listdir(os.path.join(config_path, PLUGIN_CONF + ".d")):
             if filename.endswith('.conf'):
@@ -210,6 +208,25 @@ class CoprCommand(dnf.cli.Command):
             return config.get(hub, section)
         except (NoOptionError, NoSectionError):
             return default
+
+    def _get_copr_chroot(self, config):
+        """
+        Sets the Copr root from a config file
+        This is refactored from configure() to avoid code duplication
+        """
+        copr_plugin_config = ConfigParser()
+        if not os.path.isfile(config):
+            return
+        copr_plugin_config.read(config)
+        if copr_plugin_config.has_option('main', 'distribution') and\
+                copr_plugin_config.has_option('main', 'releasever'):
+            distribution = copr_plugin_config.get('main', 'distribution')
+            releasever = copr_plugin_config.get('main', 'releasever')
+            self.chroot_config = [distribution, releasever]
+        else:
+            # check if the chroot_config is already set from a previous call
+            if not self.chroot_config:
+                self.chroot_config = [False, False]
 
     def _user_warning_before_prompt(self, text):
         sys.stderr.write("{0}\n".format(text.strip()))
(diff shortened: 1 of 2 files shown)


PizzaLovingNerd:
This seems like a much better solution than adding the Fedora detection at the risk of breaking RHEL clones.

korewaChino:
Any update on this?

plugins/copr.py
```diff
@@ -179,6 +170,16 @@ def configure(self):
         else:
             copr_hub = self.opts.hub
 
+        # Check distribution and releasever in all config files
+        if config_files:
+
+            if len(config_files) > 1:
+                self._get_copr_chroot(config_files[1])
```
Conan-Kudo, line 177: Doesn't this mean that if there are more than one drop-in file, only the first one will be read?

Conan-Kudo (changes requested):

plugins/copr.py
korewaChino: Yeah. Since I think there would only be 1 file with the branding anyway.

korewaChino: Actually, I just realized that this could go very wrong. will fix

korewaChino:
I think I've come up with a better solution than loading the files by sort order.

So instead of loading the files in `copr.conf.d` we could have a separate file called `copr.vendor.conf` that would actually contain the vendor ID, so any Fedora remix can have a branding package that provides this file and contains the necessary ID for COPR.

plugins/copr.py
```diff
@@ -179,6 +171,15 @@ def configure(self):
         else:
             copr_hub = self.opts.hub
 
+        # Check if a copr.vendor.conf file exists
+        vendor_config_file = os.path.join(config_path, PLUGIN_CONF + ".vendor.conf")
+
+        # if exists, read it
+        if vendor_config_file and os.path.isfile(vendor_config_file):
```
xsuchy, line 178: `vendor_config_file` will always be True. Can be removed.

plugins/copr.py
```diff
@@ -141,15 +142,6 @@ def configure(self):
...
 
-            copr_plugin_config.read(default_config_file)
-            if copr_plugin_config.has_option('main', 'distribution') and\
-                    copr_plugin_config.has_option('main', 'releasever'):
-                distribution = copr_plugin_config.get('main', 'distribution')
-                releasever = copr_plugin_config.get('main', 'releasever')
-                self.chroot_config = [distribution, releasever]
-            else:
-                self.chroot_config = [False, False]
-
```
xsuchy, line 152: Hu? So the ` default_config_file` is now not read at all? I think you missed calling `self._get_copr_chroot(default_config_file)`

plugins/copr.py
```diff
@@ -179,6 +171,15 @@ def configure(self):
...
             copr_hub = self.opts.hub
 
+        # Check if a copr.vendor.conf file exists
+        vendor_config_file = os.path.join(config_path, PLUGIN_CONF + ".vendor.conf")
+
+        # if exists, read it
+        if os.path.isfile(vendor_config_file):
+            self._get_copr_chroot(vendor_config_file)
+        else:
+            self.chroot_config = [False, False]
```
xsuchy, line 181: I think that this if/else can be moved inside `_get_copr_chroot()` too.

xsuchy:
I like the general approach. 
Can you please document the change in behaviour in https://github.com/rpm-software-management/dnf-plugins-core/blob/master/doc/copr.rst

Conan-Kudo:
This should read `copr.vendor.conf` OR `copr.conf`. `copr.vendor.conf` should be in `/usr/share/dnf/plugins`.

korewaChino:
> This should read `copr.vendor.conf` OR `copr.conf`. `copr.vendor.conf` should be in `/usr/share/dnf/plugins`. In other words, this config file should not be edited by users.

Should I make it read the vendor config first or read the default config first?

Conan-Kudo (approved):

Conan-Kudo (changes requested):
This looks great, can you please squash the commits down and update the commit message to match what this does?

plugins/copr.py
```diff
@@ -138,17 +139,16 @@ def configure(self):
...
-                    copr_plugin_config.has_option('main', 'releasever'):
-                distribution = copr_plugin_config.get('main', 'distribution')
-                releasever = copr_plugin_config.get('main', 'releasever')
-                self.chroot_config = [distribution, releasever]
-            else:
-                self.chroot_config = [False, False]
+        if os.path.isfile(vendor_config_file):
+            self._get_copr_chroot(vendor_config_file)
+        else:
+            self._get_copr_chroot(default_config_file)
```
xsuchy, line 151: This is wrong. You read vendor_config OR default_config. You should read both. It should be:
```
        if os.path.isfile(vendor_config_file):
            self._get_copr_chroot(vendor_config_file)
        self._get_copr_chroot(default_config_file)
```

korewaChino:
I'm sorry, I managed to screw up the squashing somehow and I managed to reset the entire thing :P

xsuchy (approved):
LGTM Great work.
I will leave it open so others have a chance to step in. Unless somebody raised an objection, I will merge this on Monday.

Conan-Kudo (approved):

Conan-Kudo:
LGTM!

PR #464, opened 2022-09-16
download plugin: respect install_weak_deps option value

dex6:
Currently, `dnf download --resolve` will always download weak dependencies, disregarding the `install_weak_deps` option value: _(tested in AlmaLinux 9.0 container)_

```
$ dnf download --setopt=install_weak_deps=0 --url --alldeps --resolve kernel | wc -l
121
$ dnf download --setopt=install_weak_deps=1 --url --alldeps --resolve kernel | wc -l
121
```

With this trivial patch it works as user may expect it to work - not downloading linux-firmware and few other packages:

```
$ dnf download --setopt=install_weak_deps=0 --url --alldeps --resolve kernel | wc -l
116
$ dnf download --setopt=install_weak_deps=1 --url --alldeps --resolve kernel | wc -l
121
```

### DIFF
--- a/plugins/download.py
+++ b/plugins/download.py
@@ -253,7 +253,7 @@ class DownloadCommand(dnf.cli.Command):
         for pkg in pkgs:
             goal = hawkey.Goal(self.base.sack)
             goal.install(pkg)
-            rc = goal.run()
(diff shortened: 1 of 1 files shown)


j-mracek:
LGTM

PR #465, opened 2022-10-21
offline-upgrade: add support for security filters (RhBug:1939975)

TarcisioLOliveira:
Adding security filter support makes it easier and safer, for example, to upgrade select packages from `updates-testing` and other testing repositories. I stumbled into this while testing updates from Bodhi, and also found a bug report about it from last year [(RhBug:1939975)](https://bugzilla.redhat.com/show_bug.cgi?id=1939975).

It's a two-line change, enabling it only for `offline-upgrade`. I tested it by refreshing the metadata using the four available filter options (`--advisory`, `--bugfix`, `--security`, and `--enhancement`) and it did filter them correctly. I downloaded a single package using `--advisory` and it installed correctly on boot. Other non-upgraded packages where left unchanged, and the package is working correctly.

I was not sure if I should update the documentation or not so I just committed the source changes, but I can update it if needed.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,6 +32,7 @@ DNF-PLUGINS-CORE CONTRIBUTORS
     Neal Gompa <ngompa13@gmail.com>
     Paul Howarth <paul@city-fan.org>
     Rickard Dybeck <r.dybeck@gmail.com>
+    Tarcísio Ladeia de Oliveira <wyrquill@gmail.com>
     Tomas Babej <tomasbabej@gmail.com>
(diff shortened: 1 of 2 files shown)


jan-kolarik:
Hello @TarcisioLOliveira and thanks for your contribution! :slightly_smiling_face: I was looking into it and testing it locally and it looks fine. I will also add some related tests into our `ci-dnf-stack` repository. 

The documentation of the changes in the `dnf` component I will done by myself as I noticed that some security options are incorrectly described there.

PR #466, opened 2022-10-30
Fix boot time calculation.

lonicerae:
* The timestamp of reading '/proc/1' comes from the kernel boot time, however this timestamp can be interpreted into different time depending on the BIOS/UEFI setting and the timezone of the system.
  * Update the method of boot time calculation.

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -34,6 +34,7 @@ import functools
 import os
 import re
 import stat
+import time
 
 
 # For which package updates we should recommend a reboot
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Hi @lonicerae, thanks for your interest in the DNF project. Unfortunately, I've tested the proposed solution on the `systemd-nspawn` container and it is not working there. This would break the existing functionality for the users that already requested this feature in the past. See also this related PR: https://github.com/rpm-software-management/dnf-plugins-core/pull/432.

jan-kolarik:
Sure! I followed the instructions from the description of this bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1913962. In the first comment there are detailed steps how to reproduce the bug which is also happening with this patch. There is also a related Fedora bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1914251 for that issue.

jan-kolarik:
There is also a [Fedora Bugzilla request](https://bugzilla.redhat.com/show_bug.cgi?id=2137935) matching this PR. @lonicerae Please could you try applying the proposed change from the [Comment 12](https://bugzilla.redhat.com/show_bug.cgi?id=2137935#c12)? This should make things working for both containers and systems without RTC where the current solution is not working.

jan-kolarik:
I am closing this as there is a [follow-up](https://github.com/rpm-software-management/dnf-plugins-core/pull/468) which is trying to fix the above-mentioned problem.

PR #467, opened 2022-11-01
[system-upgrade] Add --poweroff option to reboot subcommand

cam-rod:
Adds an option to shutdown the system after updates have been installed, instead of rebooting.

[Based on this comment](https://unix.stackexchange.com/a/704106/533888).

### DIFF
--- a/plugins/system_upgrade.py
+++ b/plugins/system_upgrade.py
@@ -62,16 +62,19 @@ DOWNLOAD_FINISHED_MSG = _(  # Translators: do not change "reboot" here
 CANT_RESET_RELEASEVER = _(
     "Sorry, you need to use 'download --releasever' instead of '--network'")
 
-STATE_VERSION = 2
+STATE_VERSION = 3
 
 # --- Miscellaneous helper functions ------------------------------------------
 
 
-def reboot():
+def reboot(poweroff = False):
     if os.getenv("DNF_SYSTEM_UPGRADE_NO_REBOOT", default=False):
         logger.info(_("Reboot turned off, not rebooting."))
     else:
-        Popen(["systemctl", "reboot"])
+        if poweroff:
+            Popen(["systemctl", "poweroff"])
+        else:
+            Popen(["systemctl", "reboot"])
 
 
 def get_url_from_os_release():
@@ -184,6 +187,7 @@ class State(object):
     upgrade_status = _prop("upgrade_status")
     upgrade_command = _prop("upgrade_command")
     distro_sync = _prop("distro_sync")
+    poweroff_after = _prop("poweroff_after")
     enable_disable_repos = _prop("enable_disable_repos")
     module_platform_id = _prop("module_platform_id")
 
@@ -360,6 +364,10 @@ class SystemUpgradeCommand(dnf.cli.Command):
                             action='store_false',
                             help=_("keep installed packages if the new "
                                    "release's version is older"))
+        parser.add_argument('--poweroff', dest='poweroff_after',
+                            action='store_true',
+                            help=_("power off system after the operation "
+                                   "is completed"))
         parser.add_argument('tid', nargs=1, choices=CMDS,
                             metavar="[%s]" % "|".join(CMDS))
         parser.add_argument('--number', type=int, help=_('which logs to show'))
@@ -568,8 +576,13 @@ class SystemUpgradeCommand(dnf.cli.Command):
         if not self.opts.tid[0] == "reboot":
             return
 
+        self.state.poweroff_after = self.opts.poweroff_after
+
         self.log_status(_("Rebooting to perform upgrade."),
                         REBOOT_REQUESTED_ID)
+
+        # Explicit write since __exit__ doesn't seem to get called when rebooting
+        self.state.write()
         reboot()
 
     def run_download(self):
@@ -688,12 +701,15 @@ class SystemUpgradeCommand(dnf.cli.Command):
         self.log_status(_("Download finished."), DOWNLOAD_FINISHED_ID)
 
     def transaction_upgrade(self):
-        Plymouth.message(_("Upgrade complete! Cleaning up and rebooting..."))
-        self.log_status(_("Upgrade complete! Cleaning up and rebooting..."),
+        power_op = "powering off" if self.state.poweroff_after else "rebooting"
+
+        Plymouth.message(_("Upgrade complete! Cleaning up and " + power_op + "..."))
+        self.log_status(_("Upgrade complete! Cleaning up and " + power_op + "..."),
                         UPGRADE_FINISHED_ID)
+
         self.run_clean()
         if self.opts.tid[0] == "upgrade":
-            reboot()
+            reboot(self.state.poweroff_after)
 
 
 class OfflineUpgradeCommand(SystemUpgradeCommand):
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,6 +20,7 @@ DNF-PLUGINS-CORE CONTRIBUTORS
     Adam Salih <salih.max@gmail.com>
     Alexander Todorov <atodorov@otb.bg>
     Anders Blomdell <anders.blomdell@gmail.com>
+    Cameron Rodriguez <rod.cam2014+dev@gmail.com>
     Cyril Jouve <jv.cyril@gmail.com>
     David Michael <fedora.dm0@gmail.com>
     François Rigault <francois.rigault@amadeus.com>
--- a/doc/system-upgrade.rst
+++ b/doc/system-upgrade.rst
@@ -44,6 +44,8 @@ Synopsis
 
 ``dnf system-upgrade reboot``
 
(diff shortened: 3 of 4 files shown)


plugins/system_upgrade.py
```diff
@@ -568,6 +577,9 @@ def run_reboot(self):
         if not self.opts.tid[0] == "reboot":
             return
 
+        if self.opts.poweroff_after is not None:
```
jan-kolarik, line 580: We could probably move it inside the `transaction_download` method aside the other states setters.

plugins/system_upgrade.py
```diff
@@ -598,6 +610,9 @@ def run_upgrade(self):
         # change the upgrade status (so we can detect crashed upgrades later)
         command = ''
         with self.state as state:
+            if self.opts.poweroff_after is not None:
```
jan-kolarik, line 613: After resolving the comment above, this is not needed anymore...

plugins/system_upgrade.py
```diff
@@ -62,16 +62,20 @@
...
     "Sorry, you need to use 'download --releasever' instead of '--network'")
 
-STATE_VERSION = 2
+STATE_VERSION = 3
 
 # --- Miscellaneous helper functions ------------------------------------------
 
 
-def reboot():
+def reboot(poweroff_after = None):
```
jan-kolarik, line 70: I would use just `poweroff` here. The `_after` suffix is not relevant in this context.

jan-kolarik: Also we could use just the default = False.

plugins/system_upgrade.py
```diff
@@ -62,16 +62,20 @@
...
     if os.getenv("DNF_SYSTEM_UPGRADE_NO_REBOOT", default=False):
         logger.info(_("Reboot turned off, not rebooting."))
     else:
-        Popen(["systemctl", "reboot"])
+        if poweroff_after is True:
```
jan-kolarik, line 74: After the above comments are taken into account this could be simplified to `if poweroff:`

plugins/system_upgrade.py
```diff
@@ -62,16 +62,20 @@
...
+            Popen(["systemctl", "poweroff"])
+        else:
+            Popen(["systemctl", "reboot"])
+
```
jan-kolarik, line 80: We could drop this extra newline.

plugins/system_upgrade.py
jan-kolarik, line 580: Simplify to `if self.opts.poweroff:`

jan-kolarik, line 580: You could probably drop this line and always assign the value from `opts` to `state`, right?

plugins/system_upgrade.py
```diff
@@ -598,6 +610,9 @@ def run_upgrade(self):
         # change the upgrade status (so we can detect crashed upgrades later)
         command = ''
         with self.state as state:
+            if self.opts.poweroff_after is not None:
+                state.poweroff_after = self.opts.poweroff_after
+
```
jan-kolarik, line 615: This whole added block is probably unwanted if I am not mistaken. When `reboot` is used with the `--poweroff` parameter, the `poweroff_after` value is saved into the state and then reboot is called. After that the state is loaded from the file and the `upgrade` here is called having the value already set.

plugins/system_upgrade.py
```diff
@@ -688,12 +703,18 @@ def transaction_download(self):
         self.log_status(_("Download finished."), DOWNLOAD_FINISHED_ID)
 
     def transaction_upgrade(self):
-        Plymouth.message(_("Upgrade complete! Cleaning up and rebooting..."))
-        self.log_status(_("Upgrade complete! Cleaning up and rebooting..."),
-                        UPGRADE_FINISHED_ID)
+        if self.state.poweroff_after is True:
```
jan-kolarik, line 706: Again simplify to `if self.state.poweroff_after:`

plugins/system_upgrade.py
```diff
@@ -688,12 +703,18 @@ def transaction_download(self):
...
+            Plymouth.message(_("Upgrade complete! Cleaning up and powering off..."))
+            self.log_status(_("Upgrade complete! Cleaning up and powering off..."),
+                            UPGRADE_FINISHED_ID)
+        else:
+            Plymouth.message(_("Upgrade complete! Cleaning up and rebooting..."))
+            self.log_status(_("Upgrade complete! Cleaning up and rebooting..."),
+                            UPGRADE_FINISHED_ID)
+
```
jan-kolarik, line 714: I would set just the message inside the above `if` block matching the reboot / poweroff intention. Then we could use the `Plymouth.message` and `self.log_status` call just once passing the message variable.

jan-kolarik (changes requested):
Hello @cam-rod and thanks for your interest in the DNF project! The proposal looks good to me in overall. I left some comments, mainly some small code refactoring. And also, did you have a chance to test the behavior, f.e. in the virtual machine?

cam-rod:
Thanks for the feedback, I'll have a look.

> And also, did you have a chance to test the behavior, f.e. in the virtual machine?

No, I'm not really familiar with DNF's plugin system, and I couldn't figure out how to run the plugin after I built it. Is there something not mentioned in the README that I should also do (or something I missed 😁)?

plugins/system_upgrade.py
cam-rod: Right, this can be taken out.

plugins/system_upgrade.py
cam-rod: Yep, misunderstood the `store_true` action, fixed.

jan-kolarik (approved):
Great, looks good to me now!

cam-rod:
Thanks, was able to get it installed, but the poweroff flag isn't working, I'll look into it a bit more.

cam-rod:
Hey, would you be able to test the code? I've tried both of the above suggestions, but I couldn't seem to get it to consistently run my code when I tried the offline reboot (ex. if I modified the journald messages that it wrote, it would still print the original). There's a decent chance the code works, I just can't get my system config to run it.

plugins/system_upgrade.py
```diff
@@ -568,6 +576,8 @@ def run_reboot(self):
         if not self.opts.tid[0] == "reboot":
             return
 
+        self.state.poweroff_after = self.opts.poweroff_after
+
         self.log_status(_("Rebooting to perform upgrade."),
                         REBOOT_REQUESTED_ID)
         reboot()
```
jan-kolarik, line 583: I was testing the functionality today and it seems there is a problem with serializing the `State` object inside the `__exit__` method, because it is not executed when triggering the reboot using the `Popen`.

As this is the only place where we want to serialize it explicitly, we can call the `self.state.write()` on this line. I was testing that and it seems working.

cam-rod: Yep, added that line and it builds now, thanks. Also added a check before writing because the unit tests were trying to write with the MagicMock object.

jan-kolarik: Hi and thanks for testing this! I would prefer not adding any test-related workarounds inside the execution code. We can avoid the error in the unit tests simply by putting the `self.command.opts.poweroff_after = False` inside the `test_run_reboot` test. Could you please update the PR?

Also, probably the last thing, could you add a comment to the `self.state.write()` line, to know why we are doing that at that place. Something like that we need to save the state explicitly here, because the `__exit__` hook is not called when triggering the reboot. Thanks!

cam-rod: Updated and pushed

jan-kolarik:
Great, thanks @cam-rod! I am shipping that into master :ocean:  :ship: :ocean:

cam-rod:
Awesome, thanks @jan-kolarik for your help and patience!

plugins/system_upgrade.py
```diff
@@ -688,12 +701,15 @@ def transaction_download(self):
         self.log_status(_("Download finished."), DOWNLOAD_FINISHED_ID)
 
     def transaction_upgrade(self):
-        Plymouth.message(_("Upgrade complete! Cleaning up and rebooting..."))
-        self.log_status(_("Upgrade complete! Cleaning up and rebooting..."),
+        power_op = "powering off" if self.state.poweroff_after else "rebooting"
+
+        Plymouth.message(_("Upgrade complete! Cleaning up and " + power_op + "..."))
+        self.log_status(_("Upgrade complete! Cleaning up and " + power_op + "..."),
```
piotrdrag, line 707: This change makes the string untranslatable. gettext produces `Upgrade complete! Cleaning up and ` and both `powering off` and `rebooting` are not translated. Just make it separate strings: https://wiki.gnome.org/TranslationProject/DevGuidelines/Never%20split%20sentences

cam-rod: Sorry, didn't understand the functionality of the strings at the time. Submitted a fix in #471

PR #468, opened 2022-11-16
needs_restarting: Fix boot time derivation for systems without RTC

utoddl:
This PR contains the implementation of the idea from Fedora bugzilla [Comment #12](https://bugzilla.redhat.com/show_bug.cgi?id=2137935#c12). See discussion also https://github.com/rpm-software-management/dnf-plugins-core/pull/466#issuecomment-1306983291

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -34,6 +34,7 @@ import functools
 import os
 import re
 import stat
+import time
 
 
 # For which package updates we should recommend a reboot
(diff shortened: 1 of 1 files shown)


jan-kolarik:
Hi Todd, thanks for working on this. I was testing that in the container and it seems working fine! I don't see any issue regarding your solution now and tests are passing, so I am for the merge.

Just to sum it up. We have two sources for the boot time which is now used in the implementation:
- `/proc/1`
  - Time of the first process that was run after booting 
  - This was the solution so far
  - Not working for machines without RTC - there zero epoch time is returned
- `/proc/uptime`
  - Current time subtracted from the seconds output of this `uptime` file
  - Not working on containers which shares the kernel with the host - there the host kernel uptime is returned

Please could you add comments/docstrings with the explanation of this change into the code? Also please mention the bugzilla ticket inside the commit description + choose better title f.e. `needs_restarting: Fix calculating boot time for systems without RTC` or something like that.

PR #472, opened 2022-12-08
Add a argument to download and import GPG keys for repos

wheelerlaw:
Dependent on rpm-software-management/dnf#1865. 

Add an argument that tells DNF to download and import GPG keys for repos. 

Tests on the way. Looking for feedback on the implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ TAGS
 *~
 build/
 libexec/dnf-utils
+libexec/dnf-utils-3
--- a/plugins/config_manager.py
+++ b/plugins/config_manager.py
@@ -50,6 +50,12 @@ class ConfigManagerCommand(dnf.cli.Command):
             '--add-repo', default=[], action='append', metavar='URL',
             help=_('add (and enable) the repo from the specified file or url'))
         parser.add_argument(
+            '--import-gpg-keys', default=False, action='store_true',
+            help=_('import gpg keys from repo'))
+        parser.add_argument(
+            '--all', default=False, action='store_true',
+            help=_('select all repos (used with --import-gpg-keys)'))
+        parser.add_argument(
             '--dump', default=False, action='store_true',
             help=_('print current configuration values to stdout'))
         parser.add_argument(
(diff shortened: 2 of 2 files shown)


wheelerlaw:
So. now that I am thinking about this, should this functionality be its own plugin? Instead of making changes to `dnf` itself to support these changes, I could just make another plugin and do a little copy-pasta?

xsuchy:
+1 to this. This will ease the `fedora-upgrade` workflow.

kontura:
> So, now that I am thinking about this, should this functionality be its own plugin? Instead of making changes to dnf itself to support these changes, I could just make another plugin and do a little copy-pasta?

Hello, sorry for the delay, I also think it would fit better into a new plugin. Plus it would be nice if we could avoid extending current `dnf` API at this point (we are focusing on new [dnf5](https://github.com/rpm-software-management/dnf5)).

wheelerlaw:
Ah, makes sense. Will the existing Python plugins work with dnf5?

kontura:
Unfortunately no.
We needed to do bigger API changes that are not compatible.

If you plan to work on this more maybe it would be more future proof to try making the plugin for dnf5. There is a [plan to switch to it in fedora 39](https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5) but it is already available in rawhide (also in our [COPR](https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf5-unstable/) for other fedoras). You might run into things that don't work yet but we would welcome feedback and reports on problems.

We already have couple of plugins for dnf5: https://github.com/rpm-software-management/dnf5/tree/main/dnf5-plugins they are in C++ though, it is possible to also create them in python but I don't think we have that documented yet.

I think it would be nice if we had for example a new command for the `repo` subcommand. It could be used like:
```
dnf5 repo import-gpg-keys
```
Or something like that.

kontura:
I am closing this, as previously discussed we should add a new separate plugin either here or even better find a solution for dnf5.

PR #473, opened 2022-12-09
Allow the local plugin to exclude packages from the local cache repo

wheelerlaw:
When using the local plugin as a cache for multiple guest machines, packages that might get installed in the guest that are ignored on the host will end up bypassing the `exclude` filters in the repo configs on the host. 

This change allows the `exclude` option to be configured in the local plugin so that these packages will still be excluded from installation.

### DIFF
--- a/plugins/local.py
+++ b/plugins/local.py
@@ -26,6 +26,7 @@ from dnfpluginscore import _, logger
 
 import dnf
 import dnf.cli
+import dnf.repo
 import os
 import shutil
 import subprocess
@@ -62,6 +63,8 @@ class LocalConfParse(object):
                                              default="/var/lib/dnf/plugins/local")
         else:
             raise KeyError("Disabled")
+
+        main["exclude"] = self.get_value("main", "exclude", "")
 
         if crepo["enabled"]:
             crepo["cachedir"] = self.get_value("createrepo", "cachedir")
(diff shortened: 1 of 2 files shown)


plugins/local.py
```diff
@@ -63,6 +64,8 @@ def parse_config(self):
         else:
             raise KeyError("Disabled")
 
+        main["exclude"] = self.get_value("main", "exclude", "")
```
pkratoch, line 67: In the main configuration, we use the option "excludepks", it is documented as such and the alias "exclude" is there just for compatibility with yum. So, I would propose to either rename this as "excludepkgs", or to support both variants.

pkratoch:
It would also be nice to add a note about this option to the documentation ( https://github.com/rpm-software-management/dnf-plugins-core/blob/master/doc/local.rst ) even though the documentation of this plugin's config options is in a bad shape already.

plugins/local.py
wheelerlaw: Is the `exclude` field a true alias of `excludepkgs`? Or are they two separate fields that need to be read independently?

wheelerlaw:
> It would also be nice to add a note about this option to the documentation ( https://github.com/rpm-software-management/dnf-plugins-core/blob/master/doc/local.rst ) even though the documentation of this plugin's config options is in a bad shape already.

I will do some updating to the documentation.

plugins/local.py
pkratoch: Yes, it is: https://github.com/rpm-software-management/libdnf/blob/dnf-4-master/libdnf/conf/ConfigMain.cpp#L470

PR #474, opened 2022-12-16
system-upgrade: Show warning always for a downstream

jan-kolarik:
As the `distro` package is not available in the BaseOS, the warning implementation will be shown always, but only in related downstreams.

= changelog =
type: bugfix
resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2152846

### DIFF
--- a/dnf-plugins-core.spec
+++ b/dnf-plugins-core.spec
@@ -148,13 +148,14 @@ Summary:    Core Plugins for DNF
 %{?python_provide:%python_provide python3-%{name}}
 BuildRequires:  python3-dbus
 BuildRequires:  python3-devel
-BuildRequires:  python3-distro
 BuildRequires:  python3-dnf >= %{dnf_lowest_compatible}
 BuildRequires:  python3-systemd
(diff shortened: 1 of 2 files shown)


j-mracek:
LGTM, but what about upstream? Shall we remove the warning?
I also thing that we should propagate the change to the RHEL 9 branches.

jan-kolarik:
@j-mracek There is already an existing PR for upstream: https://github.com/rpm-software-management/dnf-plugins-core/pull/475.

j-mracek:
I think that PR is correct and failing container build is not related to the PR.

jan-kolarik:
> I think that PR is correct and failing container build is not related to the PR.

I think so, it should be just because of RHEL8 environment not working in our CI.

j-mracek:
The PR is ready to merge, but I would prefer if the merge will be performed by Release Master. @inknos Is it OK for you?

inknos:
I will merge. please, mirror the PR to the [9.2 branch](https://github.com/rpm-software-management/dnf-plugins-core/tree/rhel-9.2)

PR #475, opened 2022-12-16
system-upgrade: Drop warning in upstream

jan-kolarik:
The `distro` package is not available in the minimal installation of related downstreams where the warning should be displayed. Therefore dropping the implementation from the upstream and affected distributions will handle the situation by themselves.

related link: https://bugzilla.redhat.com/show_bug.cgi?id=2152846


j-mracek:
LGTM

PR #479, opened 2023-01-17
Notify PackageKit when dnf installs packages.

gordonmessmer:


mcatanzaro:
Nice.

One question: what happens if PackageKit is not installed, or if it's broken and unresponsive? Will dnf hang for 30 seconds until the D-Bus call times out? I think a hang is reasonable on systems where PackageKit is installed but broken, but not reasonable on systems that don't have PackageKit at all.

gordonmessmer:
That's worth testing, but maybe a reason to favor https://github.com/PackageKit/PackageKit/pull/600 over including the plugin here.

mcatanzaro:
> That's worth testing, but maybe a reason to favor https://github.com/PackageKit/PackageKit/pull/600 over including the plugin here.

Good point tbh. I think I slightly favor including this plugin with PackageKit rather than dnf-plugins-core, since it's not useful without PackageKit. Since you don't seem to have a strong preference, you might want to close this one to prevent it from winding up in two places by mistake.

PR #483, opened 2023-03-27
Documentation update for reposync (RhBug:2132383)

j-mracek:
The update describe the behavior when `-n` and `--download-metadata` is used.

https://bugzilla.redhat.com/show_bug.cgi?id=2132383

### DIFF
--- a/doc/reposync.rst
+++ b/doc/reposync.rst
@@ -46,7 +46,11 @@ All general DNF options are accepted. Namely, the ``--repoid`` option can be use
     Delete local packages no longer present in repository.
 
 ``--download-metadata``
-    Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c on it.
+    Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c
+    on it. When the option is used with `--newest-only`, only latest packages will be downloaded, but metadata will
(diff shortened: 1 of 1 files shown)


doc/reposync.rst
```diff
@@ -46,7 +46,11 @@ All general DNF options are accepted. Namely, the ``--repoid`` option can be use
     Delete local packages no longer present in repository.
 
 ``--download-metadata``
-    Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c on it.
+    Download all repository metadata. Downloaded copy is instantly usable as a repository, no need to run createrepo_c
+    on it. When the option is used with `--newest-only`, only latest packages will be downloaded, but metadata will
+    still contain older packages. When DNF will try to download older packages it ends with an error, due to missing
+    file. It might by useful to update metadata using `createrepo_c --update` to remove packages with missing RPM file
+    from metadata.
```
pkratoch, line 53: The last two sentences are a bit clunky. How about to swap them: "It might by useful to update the metadata using `createrepo_c --update` to remove the packages with missing RPM files from metadata. Otherwise, DNF ends with an error due to the missing files whenever it tries to download older packages."

jan-kolarik: Also `might by` -> `might be`.

j-mracek: Thank you for great suggestion. I only did not use word `metadata` together with `the`.

PR #487, opened 2023-04-28
Add fix and test assertion for "no systemd unit exists for pid"

troglodyne:
= changelog =
msg: Catch exception in needs-restarting.py when no systemd unit exists for pid
type: bugfix
resolves: 2122587
related: None

### DIFF
--- a/tests/test_needs_restarting.py
+++ b/tests/test_needs_restarting.py
@@ -20,6 +20,8 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+from unittest.mock import patch, Mock
+import dbus
 import needs_restarting
 import tests.support
 
@@ -29,8 +31,6 @@ MM_FILE = '7fc4e1168000-7fc4e1169000 rw-s 1096dd000 00:05 7749' \
           '                      /dev/dri/card0'
 SO_FILE = '30efe06000-30efe07000 r--p 00006000 08:02 139936' \
           '                         /usr/lib64/libSM.so.6.0.1'
-
-
 class NeedsRestartingTest(tests.support.TestCase):
     def test_smap2opened_file(self):
         func = needs_restarting.smap2opened_file
@@ -46,6 +46,17 @@ class NeedsRestartingTest(tests.support.TestCase):
         self.assertTrue(ofile.deleted)
         self.assertEqual(ofile.name, '/usr/lib64/libXfont.so.1.4.1;5408628d')
 
+    def test_get_service_dbus_nounitforpid(self):
+        func = needs_restarting.get_service_dbus
+        # So, This is gonna look kinda screwy unless you are aware of what
+        # this proxies interface is actually doing. The GetUnitByPid function
+        # is normally "dynamically" defined by the get_dbus_method at runtime.
+        # As such there's no actual way to mock it out in any meaningful way
+        # without create=True.
+        with patch( "dbus.proxies.Interface.GetUnitByPID", create=True, side_effect=dbus.DBusException('org.freedesktop.systemd1.NoUnitForPID: PID 1234 does not belong to any loaded unit.') ), \
+             patch( "dbus.SystemBus", return_value=Mock(spec=dbus.Bus) ), \
+             patch( "dbus.bus.BusConnection.__new__", side_effect=dbus.DBusException("Never should hit this exception if mock above works")):
+                 self.assertIsNone(func(1234))
 
 class OpenedFileTest(tests.support.TestCase):
     def test_presumed_name(self):
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -138,10 +138,20 @@ def get_service_dbus(pid):
         systemd_manager_object,
         'org.freedesktop.systemd1.Manager'
     )
-    service_proxy = bus.get_object(
-        'org.freedesktop.systemd1',
-        systemd_manager_interface.GetUnitByPID(pid)
-    )
+    service_proxy = None
+    try:
+        service_proxy = bus.get_object(
+            'org.freedesktop.systemd1',
+            systemd_manager_interface.GetUnitByPID(pid)
+        )
+    except dbus.DBusException as e:
+        # There is no unit for the pid. Usually error is 'NoUnitForPid'.
+        # Considering what we do at the bottom (just return if not service)
+        # Then there's really no reason to exit here on that exception.
+        # Log what's happened then move on.
+        msg = str(e)
+        logger.warning("Failed to get systemd unit for PID {}: {}".format(pid, msg))
+        return
     service_properties = dbus.Interface(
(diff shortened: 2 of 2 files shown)


troglodyne:
Context:
I'm currently investigating many instances regarding this error reported by customers of cPanel (I'm a developer there). After reading up on this error and noting that the name and calling contexts of `get_service_dbus` seem to imply that in the failure case specified in bug 2122587 we should in fact be returning instead of allowing the exception to be unhandled.
Thus a simple try/catch around that type of exception was added, and a small test for that specific case.

toddr:
https://bugzilla.redhat.com/show_bug.cgi?id=2122587

troglodyne:
Error log seems to be here:
https://download.copr.fedorainfracloud.org/results/rpmsoftwaremanagement/CI-dnf-plugins-core-pr487/fedora-37-x86_64/05859398-dnf-plugins-core/builder-live.log.gz
```
1: .....E.................................sss................................
1: ======================================================================
1: ERROR: test_get_service_dbus (tests.test_needs_restarting.NeedsRestartingTest.test_get_service_dbus)
1: ----------------------------------------------------------------------
1: Traceback (most recent call last):
1:   File "/builddir/build/BUILD/dnf-plugins-core-4.4.0-20230428195455.2.g2563e6c/tests/test_needs_restarting.py", line 57, in test_get_service_dbus
1:     self.assertFalse(func(1234))
1:                      ^^^^^^^^^^
1:   File "/builddir/build/BUILD/dnf-plugins-core-4.4.0-20230428195455.2.g2563e6c/plugins/needs_restarting.py", line 132, in get_service_dbus
1:     bus = dbus.SystemBus()
1:           ^^^^^^^^^^^^^^^^
1:   File "/usr/lib64/python3.11/site-packages/dbus/_dbus.py", line 192, in __new__
1:     return Bus.__new__(cls, Bus.TYPE_SYSTEM, mainloop=mainloop,
1:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1:   File "/usr/lib64/python3.11/site-packages/dbus/_dbus.py", line 99, in __new__
1:     bus = BusConnection.__new__(subclass, bus_type, mainloop=mainloop)
1:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1:   File "/usr/lib64/python3.11/site-packages/dbus/bus.py", line 120, in __new__
1:     bus = cls._new_for_bus(address_or_type, mainloop=mainloop)
1:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.FileNotFound: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
1: 
1: ----------------------------------------------------------------------
```
Looks like more mocking is needed, this time of dbus.SystemBus by the looks of things (presumably because the system under test does not in fact have dbus active/working).

troglodyne:
`patch( "dbus.SystemBus", return_value='bogus' )` seems to not cause test failure on my end, now to just figure out how to repro copr environment failure on my end to make sure that "does the trick" on build server as well...

troglodyne:
Shucks. Will have to learn how to repro locally afterall.

jan-kolarik:
Hi @troglodyne and thanks for the contribution! Do you still plan to work on this or are you stuck on the local repro as stated in the last comment? I've tried to pull the PR and when running the `PYTHONPATH=./plugins python3 -m unittest discover -t . -s tests/` from the project root, I'm receiving the same error as mentioned from the copr build...

jan-kolarik:
I was playing with that a bit and it seems that could be done like that:
```python
from unittest.mock import patch, MagicMock
import dbus

...

with patch( "dbus.proxies.Interface.GetUnitByPID", create=True, side_effect=dbus.DBusException('org.freedesktop.systemd1.NoUnitForPID: PID 1234 does not belong to any loaded unit.') ), \
     patch( "dbus.SystemBus", return_value=MagicMock(spec=dbus.Bus) ), \
     patch( "dbus.bus.BusConnection.__new__", side_effect=dbus.DBusException("Never should hit this exception if mock above works")):
         self.assertIsNone(func(1234))
```

This is passing in local container and throwing an exception without your patch in `needs_restarting.py` ...

troglodyne:
Had some other work that got in the way of me picking back up on this, will try the suggested fix to test in an hour or so. Thanks for looking at this!

troglodyne:
still no dbus... looking further

jan-kolarik:
I was looking into your updated code and it differs from what I suggested. Take a look at the `patch` lines, f.e. 
`dbus.BusConnection` vs `dbus.bus.BusConnection` and also I am using all patches inside the `with` clause separated with commas...

troglodyne:
Thanks for the follow up. I was missing what in hindsight should have been obvious given your reply. This time it should be doing what you suggested precisely.

tests/test_needs_restarting.py
```diff
@@ -46,6 +46,22 @@ def test_smap2opened_file(self):
         self.assertTrue(ofile.deleted)
         self.assertEqual(ofile.name, '/usr/lib64/libXfont.so.1.4.1;5408628d')
 
+    def test_get_service_dbus(self):
```
jan-kolarik, line 49: I'd suggest to rename it like `test_get_service_dbus_nounitforpid`.

tests/test_needs_restarting.py
```diff
@@ -46,6 +46,22 @@ def test_smap2opened_file(self):
...
+        # So, This is gonna look kinda screwy unless you are aware of what
+        # this proxies interface is actually doing. The GetUnitByPid function
+        # is normally "dynamically" defined by the get_dbus_method at runtime.
+        # As such there's no actual way to mock it out in any meaningful way
+        # without create=True.
+        with patch( "dbus.proxies.Interface.GetUnitByPID", create=True, side_effect=dbus.DBusException('org.freedesktop.systemd1.NoUnitForPID: PID 1234 does not belong to any loaded unit.') ), \
+             patch( "dbus.SystemBus", return_value=Mock(spec=dbus.Bus) ), \
+             patch( "dbus.bus.BusConnection.__new__", side_effect=dbus.DBusException("Never should hit this exception if mock above works")):
+                 self.assertIsNone(func(1234))
+        # Now, I'd like to have tested additionally the "what if pid gives unit"
```
jan-kolarik, line 60: I think it's fine for now to just add a test which is testing the added functionality meaning not throwing an exception when `NoUnitForPID` happens. So I'd suggest to drop this comment block.

jan-kolarik:
Also please could you modify the commit message so the `resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2122587`?

troglodyne:
Commentary should be addressed now, thanks again for looking at this :)

jan-kolarik:
Great, thanks for that!

PR #489, opened 2023-05-09
Avoid a long obsolete rpm API use

pmatilai:
dsFromHeader() and dsOfHeader() header methods were removed in rpm 4.19, passing header to rpm.ds constructor is the preferred way since rpm 4.8 / RHEL 6.


jan-kolarik:
Thanks for fixing that obsolete usage!

PR #494, opened 2023-06-06
Prospective fix for strange issue with invalid UTF-8 with test

troglodyne:
For now this PR is only for purposes of showing the problem and *one* potential fix to it.

Related bug:
https://bugzilla.redhat.com/show_bug.cgi?id=2212953

I can fixup the commit later for formatting, etc. -- in the meantime I am going to try to gather more data from cPanel customers experiencing this to see just what the heck is in the smaps files for the failing cases.

Regardless, this appeared to me to be a real bug in the upstream code, so I figured I'd approach it in that manner.

### DIFF
--- a/tests/test_needs_restarting.py
+++ b/tests/test_needs_restarting.py
@@ -1,3 +1,4 @@
+# coding: utf-8
 # Copyright (C) 2014 Red Hat, Inc.
 #
 # This copyrighted material is made available to anyone wishing to use,
@@ -15,7 +16,6 @@
 # Red Hat, Inc.
 #
 
-
 from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
@@ -24,6 +24,7 @@ from unittest.mock import patch, Mock
 import dbus
 import needs_restarting
 import tests.support
+import tempfile
 
 DEL_FILE = '3dcf000000-3dcf032000 r-xp 00000000 08:02 140759                ' \
            '         /usr/lib64/libXfont.so.1.4.1;5408628d (deleted)'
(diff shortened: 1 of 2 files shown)


troglodyne:
Test looking like it failed to write the thing I was writing to tmp. Figured something like that might occur. Will clean this up later presuming the general approach even seems sane here to fix the problem.

jan-kolarik:
I found your approach OK as it just allows listing opened files in additional (corner) use cases that you've provided. As we currently focus our efforts into the development of DNF5, I wouldn't develop the current solution further as it seems fixing your problematic cases.

I guess that the CI failure could be related to passing the `uid = 0` in the test case. Could you please try passing there `None`, like: `ofiles = list(needs_restarting.list_opened_files(None))`? Also if it is working, please eventually drop the reviewer / TODO notes from the PR, do the final polishing and it should be good to go. Thanks for that!

tests/test_needs_restarting.py
```diff
@@ -58,6 +58,18 @@ def test_get_service_dbus_nounitforpid(self):
              patch( "dbus.bus.BusConnection.__new__", side_effect=dbus.DBusException("Never should hit this exception if mock above works")):
                  self.assertIsNone(func(1234))
 
+    def test_list_opened_files_garbage_filename(self):
+        # XXX TODO name this file randomly instead of bogus bogus
+        with open("/tmp/bogusbogus", 'wb') as bogusFile:
+            bogusFile.write(b'151e7f7b7000-151e7f7b8000 r--p 00006000 fd:01 14744                      /usr/lib64/lib\xe5Evil-13.37.so')
+        smaps = [[1234,"/tmp/bogusbogus"]]
+        with patch("needs_restarting.list_smaps", return_value=smaps):
+            ofiles = list(needs_restarting.list_opened_files(0));
```
jonathanspw, line 67:
```suggestion
            ofiles = list(needs_restarting.list_opened_files(None))
```

troglodyne:
Reworked commit, thanks for the input. Sorry to have missed the notification earlier.

tests/test_needs_restarting.py
```diff
@@ -58,6 +59,17 @@ def test_get_service_dbus_nounitforpid(self):
              patch( "dbus.bus.BusConnection.__new__", side_effect=dbus.DBusException("Never should hit this exception if mock above works")):
                  self.assertIsNone(func(1234))
 
+    def test_list_opened_files_garbage_filename(self):
+        tempObj = tempfile.NamedTemporaryFile()
+        tempFile = tempObj.name
```
troglodyne, line 64: Probably didn't need to assign it to a var, but eh

troglodyne:
Hmm. The failures in the integration test here don't really seem related to needs-restarting as far as I can tell. A reposync test failed, line 41563 of log, but nothing else that appears to be failing in the set.

troglodyne:
Everything looking good now, it seems :)

jan-kolarik (approved):
LGTM, thank you!

PR #495, opened 2023-06-07
"dnf copr enable" on "Asahi Fedora Linux Remix" guesses epel..x86_64

ericcurtin:
On Asahi Fedora Linux Remix, this can occur when doing a dnf copr enable:

Repository 'epel-37-x86_64' does not exist in project '@asahi/mesa'

There are two bugs here, the code doesn't recongize this as a Fedora distro and defaults to x86_64 when in this case it's actually an aarch64 system.

### DIFF
--- a/plugins/copr.py
+++ b/plugins/copr.py
@@ -462,7 +462,7 @@ Bugzilla. In case of problems, contact the owner of this repository.
             dist = linux_distribution()
         # Get distribution architecture
         distarch = self.base.conf.substitutions['basearch']
-        if any([name in dist for name in ["Fedora", "Fedora Linux"]]):
+        if "Fedora" in dist[0]:
             if "Rawhide" in dist:
                 chroot = ("fedora-rawhide-" + distarch)
             # workaround for enabling repos in Rawhide when VERSION in os-release
@@ -488,7 +488,7 @@ Bugzilla. In case of problems, contact the owner of this repository.
             else:
                 chroot = ("opensuse-leap-{0}-{1}".format(dist[1], distarch))
         else:
-            chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
+            chroot = ("epel-{}-{}".format(dist[1].split(".", 1)[0], distarch if distarch else "x86_64"))
         return chroot
 
     def _download_repo(self, project_name, repo_filename):


xsuchy:
@praiskup can you please review?

plugins/copr.py
```diff
@@ -462,7 +463,7 @@ def _guess_chroot(self):
             dist = linux_distribution()
         # Get distribution architecture
         distarch = self.base.conf.substitutions['basearch']
-        if any([name in dist for name in ["Fedora", "Fedora Linux"]]):
+        if "Fedora" in dist:
```
praiskup, line 466: Thank you for the patch;  what do you have in `dist` in particular?  In my case it is `('Fedora Linux', '38', '')` triplet, so `Fedora` is not there.

ericcurtin: In my case it's the same except the first string is 'Asahi Fedora Linux Remix'

ericcurtin: To be honest this check should work for 'Fedora Linux' also, as it has the 'Fedora' substring

ericcurtin: Maybe we need dist[0] there, what do you think?

```
triplet1 = ('Fedora', '31', '')
triplet2 = ('Asahi Fedora Linux Remix', '37', '')
triplet3 = ('Fedora Linux', '38', '')
triplet4 = ('Debian', '10', '')

if "Fedora" in triplet1[0]:
  print("praiskup1")
if "Fedora" in triplet2[0]:
  print("praiskup2")
if "Fedora" in triplet3[0]:
  print("praiskup3")
if "Fedora" in triplet4[0]:
  print("praiskup4")

```

This prints:

```
praiskup1
praiskup2
praiskup3
```

praiskup: Perhaps that would be better, yes.

praiskup: Btw., while you are taking a look ... testing `dnf5` and `dnf5-plugins` would be worth it in your remix; there's a different detection mechanism and it might be broken as well.

ericcurtin: Tested on Asahi and pushed, hopefully I get time to try out dnf5 in the next month or so

plugins/copr.py
```diff
@@ -488,7 +489,7 @@ def _guess_chroot(self):
             else:
                 chroot = ("opensuse-leap-{0}-{1}".format(dist[1], distarch))
         else:
-            chroot = ("epel-%s-x86_64" % dist[1].split(".", 1)[0])
+            chroot = ("epel-{}-{}".format(dist[1].split(".", 1)[0], platform.machine()))
```
praiskup, line 492: > defaults to x86_64 when in this case it's actually an aarch64 system.

The use of `platform` is not ideal.  Can you please use a different mechanism already used for other distributions, and fallback to x86_64 if unable to detect?

ericcurtin: platform is part of the standard python library are we sure that's not ok?

ericcurtin: platform removed

praiskup (approved):
LGTM, thank you.

jan-kolarik:
> Unrelated build failure:
> 
> ```
> ../dnf/dnf-main.c:27:10: fatal error: libdnf/libdnf.h: No such file or directory
>    27 | #include <libdnf/libdnf.h>
>       |          ^~~~~~~~~~~~~~~~~
> compilation terminated.
> ```

This should be fixed after merging [this](https://github.com/rpm-software-management/dnf5/pull/601) PR. I will retrigger the build again after that and eventually merge this.

jan-kolarik (approved):

plugins/copr.py
xsuchy: You already removed it, but platform is deprecated.
https://docs.python.org/3.5/library/platform.html#platform.linux_distribution
`Deprecated since version 3.5, will be removed in version 3.7.`

ericcurtin: That's good to know, I use this in at least one other python script I use regularly, but what I normally use is platform.machine() like in this line. That looks like it's not being deprecated and removed if I'm reading correctly? I've never used platform.linux_distribution, specifically.

jan-kolarik:
> @jan-kolarik time to kick off a rebuild?

Yep, planning it tomorrow when new dnf5 is ready in nightly.

jan-kolarik:
Hmm, now I realized the dnf5 build here is not taken from nightly, but from regular F38 repo. Therefore it will not work. I will sync with guys what to do with that for now. We should release a new dnf5 in Fedora, but also probably test against dnf5 nightly builds...

jan-kolarik:
The CI fail is unrelated, ready to merge...

PR #496, opened 2023-06-09
system-upgrade: Wait until the upgrade is done before poweroff (RhBug:2211844)

jan-kolarik:
Add a systemd dependency to wait until upgrade service is finished before executing the poweroff when passing the `--poweroff` option in `system-upgrade` plugin.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2211844

### DIFF
--- a/etc/systemd/dnf-system-upgrade.service
+++ b/etc/systemd/dnf-system-upgrade.service
@@ -6,7 +6,7 @@ Documentation=http://www.freedesktop.org/wiki/Software/systemd/SystemUpdates
 DefaultDependencies=no
 Requires=sysinit.target
 After=sysinit.target systemd-journald.socket system-update-pre.target
-Before=shutdown.target system-update.target
+Before=poweroff.target reboot.target shutdown.target system-update.target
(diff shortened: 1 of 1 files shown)


etc/systemd/dnf-system-upgrade.service
```diff
@@ -6,7 +6,7 @@ Documentation=http://www.freedesktop.org/wiki/Software/systemd/SystemUpdates
 DefaultDependencies=no
 Requires=sysinit.target
 After=sysinit.target systemd-journald.socket system-update-pre.target
-Before=shutdown.target system-update.target
+Before=poweroff.target shutdown.target system-update.target
```
jan-kolarik, line 9: I am not sure if not to explicitly add also the `reboot.target` which is triggered by default with the `reboot` subcommand.

jan-kolarik:
Failures in CI test run are unrelated.

etc/systemd/dnf-system-upgrade.service
kontura: I think it would make sense.

PR #497, opened 2023-06-21
Fix zlib reboot requirement (RhBug:2092033)

inknos:


plugins/needs_restarting.py
```diff
@@ -277,15 +279,28 @@ def run(self):
         NEED_REBOOT.extend(opt)
         if self.opts.reboothint:
             need_reboot = set()
+            need_reboot_depends_on_dbus = set()
             installed = self.base.sack.query().installed()
             for pkg in installed.filter(name=NEED_REBOOT):
                 if pkg.installtime > process_start.boot_time:
                     need_reboot.add(pkg.name)
-            if need_reboot:
+            base = dnf.Base()
```
jrohel, line 287: Do we need a new `base` to get the installed packages?
 I see on line 283: `installed = self.base.sack.query().installed()`

j-mracek:
It looks like that zlib is a dependency of dbus in RHEL and not in Fedora 36+. What about to have the patch only for RHEL downstream?

inknos:
> It looks like that zlib is a dependency of dbus in RHEL and not in Fedora 36+. What about to have the patch only for RHEL downstream?

the patch is for the rhel-8.9 branch, it will not be upstream but only tracked in the maint branch

plugins/needs_restarting.py
```diff
@@ -277,15 +279,24 @@ def run(self):
...
+                for pkg in installed.filter(name=NEED_REBOOT_DEPENDS_ON_DBUS):
+                    if pkg.installtime > process_start.boot_time:
+                        need_reboot_depends_on_dbus.add(pkg.name)
+            if need_reboot or need_reboot_depends_on_dbus:
                 print(_('Core libraries or services have been updated '
                         'since boot-up:'))
                 for name in sorted(need_reboot):
                     print('  * %s' % name)
+                for name in sorted(need_reboot_depends_on_dbus):
+                    print('  * %s (depends on dbus that needs reboot)' % name)
```
j-mracek, line 299: zlib is a dependency of dbus, therefore the message is somehow cryptic.

j-mracek: What about:
`'  * %s (dependency of dbus that needs reboot)'`

j-mracek: `'  * %s (dependency of dbus. Recommending to reboot dbus)'`

inknos: @jrohel had the same perplexity.
would you rather add zlib to the list of printed packages?

inknos: > '  * %s (dependency of dbus. Recommending to reboot dbus)'

This is more clear

j-mracek: `'  * %s (dependency of dbus. Recommending reboot of dbus)'`

jrohel: From my point of view. Computer can be **rebooted**. But dbus is a service that can be **restarted**.

If I understand correctly what is happening,  upgrading zlib will cause the dbus service to restart. And it is recomended to reboot computer. Am I right?

inknos: correct

j-mracek:
LGTM

PR #510, opened 2023-11-23
needs-restarting: Add microcode_ctl to a reboot list

ppisar:
To fully update CPU microcode, a reboot is needed because the microcode update should be applied before starting a kernel and other process.

Therefore recommend a reboot after installing or updating microcode_ctl package.

https://issues.redhat.com/browse/RHEL-4600


jan-kolarik (approved):
LGTM

PR #514, opened 2024-01-02
needs-restarting: Add microcode_ctl to a reboot list

ppisar:
commit 9f46488c6b1ac553989a3fd8a9ea18271451c5e8 upstream.

To fully update CPU microcode, a reboot is needed because the microcode update should be applied before starting a kernel and other process.

Therefore recommend a reboot after installing or updating microcode_ctl package.

https://issues.redhat.com/browse/RHEL-4600


ppisar:
The test fails for two independent reasons: This RHEL branch uses Fedora 38 packages. The COPR build root erroneously contains microdnf which is obsoleted by dnf5, but the test attempt to install it by an RPM package.

jan-kolarik (approved):
Thanks.

PR #518, opened 2024-01-18
needs-restarting: fix for RTC not in UTC

inknos:
For systems where the RTC is not in UTC, the boot time is computed in the future if the timezone is +something.
The correct time can be determined by btime in /proc/stat. Such systems are usually configured with `timedatectl set-local-rtc 1` which changes the 3rd line of /etc/adjtime from UTC to LOCAL. See man timedatectl.

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -227,9 +227,24 @@ class ProcessStart(object):
              Works for machines without RTC iff the current time is reasonably correct.
              Does not work on containers which share their kernel with the
              host - there the host kernel uptime is returned
+        - /proc/stat
+             For systems where the RTC is not in UTC, the boot time is computed
+             in the future if the timezone is +something.
+             The correct time can be determined by btime in /proc/stat.
+             Such systems are usually configured with `timedatectl set-local-rtc 1`
+             which changes the 3rd line of /etc/adjtime from UTC to LOCAL.
+             See man timedatectl.
         """
 
         proc_1_boot_time = int(os.stat('/proc/1').st_mtime)
+        with open('/etc/adjtime', 'r') as f:
+            lines = f.readlines()
+        # if LOCAL RTC detected use /proc/stat to determine boot time
+        if lines[2].rstrip() == 'LOCAL':
+            with open('/proc/stat', 'r') as f:
+                btime = [line for line in f if 'btime' in line]
+                proc_stat_boot_time = int(btime[0].strip().split(' ')[1])
+            return max(proc_1_boot_time, proc_stat_boot_time)
         if os.path.isfile('/proc/uptime'):
             with open('/proc/uptime', 'rb') as f:
                 uptime = f.readline().strip().split()[0].strip()


plugins/needs_restarting.py
```diff
@@ -227,9 +227,24 @@ def get_boot_time():
...
 
         proc_1_boot_time = int(os.stat('/proc/1').st_mtime)
+        with open('/etc/adjtime', 'r') as f:
+            lines = f.readlines()
+        # if LOCAL RTC detected use /proc/stat to determine boot time
+        if lines[2].rstrip() == 'LOCAL':
+            with open('/proc/stat', 'r') as f:
+                btime = [line for line in f if 'btime' in line]
+                proc_stat_boot_time = int(btime[0].strip().split(' ')[1])
+            return max(proc_1_boot_time, proc_stat_boot_time)
```
evan-goode, line 247: Won't doing the `max` here not work if (1) `proc_1_boot_time` is X hours ahead of the correct boot time, due to `set-local-rtc 1` and (2) `proc_stat_boot_time` is correct?

inknos: right. actually, returning `proc_stat_boot_time` should be enough since it should work every time local rtc is on... if not in a container, like you mention here (https://github.com/rpm-software-management/dnf5/issues/1184)

evan-goode (changes requested):
Does the `btime` from `/proc/stat` give us any more info than `/proc/uptime`? My current understanding is:

1. When we are NOT in a container:
    - mtime of `/proc/1` is unreliable because an RTC may be unavailable, or it may be incorrect due to `set-local-rtc 1`.
    - current time minus `/proc/uptime` is correct
    - `/proc/stat` `btime` is correct
2. When we ARE in a container:
    - mtime of `/proc/1` is correct
    - `/proc/stat` `btime` is incorrect, has uptime of host system
    - `/proc/uptime` is incorrect, has uptime of the host system

I propose getting `UserspaceTimestampMonotonic`, which is essentially uptime, from systemd (I mentioned [here](https://github.com/rpm-software-management/dnf5/issues/1184#issuecomment-1904229811) also). If systemd is available, this number will definitely be correct. If systemd is not available, we are probably in a container, and we fall back to the old method of `max(proc_1_boot_time, proc_uptime_boot_time)`.

EDIT: I misunderstood `UserspaceTimestampMonotonic`: https://github.com/rpm-software-management/dnf5/issues/1184#issuecomment-1906728174

inknos:
> I propose getting UserspaceTimestampMonotonic, which is essentially uptime, from systemd (I mentioned https://github.com/rpm-software-management/dnf5/issues/1184#issuecomment-1904229811 also). If systemd is available, this number will definitely be correct. If systemd is not available, we are probably in a container, and we fall back to the old method of max(proc_1_boot_time, proc_uptime_boot_time).

this would definitely improve the situation, and it makes sense to see it implemented in dnf4/5.

from my pessimistic point of view, I believe we'll hit more cases in the future that will not fall in this generalization, but I would say yours is a solid approach.

evan-goode:
See https://github.com/rpm-software-management/dnf5/issues/1184#issuecomment-1906728174 and https://github.com/rpm-software-management/dnf5/pull/1198, I think the same fix could be applied here to DNF 4.

inknos:
Closing in favor of this https://github.com/rpm-software-management/dnf-plugins-core/pull/527

PR #527, opened 2024-03-13
needs-restarting: get systemd boot time from UnitsLoadStartTimestamp

inknos:


plugins/needs_restarting.py
```diff
@@ -228,6 +233,25 @@ def get_boot_time():
...
+                dbus.PROPERTIES_IFACE
+            )
+            units_load_start_timestamp = props.Get(
+                'org.freedesktop.systemd1.Manager',
+                'UnitsLoadStartTimestamp'
+            )
+            if units_load_start_timestamp != 0:
+                return int(units_load_start_timestamp / (1000 * 1000))
+        except dbus.exceptions.DBusException as e:
+            logger.warning("D-Bus error getting boot time from systemd: {}".format(e))
```
xsuchy, line 254:
```
        # psutil-2.x.x is not backward compatible to psutil-1.x.x                                                                                                                            
        try: return psutil.boot_time()
        except AttributeError: return psutil.get_boot_time()
```

inknos: thanks!

inknos:
thanks for the review. apparently there is a failing test that need to pass before merging

jan-kolarik (approved):

PR #528, opened 2024-03-27
Document needs-restarting boot time

inknos:
Also, add a warning in the docs if the system is running in RTC

docs for this change: https://github.com/rpm-software-management/dnf-plugins-core/pull/527


evan-goode:
I just made some small edits and added a reference to `man timedatectl`. Let me know if that looks good and I'll merge.

evan-goode (approved):

inknos:
yeah I like it

PR #537, opened 2024-06-06
rhel-9.5: revert systemd boot time

ppisar:
This revers rhel-9.5 commits:

486e9d4d7f03309f3eaf71e297ae956e18357bd8
a0ac8717219a1ec9f466162e87f40afa9aa65284
74de8fe84679e509b16ca1d3f12b7d292210ca27
Resolves: https://issues.redhat.com/browse/RHEL-39775

because it causes a regression.

@evan-goode, could you review this revert?

### DIFF
--- a/doc/needs_restarting.rst
+++ b/doc/needs_restarting.rst
@@ -36,30 +36,30 @@ Description
 Note that in most cases a process should survive update of its binary and libraries it is using without requiring to be restarted for proper operation. There are however specific cases when this does not apply. Separately, processes often need to be restarted to reflect security updates.
 
 .. note::
-   Needs-restarting will try to guess the boot time using three different methods:
+   Needs restarting will try to guess the boot time using three different methods:
 
    ``UnitsLoadStartTimestamp``
-        D-Bus property on ``/org/freedesktop/systemd1``.
+        property on ``/org/freedesktop/systemd1``.
         Works unless the system was not booted with systemd,
         such as in (most) containers.
    ``st_mtime of /proc/1``
-        Reflects the time the first process was run after booting.
+        Reflects the time the first process was run after booting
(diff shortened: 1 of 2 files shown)


evan-goode:
I would prefer to keep in the documentation changes (commits "Needs-restarting: punctuation edits" and "Document needs-restarting boot time") but with the systemd method removed:

```
.. note::
   Needs-restarting will try to guess the boot time using two different methods:

   ``st_mtime of /proc/1``
        Reflects the time the first process was run after booting.
        This works for all known cases except machines without
        a RTC—they awake at the start of the epoch.
   ``/proc/uptime``
        Seconds field of ``/proc/uptime`` subtracted from the current time.
        Works for machines without RTC if the current time is reasonably correct.
        Does not work on containers which share their kernel with the
        host—there, the host kernel uptime is returned.


.. warning::
    Some systems are configured to read the RTC time in the local time
    zone. This mode cannot be fully supported. It will create various problems
    with time zone changes and daylight saving time adjustments. The RTC time
    is never updated, it relies on external facilities to maintain it. **If at
    all possible, use RTC in UTC by calling** ``timedatectl set-local-rtc 0``.
    See ``man timedatectl`` for more information.
```

The warning about RTC-in-local-TZ is good to have in there.

ppisar:
I did a partial revert like that in rhel-10. I can post it in a separate pull request for master branch if you want it in master.

But this full revert is for RHEL-9.5. We cannot just enhance the documentation because that would not fulfill approved acceptance criteria.

evan-goode (approved):

ppisar:
> > I did a partial revert like that in rhel-10. I can post it in a separate pull request for master branch if you want it in master.
> 
> That would be great, thanks.
> 
#538

PR #538, opened 2024-06-06
needs-restarting: Revert using systemd start time

ppisar:
Since commit a0ac8717219a1ec9f466162e87f40afa9aa65284 "dnf needs-restarting -s" failed to spot services started withing few seconds before updating their RPM package.

The cause was using systemd start time instead of a kernel boot time when computing an absolute start time of processes.

This patch removes using the systemd start time until a proper fix is found.

https://issues.redhat.com/browse/RHEL-39775

### DIFF
--- a/doc/needs_restarting.rst
+++ b/doc/needs_restarting.rst
@@ -36,12 +36,8 @@ Description
 Note that in most cases a process should survive update of its binary and libraries it is using without requiring to be restarted for proper operation. There are however specific cases when this does not apply. Separately, processes often need to be restarted to reflect security updates.
 
 .. note::
-   Needs-restarting will try to guess the boot time using three different methods:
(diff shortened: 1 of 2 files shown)


doc/needs_restarting.rst
```diff
@@ -38,10 +38,6 @@ Note that in most cases a process should survive update of its binary and librar
 .. note::
    Needs-restarting will try to guess the boot time using three different methods:
```
evan-goode, line 39: s/three/two

evan-goode (changes requested):

doc/needs_restarting.rst
ppisar: Good catch. I fixed it and force-pushed.

PR #539, opened 2024-06-07
fix(download): resolve dependencies for debuginfo and debugsource packages

### LINKED ISSUE #535: `dnf download` with `--debuginfo` or `--debugsource` doesn't respect `--resolve`
debuginfo packages can have dependencies.

Intended behavior: `dnf download` with `--debuginfo --resolve` downloads the requested debuginfo package with its dependencies. 
Actual current behavior: Dependencies are not downloaded. Tested with package `libstdc++-debuginfo` which depends on `gcc-debuginfo`.

I believe the same to `--debugsource` although I'm unaware of any debugsource packages have dependencies and don't know if it exists.

TheHillBright:
Fixes #535 . Tested with debuginfo package `libstdc++` on Fedora 40.

There're compatible interface changes to `_get_packages_with_deps` that removes the unused parameter `source` and now accepts `set[Package]` objects as input.

We hope to get a new release after this is merged and eventually available in Fedora 40. So, our effort to remove the current internal workaround is minimized.

CC @kontura

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,3 +39,4 @@ DNF-PLUGINS-CORE CONTRIBUTORS
     Wieland Hoffmann <themineo@gmail.com>
     Otto Urpelainen <oturpe@iki.fi>
     Matija Skrgulja <mskrgulja@gmail.com>
+    Xiao Liang <github.com/TheHillBright>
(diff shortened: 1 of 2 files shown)


kontura:
It looks good, thank you.
I will ask about the fedora release in the team.

PR #540, opened 2024-06-12
{pre,post}-transaction-actions: fix package files filtering regression

kszczek:
According to DNF User's FAQ, starting with Fedora 40 filelists metadata became optional. This caused a regression in pre and post transaction actions plugins which rely on this information. Let's request filelists metadata whenever the pre or post transaction actions plugin is enabled.


kszczek:
Force-pushed to make this patch more robust. Instead of requesting filelists metadata whenever the plugin is enabled, it now also checks if at least one action line contains a file pattern before requesting filelists.

kszczek:
To workaround this issue in the meantime, just set `optional_metadata_types=filelists` under the `[main]` section in `/etc/dnf/dnf.conf`.

PR #541, opened 2024-06-14
build: Disable debug plugin on Fedora > 40 and RHEL > 9

ppisar:
This adds -DWITHOUT_DEBUG:str=1 CMake option to disable installing debug plugin (i.e. debug-dump and debug-restore).

The plugin is unhelp and broken.

https://issues.redhat.com/browse/RHEL-23706


evan-goode:
Nice, I fixed a typo but this LGTM.

evan-goode (approved):

PR #543, opened 2024-06-17
needs-restarting: detect packages providing NEED_REBOOT.

jmkeyes:
When setting the reboot hint, search for newly installed or updated packages that provide the package names listed in `NEED_REBOOT`. This allows the `needs-restarting` plugin to detect if any alternative kernels have been installed/updated since the system last booted.


evan-goode (approved):
I see no problem with this, thank you!

PR #547, opened 2024-07-26
expired-pgp-keys: Fix calling the hook at resolved time

jan-kolarik:
Because gpg signatures of packages are checked before actually running the `plugins.pre_transaction` hook, we need to call it earlier.


evan-goode (approved):
LGTM

PR #552, opened 2024-09-23
bootc: First draft of the bootc plugin

dcantrell:
The only subcommand implemented right now is 'status'.  Most options on rpm-ostree are available via the plugin.  The notable exception --advisories since that conflicts with an existing dnf option.

Option groups also don't work, so all of the options are at the plugin level and I am noting the subcommand(s) they apply to in parentheses for the description on each option.

After that we just process the options to build an argument vector and in run() we run it and handle the exit code and log appropriately.


evan-goode:
> Option groups also don't work, so all of the options are at the plugin level and I am noting the subcommand(s) they apply to in parentheses for the description on each option.

Ah, this must be the unfortunately-deprecated argparse feature you mentioned. I think this is fine, the other DNF 4 plugins with subcommands (e.g. system-upgrade) do this.

plugins/bootc.py
```diff
@@ -0,0 +1,125 @@
...
+                self.extargs.append("--sysroot=%s" % self.opts.installroot)
+
+    def run(self):
+        proc = subprocess.Popen(
+            self.extargs, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        output = proc.communicate()
+
+        if proc.returncode != 0:
+            logger.critical(output)
```
evan-goode, line 122: I made a commit [here](https://github.com/evan-goode/dnf-plugins-core/tree/evan-goode/bootc) to fix the output, currently it is printing the `__str__()` of the stdout, stderr tuple which looks like `(b"stderr\n", b"stdout\n") and is not pretty.

In general, we will want better logic here to display the output from rpm-ostree that preserves the order of lines logged to stdout and stderr (versus logging all stdout, then all stderr). But this is alright for now; the `status` subcommand is pretty simple.  Ideally, colors from rpm-ostree's output would be displayed too, but that might be awkward since colored output shouldn't go in logs. And it doesn't seem `rpm-ostree` has a `--color` flag to force colors on and off, it's probably checking whether its stdout is a terminal.

evan-goode (changes requested):
Nice, glad to see this in the dnf-plugins-core repo. Perhaps we should merge this to a `bootc` branch though to keep the master branch somewhat releasable? Or maybe we could clearly mark the command as experimental.

I added a commit here to get the RPM to build: https://github.com/evan-goode/dnf-plugins-core/tree/evan-goode/bootc.

For testing this, I found it useful to use `rpm-ostree usroverlay` to get a temporary writable `/usr` and then installing with `make install`. Working out of a virtualenv could work too, I haven't tried that yet.

dcantrell:
> > Option groups also don't work, so all of the options are at the plugin level and I am noting the subcommand(s) they apply to in parentheses for the description on each option.
> 
> Ah, this must be the unfortunately-deprecated argparse feature you mentioned. I think this is fine, the other DNF 4 plugins with subcommands (e.g. system-upgrade) do this.

Yeah, this was sort of frustrating to me because option groups would have made the --help output a little nicer.  However this is just a limitation of Python argparse so we can run with it.

plugins/bootc.py
dcantrell: > I made a commit [here](https://github.com/evan-goode/dnf-plugins-core/tree/evan-goode/bootc) to fix the output, currently it is printing the `__str__()` of the stdout, stderr tuple which looks like `(b"stderr\n", b"stdout\n")` and is not pretty.

Oh yeah, gross.  I was thinking sending stderr to STDOUT would keep the output interleaved as it would display on the terminal, which would be suitable for logging.

> In general, we will want better logic here to display the output from rpm-ostree that preserves the order of lines logged to stdout and stderr (versus logging all stdout, then all stderr). But this is alright for now; the `status` subcommand is pretty simple. Ideally, colors from rpm-ostree's output would be displayed too, but that might be awkward since colored output shouldn't go in logs. And it doesn't seem `rpm-ostree` has a `--color` flag to force colors on and off, it's probably checking whether its stdout is a terminal.

Agreed, this is a part of the plugin I envision evolving as more commands are added.  Color may be a reach, but probably not impossible.

dcantrell:
> Nice, glad to see this in the dnf-plugins-core repo. Perhaps we should merge this to a `bootc` branch though to keep the master branch somewhat releasable? Or maybe we could clearly mark the command as experimental.
> 
> I added a commit here to get the RPM to build: https://github.com/evan-goode/dnf-plugins-core/tree/evan-goode/bootc.
> 
> For testing this, I found it useful to use `rpm-ostree usroverlay` to get a temporary writable `/usr` and then installing with `make install`. Working out of a virtualenv could work too, I haven't tried that yet.

Yeah, let me redo this PR so it goes in to the bootc branch instead and add your fixes.  The usroverlay thing is what I am using as well, but I'm not sure how good that will be for all testing.  I guess we can just expand this as we go.

dcantrell:
Closing this to resubmit a new PR against the correct branch.

PR #555, opened 2024-09-24
bootc: First draft of the bootc plugin

dcantrell:
The only subcommand implemented right now is 'status'.  Most options on rpm-ostree are available via the plugin.  The notable exception --advisories since that conflicts with an existing dnf option.

Option groups also don't work, so all of the options are at the plugin level and I am noting the subcommand(s) they apply to in parentheses for the description on each option.

After that we just process the options to build an argument vector and in run() we run it and handle the exit code and log appropriately.


dcantrell:
So with subprocess.Popen(), you can set text=True to get text output tied to the default system encoding for stdout and stderr.  Set stderr to subprocess.STDOUT and set stdout to subprocess.PIPE and when you call communicate(), you do still get a tuple but stderr will be None because it's combined with stdout.  The ordering should be as it would be if you ran the command directly.

evan-goode (approved):

evan-goode:
> So with subprocess.Popen(), you can set text=True to get text output tied to the default system encoding for stdout and stderr. Set stderr to subprocess.STDOUT and set stdout to subprocess.PIPE and when you call communicate(), you do still get a tuple but stderr will be None because it's combined with stdout. The ordering should be as it would be if you ran the command directly.

OK, still not ideal but good enough for now.

PR #556, opened 2024-09-24
needs-restarting: Add --exclude-services

evan-goode:
For consumers of `dnf4 needs-restarting`'s output, it's useful to separate the systemd services needing restart vs. the processes not managed by systemd that need restarting. This new `dnf4 needs-restarting --exclude-services` ONLY prints information about processes NOT managed by systemd, and for listing systemd services that need to be restarted, the existing `--services` flag can be used.

When both `--services` and `--exclude-services` are passed `--exclude-services` is ignored.

For https://issues.redhat.com/browse/RHEL-56137.

CI PR on the way.

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -264,6 +264,8 @@ class NeedsRestartingCommand(dnf.cli.Command):
                                    "(exit code 1) or not (exit code 0)"))
         parser.add_argument('-s', '--services', action='store_true',
                             help=_("only report affected systemd services"))
+        parser.add_argument('--exclude-services', action='store_true',
+                            help=_("don't list stale processes that correspond to a systemd service"))
 
     def configure(self):
         demands = self.cli.demands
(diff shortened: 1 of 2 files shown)


evan-goode:
@martinpitt would you mind reviewing this to see whether it would work for you?

martinpitt (approved):
Thanks @evan-goode ! This looks correct and provides what we need for dnf4.

evan-goode:
Hmm, I guess our ci-dnf-stack test suite is not really prepared to test this functionality (it doesn't test `dnf4 needs-restarting --services` either) since it runs in a systemd-less container. We will want to update our Beaker tests for this. Marking as RFR.

plugins/needs_restarting.py
```diff
@@ -303,19 +305,31 @@ def run(self):
...
 
         if self.opts.services:
-            names = set([get_service_dbus(pid) for pid in sorted(stale_pids)])
-            for name in names:
-                if name is not None:
-                    print(name)
-            return 0
+            for stale_service_name in sorted(stale_service_names):
+                print(stale_service_name)
+
```
kontura, line 333: Given that the `return 0` was removed here will it not print both the service name and `pid` when the `--services` option is used?

plugins/needs_restarting.py
```diff
@@ -303,19 +305,32 @@ def run(self):
...
+            if pkg.installtime <= process_start(pid):
+                continue
+            if self.opts.services or self.opts.exclude_services:
+                service_name = get_service_dbus(pid)
+                if service_name is None:
+                    stale_pids.add(pid)
+                else:
+                    stale_service_names.add(service_name)
+                    if not self.opts.exclude_services:
+                        stale_pids.add(pid)
```
kontura, line 324: I didn't realize this before but now that the `--services` option prints only `stale_service_names` there is no point in adding to the `stale_pids` when `--services` is set right?

glum23:
The PR has been tested downstream in RHEL, everything passed.

glum23:
The PR has also been tested in Fedora40, everything passed. I'd say it's possible to merge the PR.

plugins/needs_restarting.py
evan-goode: I suppose, but I think it's fine to leave it.

PR #558, opened 2024-10-01
bootc: Simplify I/O with rpm-ostree subprocess

evan-goode:
See commit messages.

Also merges changes from master branch.

For https://issues.redhat.com/browse/SWM-3432

### DIFF
--- a/plugins/bootc.py
+++ b/plugins/bootc.py
@@ -211,5 +211,4 @@ class BootcCommand(dnf.cli.Command):
     def run(self):
         proc = subprocess.Popen(self.extargs, stdout=sys.stdout, stderr=sys.stderr)
         proc.wait()
-        if proc.returncode != 0:
-            raise dnf.cli.CliError
+        sys.exit(proc.returncode)
--- a/plugins/bootc.py
+++ b/plugins/bootc.py
@@ -9,6 +9,7 @@ from __future__ import unicode_literals
 import dnf
 import logging
 import subprocess
+import sys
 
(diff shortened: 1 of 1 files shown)


plugins/bootc.py
```diff
@@ -112,16 +113,5 @@ def configure(self):
...
-
-        # stdout and stderr will be combined in stdout, err will be None here
-        (out, err) = proc.communicate()
-
-        if proc.returncode != 0:
-            logger.critical(out)
-            raise dnf.cli.CliError
-        else:
-            logger.info(out)
+        proc = subprocess.Popen(self.extargs, stdout=sys.stdout, stderr=sys.stderr)
```
dcantrell, line 116: The main reason I did the combining of stdout and stderr and use logger.critical(out) is because I thought we should capture rpm-ostree output and put it in the dnf log.  However, I guess it doesn't really matter since it's ultimately running rpm-ostree and I imagine it has its own logging mechanism.  I am fine with this change, but we should keep in mind that someone may ask down the road where the rpm-ostree stuff is in the dnf logs and if that comes up, we're back in this conversation.

dcantrell (approved):

plugins/bootc.py
evan-goode: Yeah, eventually we should make sure to log to `/var/log/dnf.log`. It's just tricky to do now when all we have is unstructured output from rpm-ostree.

evan-goode:
Rebased and added a missing `proc.wait()`.

PR #560, opened 2024-10-07
needs-restarting: Get boot time from systemd UnitsLoadStartTimestamp

evan-goode:
Resolves https://issues.redhat.com/browse/RHEL-35577.

Get the boot time from `UnitsLoadStartTimestamp` if systemd is available, but make sure to use the kernel boot time for calculating process start times using data from procfs. The previous attempt [1] at this failed to do so and introduced a regression [2].

Also, get the kernel boot time from the btime field of /proc/stat instead of calculating it from /proc/uptime, to be consistent with what procps-ng does.

[1] https://github.com/rpm-software-management/dnf-plugins-core/pull/527
[2] https://issues.redhat.com/browse/RHEL-39775

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -206,36 +206,89 @@ class OpenedFile(object):
                 return match.group(1)
         return self.name
 
-
 class ProcessStart(object):
     def __init__(self):
-        self.boot_time = self.get_boot_time()
-        self.sc_clk_tck = self.get_sc_clk_tck()
+        self.kernel_boot_time = ProcessStart.get_kernel_boot_time()
+        self.boot_time = ProcessStart.get_boot_time(self.kernel_boot_time)
+        self.sc_clk_tck = ProcessStart.get_sc_clk_tck()
 
     @staticmethod
-    def get_boot_time():
+    def get_kernel_boot_time():
+        try:
+            with open('/proc/stat', 'r') as file:
+                for line in file:
+                    if line.startswith("btime "):
+                        key, value = line.split()
+                        return float(value)
+        except OSError as e:
+            logger.debug("Couldn't read /proc/stat: %s", e)
+        return 0
+
+    @staticmethod
+    def get_boot_time(kernel_boot_time):
         """
-        We have two sources from which to derive the boot time. These values vary
+        We have three sources from which to derive the boot time. These values vary
         depending on containerization, existence of a Real Time Clock, etc.
-        For our purposes we want the latest derived value.
+        - UnitsLoadStartTimestamp property on /org/freedesktop/systemd1
+             The start time of the service manager, according to systemd itself.
+             Seems to be more reliable than UserspaceTimestamp when the RTC is
+             in local time. Works unless the system was not booted with systemd,
+             such as in (most) containers.
         - st_mtime of /proc/1
-             Reflects the time the first process was run after booting
-             This works for all known cases except machines without
-             a RTC - they awake at the start of the epoch.
-        - /proc/uptime
-             Seconds field of /proc/uptime subtracted from the current time
-             Works for machines without RTC iff the current time is reasonably correct.
-             Does not work on containers which share their kernel with the
-             host - there the host kernel uptime is returned
+             Reflects the time the first process was run after booting. This
+             works for all known cases except machines without a RTC---they
+             awake at the start of the epoch.
+        - btime field of /proc/stat
+             Reflects the time when the kernel started. Works for machines
+             without RTC iff the current time is reasonably correct. Does not
+             work on containers which share their kernel with the host---there,
+             the host kernel uptime is returned.
         """
 
-        proc_1_boot_time = int(os.stat('/proc/1').st_mtime)
-        if os.path.isfile('/proc/uptime'):
-            with open('/proc/uptime', 'rb') as f:
-                uptime = f.readline().strip().split()[0].strip()
-                proc_uptime_boot_time = int(time.time() - float(uptime))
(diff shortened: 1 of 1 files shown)


evan-goode:
@bajertom said:

> https://issues.redhat.com/browse/RHEL-14900 fails with current dnf-plugins-core.
> https://issues.redhat.com/browse/RHEL-39775 passes with the current dnf-plugins-core.
> I have rebuild dnf-plugins-core with the patch from PR#560 and installed it on a system and then both test cases PASS, therefore no regression is introduced.
> I have also tried it on RHEL-10 and the result is the same. (Actually the latter fails on an --exclude-services option being unknown to needs-restarting command, however I see that it is going to get implemented by you :slightly_smiling_face: )

Marking this as RFR. @ppisar I'll request your review since you investigated the regression last time :)

plugins/needs_restarting.py
```diff
@@ -208,46 +208,91 @@ def presumed_name(self):
...
+                systemd1,
+                dbus.PROPERTIES_IFACE
+            )
+            units_load_start_timestamp = props.Get(
+                'org.freedesktop.systemd1.Manager',
+                'UnitsLoadStartTimestamp'
+            )
+            if units_load_start_timestamp != 0:
+                systemd_boot_time = units_load_start_timestamp / (1000 * 1000)
+                logger.debug(f"Got boot time from systemd: {systemd_boot_time}")
```
ppisar, line 264: F-strings are available since Python 3.6. If you really want to use them, consider that other code of dnf-plugins-core does not use them, please add the dependency to the spec file.

ppisar (commented):
Otherwise, the code looks good. I will try to do some tests.

plugins/needs_restarting.py
```diff
@@ -208,46 +208,91 @@ def presumed_name(self):
...
-    def get_sc_clk_tck():
+        units_load_start_timestamp = None
+        try:
+            # systemd timestamps are the preferred method to determine boot
+            # time. max(proc_1_boot_time, kernel_boot_time) does not allow us
+            # to disambiguate between an unreliable RTC (e.g. the RTC is in
+            # UTC+1 instead of UTC) and a container with a proc_1_boot_time >
+            # kernel_boot_time. So we use UnitsLoadStartTimestamp if it's
+            # available, else fall back to the other methods.
+            bus = dbus.SystemBus()
```
ppisar, line 249: If dbus daemon is not running (e.g. in a UBI container), "dnf needs-restarting -s" dies with an unhandled exception:

~~~~
  File "/usr/bin/dnf", line 62, in <module>
  File "/usr/lib/python3.9/site-packages/dnf/cli/main.py", line 201, in user_main
    errcode = main(args)
  File "/usr/lib/python3.9/site-packages/dnf/cli/main.py", line 67, in main
    return _main(base, args, cli_class, option_parser_class)
  File "/usr/lib/python3.9/site-packages/dnf/cli/main.py", line 106, in _main
    return cli_run(cli, base)
  File "/usr/lib/python3.9/site-packages/dnf/cli/main.py", line 122, in cli_run
    cli.run()
  File "/usr/lib/python3.9/site-packages/dnf/cli/cli.py", line 1057, in run
    return self.command.run()
  File "/usr/lib/python3.9/site-packages/dnf-plugins/needs_restarting.py", line 356, in run
    names = set([get_service_dbus(pid) for pid in sorted(stale_pids)])
  File "/usr/lib/python3.9/site-packages/dnf-plugins/needs_restarting.py", line 356, in <listcomp>
    names = set([get_service_dbus(pid) for pid in sorted(stale_pids)])
  File "/usr/lib/python3.9/site-packages/dnf-plugins/needs_restarting.py", line 133, in get_service_dbus
    bus = dbus.SystemBus()
  File "/usr/lib64/python3.9/site-packages/dbus/_dbus.py", line 195, in __new__
    return Bus.__new__(cls, Bus.TYPE_SYSTEM, mainloop=mainloop,
  File "/usr/lib64/python3.9/site-packages/dbus/_dbus.py", line 102, in __new__
    bus = BusConnection.__new__(subclass, bus_type, mainloop=mainloop)
  File "/usr/lib64/python3.9/site-packages/dbus/bus.py", line 124, in __new__
    bus = cls._new_for_bus(address_or_type, mainloop=mainloop)
dbus.exceptions.DBusException: org.freedesktop.DBus.Error.FileNotFound: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
~~~~

ppisar: My bad. An unrelated bug in get_service_dbus().

ppisar:
Manual testing did not find any problem.

plugins/needs_restarting.py
```diff
@@ -208,46 +208,91 @@ def presumed_name(self):
...
 class ProcessStart(object):
-    def __init__(self):
-        self.boot_time = self.get_boot_time()
-        self.sc_clk_tck = self.get_sc_clk_tck()
-
-    @staticmethod
-    def get_boot_time():
+    @functools.cached_property
+    def kernel_boot_time(self):
+        with open('/proc/stat', 'r') as file:
```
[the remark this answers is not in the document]
evan-goode: Sure, I've changed it to log the error and return 0.

plugins/needs_restarting.py
evan-goode: Ah, didn't realize we were still supporting Python so old. `functools.cached_property` is from 3.8 also. I've taken that and the f-strings out.

ppisar (approved):
The Ansible test failure (mismatched libstdc++.so.6(CXXABI_1.3.15)) is unrelated.

PR #562, opened 2024-10-17
bootc: Implement the 'remove' subcommand

dcantrell:
Implement the 'remove' subcommand for the bootc plugin and add documentation.  Most of the options for this subcommand will come from global dnf command line option settings.  The only option not handled is --os=OSNAME from rpm-ostree.

https://issues.redhat.com/browse/SWM-3098

### DIFF
--- a/doc/bootc.rst
+++ b/doc/bootc.rst
@@ -20,6 +20,9 @@ Arguments (bootc)
 ``bootc install [options] <pkg-spec>...``
     Install one or more packages.
 
+``bootc remove [options] <pkg-spec>...``
+    Remove one or more packages.
+
 -------
 Options
 -------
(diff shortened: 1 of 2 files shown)


evan-goode:
Oh, I thought we were not going to implement install and remove via rpm-ostree since these will break `bootc upgrade`: https://docs.fedoraproject.org/en-US/bootc/rpm-ostree/#_rpm_ostree_installoverride_replace. But I suppose we will decide that on Monday's meeting with the bootc people.

IMO the plan for now should be to have the `dnf4 bootc` install and remove commands support `--transient` only and call `rpm-ostree usroverlay` and then set up an ordinary DNF transaction.

PR #566, opened 2024-11-04
reposync: Avoid multiple downloads of duplicate packages

m-blaha:
Download each package only once if it would have been saved to the same location. This can occur if the repository metadata contains duplicate entries for the same package.

Resolves: https://issues.redhat.com/browse/RHEL-64320


kontura:
I believe the Ansible tests fail due to: https://github.com/rpm-software-management/ci-dnf-stack/pull/1591.

kontura (approved):
This looks good, I would just like to wait for the Ansible tests fix.

Issue #520, opened 2024-02-12
Failed loading plugin "debuginfo-install": module 'dnf' has no attribute 'cli'


L-U-T-i:
dnfdaemon is throwing me out the following error:
`Failed loading plugin "debuginfo-install": module 'dnf' has no attribute 'cli'`

I've looked at the code of [debuginfo-install.py](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/debuginfo-install.py) and found only:
`import dnf`
line.

Considering 'cli' module resides in its own subdirectory, shouldn't there be also a line:
`import dnf.cli`
there? Or, any 'cli' being referenced as 'dnf.cli' (and not just 'cli')?

I've manually added this import line, and it seems to resolve the issue (that error is not logged anymore at least).

L-U-T-i:
Resolved, thank you.

L-U-T-i:
Resolved, thank you.

jan-kolarik:
Hi, thanks for the report. It seems that when the `debuginfo-install` plugin is used by dnf itself, the `dnf.cli` module is imported by some other script beforehand, so it's working. However, as you've pointed out, it should be referenced directly in `debuginfo-install.py`.

PR #578, opened 2025-01-28
Fix missing dnf.cli imports

moreati:
fixes https://github.com/rpm-software-management/dnf-plugins-core/issues/520


moreati:
```console
➜  dnf-plugins-core git:(issue520) ✗ diff -u <(ag -l --python 'import dnf\.cli' | sort) <(ag -l --python 'dnf\.cli' | sort) | wc -l
       0
```

m-blaha:
Cool, thanks again.

PR #581, opened 2025-03-14
README: direct to Libera.Chat instead of Freenode

evan-goode:
Companion to https://github.com/rpm-software-management/dnf/pull/2221. @ppisar


ppisar (approved):

dcantrell:
This is fine, but I'm not sure we're even still on Libera.chat anymore.  I have been but it's dead for all intents and purposes.

PR #583, opened 2025-03-27
multisig: A new plugin for verifying extraordinary RPM signatures

ppisar:
This plugin executes a dedicated rpmkeys(8) tool for verifying RPM packages bound into a transaction.

The dedicated tool is supposed to understand signatures in RPM version 6 format. One feature of that format is that you can have multiple signatures on a single RPM package.

This plugin takes public keys from a key store separate from the native RPM database because the native database might reject storing OpenPGP keys with an unsupported key schema.

FIXME: Report which key is missing.

FIXME: Path for the dedicated rpmkeys is hard-coded to nonproduction path.

FIXME: Dependencies in spec file, conditions for RHEL?

FIXME: This plugin does not support importing missing public keys.


ppisar:
This pull request was retarget to rhel-9.7 branch (#587).

PR #525, opened 2024-03-01
Detect CentOS and RHEL properly instead of defaulting to epel

jakub-vavra-cz:
= changelog =
msg: Fix detection of CentOS Stream and Red Hat Enterprise Linux
type: bugfix
resolves: https://github.com/rpm-software-management/dnf-plugins-core/issues/524

PR #586, opened 2025-06-21
Improve chroot guessing on EL

### LINKED ISSUE #524: copr plugin: _guess_chroot does not handle centos stream properly
I am trying to enable repo on centos stream 9  "dnf -y copr enable --hub copr.fedorainfracloud.org  @sssd/pr7193"
It is incorrectly defaulting to epel-9-x86_64 instead of centos-stream-9-x86_64:

2024-03-01T11:35:55     Repository 'epel-9-x86_64' does not exist in project '@sssd/pr7193'.
2024-03-01T11:35:55     Available repositories: 'fedora-38-x86_64', 'fedora-40-x86_64', 'centos-stream-9-x86_64', 'fedora-rawhide-x86_64', 'fedora-39-x86_64'

I do not want to put there repo explicitly as I want to reuse same code on centos, fedora and rhel.

ekohl:
When a project is not built in the `epel` buildroot but instead in `centos-stream` or `rhel` the command `copr enable $copr` will fail.

Copr can provide available chroots when a repository is not found and this information is already shown to the user. we can improve our guessing with this information when a user hasn't provided the chroot.

The result is that it'll always prefer `epel`, but can also use the `centos-stream` or the `rhel` buildroot when available. This makes it easier for projects that build without EPEL.

The PR is organized into some refactoring commits that shouldn't affect anything. The result is that the final commit actually making the change is relatively small. I've chosen to first attempt the primary guess instead of probing for available repos first. This should result in lower server load for copr because in the majority of cases the guess is still valid.

This is an alternative take on https://github.com/rpm-software-management/dnf-plugins-core/pull/525.

Fixes #524


plugins/copr.py
```diff
@@ -491,18 +497,27 @@ def _guess_chroot(self):
...
             chroot = "amazonlinux-{}-{}".format(dist[1], distarch if distarch else "x86_64")
         else:
-            chroot = ("epel-{}-{}".format(dist[1].split(".", 1)[0], distarch if distarch else "x86_64"))
+            releasever = dist[1].split(".", 1)[0]
+            arch = distarch if distarch else "x86_64"
+
+            if available_chroots:
+                guesses = ['epel', 'rhel']
+                if "CentOS Stream" in dist:
+                    guesses.insert(1, 'centos-stream')
```
ekohl, line 506: https://github.com/rpm-software-management/dnf-plugins-core/pull/525#issuecomment-1985507234 showed me there's also the `centos-stream+epel-next` chroot. Should we support that too?
```suggestion
                if "CentOS Stream" in dist:
                    guesses = ['centos-stream+epel-next', 'epel', 'centos-stream', 'rhel']
                else:
                    guesses = ['epel', 'rhel']
```

carlwgeorge: The `centos-stream+epel-next` chroots are tricky.  For versions 8 and 9 (the only ones we have EPEL Next), real systems should be one of:

* RHEL + EPEL
* CentOS Stream + EPEL + EPEL Next

But in mock-core-configs we have all of these:

* `rhel+epel`
* `centos-stream+epel`
* `centos-stream+epel-next`

For versions 8 and 9, EPEL on CentOS Stream means both EPEL and EPEL Next.  Using only EPEL without EPEL Next on CentOS Stream is not valid and will result in uninstallable packages (albeit rarely).  I think the `centos-stream+epel-next` chroots should be retired and `centos-stream+epel` chroots should be updated to also include the `epel-next` template (matching what `centos-stream+epel-next` is now).  If we do that, there won't be any need to include logic in the copr plugin for `centos-stream+epel-next`.

Also, EPEL 8 Next is already EOL, EPEL 9 Next goes EOL in about two years, and EPEL 10 uses a totally different structure that obsoletes EPEL Next.

carlwgeorge:
Thanks for working on this!  I'd definitely like to see something like this implemented.  The current logic makes sense for EPEL 8/9, but I think it's not quite right for EPEL 10.  The `epel-10` chroot is actually `centos-stream+epel-10`.  This chroot will often (but not always) work fine on RHEL 10, but `rhel+epel-10` is the better first choice.  Because of this I don't think we should make the initial assumption of the `epel` chroot being the best one.  If we ignore the "make an initial assumption to usually avoid fetching the available chroots" idea for a moment, this is how I envision the priority order for each distro.

* RHEL 9 (same logic for RHEL 8)
    * `epel-9` (`rhel+epel-9`)
    * `rhel-9`
    * `centos-stream+epel-next-9`
    * `centos-stream-9`
* CentOS Stream 9 (same logic for CentOS Stream 8, although it's already EOL)
    * `centos-stream+epel-next-9`
    * `centos-stream-9`
    * `epel-9` (`rhel+epel-9`)
    * `rhel-9`
* RHEL 10
    * `rhel+epel-10`
    * `rhel-10`
    * `epel-10` (`centos-stream+epel-10`)
    * `centos-stream-10`
* CentOS Stream 10
    * `epel-10` (`centos-stream+epel-10`)
    * `centos-stream-10`
    * `rhel+epel-10`
    * `rhel-10`

I don't know if it's worth the effort to assume the first option for each distro and then check for alternatives, or if we should just fetch the available chroots first and then pick the best one.

PR #596, opened 2025-09-12
multisig: Fix importing OpenPGPv6 keys

ppisar:
If a key in OpenPGP version 6 format format was defined for a repository, the key was always omitted and never imported to pqrpm key store. As a result, packages signed with that key could not be verified:

    [...]
    Importing GPG key 0xC9D3B13C:
     Userid     : "rsa4k_user"
     Fingerprint: CE77 8FBE 1E8D 4DD2 7691 FB33 CA58 1189 C9D3 B13C
     From       : /root/repo/rsa4k.cert
    Key imported successfully
    foo                                                             16 MB/s |  16 kB     00:00
    Import of key(s) didn't help, wrong key(s)?
    Public key for foo-1.0-1.noarch.rpm is not installed. Failing package is: foo-1.0-1.noarch
     GPG Keys are configured as: file:///root/repo/rsa4k.cert, file:///root/repo/mldsa87.cert
    Error: GPG check FAILED

The cause was that multisig plugin called dnf.crypto.retrieve() to obtain a list of OpenPGP objects from the gpgkey URLs. That DNF function uses dnf.crypto.rawkey2infos() to parse the OpenPGP packets with a gpgme library. But library does not support OpenPGPv6. As a result, multisig got an empty list of OpenPGP objects and dis not import any key in OpenPGPv6 format into pqrm key store.

Considering that the only available OpenPGPv6 parser is rpm-sequoia library wrapped by pqrpm's librpmio library, and this plugin cannot load them into its name space not to clash with the system librpmio library, an ideal fix would have to develop standalone executable on top of them and parse it output by this plugin. And considering that best code is no code,

I removed printing any details about the imported keys from this plugin. The only details about the key this plugin now prints is URL or local path to the key file:

    Importing GPG keys from: /root/repos/opgp6/mldsa65.cert
    Is this ok [y/N]: y
    Key imported successfully

Because the user ID and key ID were also required for DNSSEC validation, I had to remove that code. This patch also removes all code that became unreachable because of this fix.

Another downside is that now the plugin cannot does not check whether a key has already been imported, resulting always asking a using for importing the keys associated with a repository the unverifiable package comes from.

Resolve: https://issues.redhat.com/browse/RHEL-114424

### DIFF
--- a/plugins/multisig.py
+++ b/plugins/multisig.py
@@ -1,15 +1,25 @@
 from __future__ import print_function, absolute_import, unicode_literals
 import dnf
-import dnf.crypto
-import dnf.dnssec
 import dnf.exceptions
 from dnf.i18n import ucd
 import dnf.rpm.transaction
 import dnf.transaction
+import dnf.util
 from dnfpluginscore import _, logger
 import os
 import subprocess
 import sys
+
+class MultiSigKey(object):
+    def __init__(self, data, url):
+        self.id_ = None
+        self.fingerprint = None
+        self.timestamp = None
+        self.raw_key = data
+        self.url = url
+        self.userid = None
+        self.short_id = None
+        self.rpm_id = None
 
 class MultiSig(dnf.Plugin):
     """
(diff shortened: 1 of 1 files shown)


ppisar:
> The approach is fine with me and the code looks good. I wasn't able to test it since I don't know how to set up a "post-quantum repository".

You can use keys and packages from a URL in the linked Jira ticket.

evan-goode (approved):
Tested, this does make the key import succeed in the reproducer documented in RHEL-114424. Thanks!

PR #614, opened 2026-04-28
manifest: Split downloaded packages by basearch with --per-arch option

pkratoch:
Related: https://redhat.atlassian.net/browse/SWM-5555

### DIFF
--- a/plugins/manifest.py
+++ b/plugins/manifest.py
@@ -54,7 +54,7 @@ class ManifestCommand(dnf.cli.Command):
         parser.add_argument('--use-system', action='store_true', help=_('use installed packages for resolving dependencies'))
         parser.add_argument("--source", action='store_true', help=_('include also source packages'))
         parser.add_argument("--archs", nargs='+', help=_('explicitly specify basearchs to use'))
-        parser.add_argument("--per-arch", action='store_true', help=_('separate packages by basearch into individual manifest files'))
+        parser.add_argument("--per-arch", action='store_true', help=_('separate packages by basearch into individual manifest files or download directories'))
 
     def configure(self):
         self.cmd = self.opts.subcommand[0]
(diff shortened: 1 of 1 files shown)


plugins/manifest.py
```diff
@@ -157,8 +157,23 @@ def _download(self):
         for arch in self.archs:
             specs = self._manifest_to_pkg_specs(arch)
             self._prepare_for_arch(arch)
-            pkgs = self._get_packages(specs)
-            self.base.download_packages(pkgs, self.base.output.progress)
+
+            if self.opts.per_arch:
+                specs_by_arch = {arch: [], "noarch": [], "src": []}
```
kontura, line 162: Do I understand correctly that if there are multiple arches in `self.archs` this will handle all `noarch` and `src` packages repeatedly for each arch?

I know this was/is already happening even without the `--per-arch` but should we improve it?

plugins/manifest.py
```diff
@@ -157,8 +157,23 @@ def _download(self):
...
+                for spec in specs:
+                    nevra = hawkey.split_nevra(spec)
+                    if nevra.arch in ["noarch", "src"]:
+                        specs_by_arch[nevra.arch].append(spec)
+                    else:
+                        specs_by_arch[arch].append(spec)
+                for category, category_specs in specs_by_arch.items():
+                    if category_specs:
+                        self.base.repos.all().pkgdir = os.path.join(self.base.conf.destdir, category)
```
kontura: For local repositories the packages end up duplicated in `./packages.manifest` and in `./packages.manifest/{arch}/`.

In fact it seems that with this very recent PR: https://github.com/rpm-software-management/dnf/pull/2325 it seems to happen for all packages.

plugins/manifest.py
pkratoch: The packages won't get downloaded multiple times, but they are processed multiple times in the downloader, so I agree it could be improved, but I left it here for now, because it's not regression from the previous behavior and I wanted to keep the PR small.

plugins/manifest.py
pkratoch: Thanks for catching this. It should work properly now.

kontura (approved):
Thank you.

PR #615, opened 2026-05-06
reposync: add `--min-buildtime YYYY-MM-DD` filter

kontura:
https://redhat.atlassian.net/browse/RHEL-119475

This only supports `YYYY-MM-DD` to support also `HH:MM` we would probably need a new libdnf query filter.

I am thinking whether we should also add a runtime warning about: `Note that this option may result in partially synchronized repositories and the missing packages can cause dependency issues.`?


pkratoch (approved):
LGTM, thanks!

I don't think a runtime warning is necessary. You can end up with a package with missing dependencies even with the `--newest-only` option, although not as probable. If it turns out it's required for the reported issue, we can add it later.

PR #616, opened 2026-05-12
needs-restarting: Don't suggest restarting NEED_REBOOT services (RHEL-98293)

m-blaha:
"needs-restarting -s" used to list dbus-broker.service for restart after package updates like glibc. Restarting dbus-broker breaks system logins, firewalld, and other D-Bus clients.
  
Two layers of filtering are added to the -s output:
  
1. Stale files from NEED_REBOOT packages updated since boot are skipped. This reuses the existing -r (reboothint) check.
  
2. Services whose binary belongs to a NEED_REBOOT package are excluded. This covers the case where a service like dbus-broker is stale due to a non-NEED_REBOOT dependency (e.g. libselinux).
  
The -r mode now also detects services that cannot be safely restarted and includes them in the reboot recommendation.
  
Default mode (no flags) is unchanged — all stale PIDs are reported.

For: https://redhat.atlassian.net/browse/RHEL-98293

### DIFF
--- a/plugins/needs_restarting.py
+++ b/plugins/needs_restarting.py
@@ -34,6 +34,7 @@ import functools
 import os
 import re
 import stat
+import sys
 import time
 
 
@@ -163,7 +164,12 @@ def get_service_dbus(pid):
         'Id'
     )
     if name.endswith(".service"):
-        return name
+        # FragmentPath is the path to the unit file that defines the service
+        fragment_path = service_properties.Get(
+            "org.freedesktop.systemd1.Unit",
+            'FragmentPath'
+        )
+        return name, fragment_path
     return
 
 def smap2opened_file(pid, line):
@@ -358,7 +364,7 @@ class NeedsRestartingCommand(dnf.cli.Command):
                 return None
 
         stale_pids = set()
-        stale_service_names = set()
+        stale_services = {}
         uid = os.geteuid() if self.opts.useronly else None
         for ofile in list_opened_files(uid):
             pkg = owning_pkg_fn(ofile.presumed_name)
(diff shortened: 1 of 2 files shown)


plugins/needs_restarting.py
```diff
@@ -367,23 +353,75 @@ def run(self):
...
+            # non-NEED_REBOOT dependency like libselinux.
+            reboot_service_names = set()
+            for svc, pids in stale_service_pids.items():
+                for pid in pids:
+                    try:
+                        exe_path = os.readlink('/proc/%d/exe' % pid)
+                        if exe_path.endswith(' (deleted)'):
+                            exe_path = exe_path[:-len(' (deleted)')]
+                        exe_pkg = owning_pkg_fn(exe_path)
+                        if exe_pkg and exe_pkg.name in installed_need_reboot_pkgs:
```
evan-goode: Perhaps a better heuristic would be to check whether the service file is provided by a NEED_REBOOT package? If `bash` is a protected package, a service with `ExecStart=/usr/bin/bash -c foobar` is not necessarily unsafe to restart.

DNF5 does that here: https://github.com/rpm-software-management/dnf5/blob/28702d017ccba144d56a341004dc3c7b61f2ed66/dnf5-plugins/needs_restarting_plugin/needs_restarting.cpp#L411

evan-goode (commented):
Splitting the systemd services into NEED_REBOOT under `-r` and NOT NEED_REBOOT under `-s` is arguably better UX but it borders on being too big of a behavior change.

Also I don't follow "Stale files from NEED_REBOOT packages updated since boot are skipped. This reuses the existing -r (reboothint) check." If I update glibc, and foo.service depends on glibc, then it's correct to restart foo.service, right?

What about a simpler approach: if the unit file of any service listed by `dnf4 needs-restarting --services` is provided by a NEED_REBOOT package, print a warning to stderr:

```
sudo dnf4 needs-restarting --services
NetworkManager.service
dbus-broker.service
Warning: the following services should not be restarted. It is recommended you reboot the system instead: dbus-broker.service
```

That way stdout stays the same and the reporter's request is still satisfied. WDYT?l

evan-goode (approved):
Nice, looks great! Omitting the unsafe-to-restart services from stdout is probably better.

PR #618, opened 2026-06-25
Fix for issue #617, incorrect parsing of /proc/PID/stat

kbenson-sonic:
Fix for issue #617 to correctly parse /proc/PID/stat for cases where process names have a space.  The proc_pid_stat man page clearly states that the command name field is in parenthesis, and I've confirmed that this was true back to some ancient versions of the kernel (2.6 series), but the kernel source would give the true time (if it ever was anything different).

Confirmed working on an RHEL 9 based system with the shipping needs-restarting plugin in that, patched into same portion of master repo.


jrohel (approved):
Simple, working fix. Thanks for the PR!

PR #620, opened 2026-07-13
needs-restarting: Don't suggest restarting NEED_REBOOT services (RHEL-98293)

m-blaha:
"needs-restarting -s" used to list services like dbus-broker.service for restart after package updates. Restarting dbus-broker breaks system logins, firewalld, and other D-Bus clients.

Services whose unit file is provided by a NEED_REBOOT package are now excluded from the -s stdout and reported as a warning on stderr instead.

The -r mode and default mode (no flags) are unchanged.

For: https://redhat.atlassian.net/browse/RHEL-98293
Original upstream commit: ca40caa


evan-goode (approved):
Thanks, the needs-restarting.feature test passes on the rhel-10.3 branch of ci-dnf-stack.

